PR #8, opened 2026-05-26
[Hackathon] linux-kernel: per-hop latency models for the in-memory transport

mariagorskikh:
## Layer picked

**Transport** (layer 1) — reference plugin extension + simulator wiring.

## Why

The README has a section titled "Determinism & what the clock does" that explicitly flags this gap:

> The bundled `in_memory` transport delivers at `time = now`, so the virtual clock stays at `0.0` and `mean_latency` / `duration` will both be `0.0` in your trace... Latency *numbers* become meaningful only when ... you write a transport plugin that introduces per-hop delay.

For a testing tool that brags about "diff a trace against properties you care about," a transport that always reports zero latency is a sharp edge. NEST already exposes `mean_latency` as a built-in metric; it's just always 0.0. This PR fixes that without breaking anything.

## Core idea

A tiny, composable latency-model layer wired into the existing in-memory transport. One thing, well.

- `Simulator(latency_model=...)` — optional callable `(rng, sender, receiver) -> delay_seconds`. The simulator gets a **dedicated, seed-derived `_latency_rng`** so adding latency never perturbs the per-agent or failure RNG streams — existing traces stay byte-identical.
- `InMemoryTransport.send` schedules deliveries at `now + delay`, clamped at zero so the clock can never go backwards.
- New scenario field `transport_config:` (parsed by `runner._build_latency_model_from_config`) so users get realistic latency from YAML alone.

Six pure, composable models in `nest_plugins_reference.transport.latency`:

| `kind`        | Parameters                          | Use case |
|---------------|-------------------------------------|----------|
| `constant`    | `mean`                              | Every hop same. |
| `uniform`     | `low`, `high`                       | Bounded jitter. |
| `exponential` | `mean`                              | Long-tail packet network. |
| `normal`      | `mean`, `stddev`, `min_delay`       | Tight RTT distribution. |
| `pair_matrix` | `matrix: {"a,b": delay, ...}`, `default` | Heterogeneous topology (intra-DC fast, inter-DC slow). |
| `zero`        | —                                   | Explicit legacy default. |

Plus `with_jitter(base, jitter)` to wrap any model in ±uniform jitter.

## How to test

```bash
pip install -e packages/nest-core -e packages/nest-sdk -e packages/nest-plugins-reference -e packages/nest-cli -e packages/nest-scenarios -e packages/nest-mocks
pip install pytest pytest-asyncio hypothesis

# All 237 tests pass (193 baseline + 36 new + 8 runner-level).
pytest packages/nest-core/tests packages/nest-plugins-reference/tests -q

# End-to-end smoke: copy the marketplace YAML, append a latency block, run it.
cp scenarios/marketplace.yaml /tmp/m.yaml
cat >> /tmp/m.yaml <<EOF

transport_config:
  kind: exponential
  mean: 0.020
  jitter: 0.005
EOF
nest run /tmp/m.yaml -o /tmp/m.jsonl

# Now mean_latency is ~0.02 instead of 0.0; virtual clock advances.
python -c "
import json, statistics
ev=[json.loads(l) for l in open('/tmp/m.jsonl') if l.strip()]
sends={}; ds=[]
for e in ev:
    c=e.get('corr','')
    if not c: continue
    if e['kind']=='send': sends[c]=e['ts']
    elif e['kind']=='receive' and c in sends: ds.append(e['ts']-sends[c])
print('mean_latency:', round(statistics.fmean(ds),5))
print('clock_max   :', round(max(e['ts'] for e in ev),5))
"
```

Run it twice — the traces are byte-identical (`md5sum` confirms).

## Key assumptions / design notes

- **Backward-compatible by construction.** No `transport_config` ⇒ no latency model ⇒ `time = now` ⇒ existing traces are byte-for-byte unchanged. I verified the bundled `marketplace` baseline trace MD5-matches across a fresh run on this branch.
- **Determinism preserved.** The latency RNG is derived from `seed` directly (`(seed * 2654435761 + 0xA1A7C7) & ((1<<63)-1)`), not from the master RNG sequence. Adding/removing the latency feature does not change the per-agent or failure-RNG streams for an existing scenario.
- **Clamp at zero.** All models clamp the delay to `max(0.0, ...)` so the virtual clock can never move backwards — important for the `normal` model and for `with_jitter`.
- **Compatible with existing failure injection.** Verified by `test_latency_compatible_with_drop_rate`: latency + `message_drop_rate` + Byzantine + partitions all coexist.
- **Discoverable from YAML, scriptable from Python.** The factory accepts either `transport_config: {kind: ..., mean: ...}` at the top level or nested as `transport_config: {latency: {kind: ..., ...}}` — whichever reads better.

## Future work

- A `latency_model` could itself become a named plugin in `nest.plugins.transport`, so the registry-driven plugin path (currently shadowed by the simulator's hard-wired `InMemoryTransport`) gets exercised.
- A `latency_p50`/`latency_p99` metric alongside `mean_latency` — easy now that the data is real.
- Drop / corruption rate per-pair (`pair_matrix`-style) to model lossy WAN links.
- A `replay` transport that reads latencies from a captured production trace.

## Persona

Linux kernel maintainer — networking, queues, scheduling. Small, sharp tools. Pure functions, no global state, one knob per job.

---

## Summary by Sourcery

Introduce configurable per-hop latency modeling to the in-memory transport and wire it through the simulator and scenario runner while preserving backward-compatible zero-latency behavior.

New Features:
- Add optional per-hop latency configuration to scenarios via a new transport_config field that drives a latency model for the in-memory transport.
- Provide a latency_model hook on the Simulator and InMemoryTransport to delay message deliveries in virtual time using a dedicated RNG.
- Implement a set of reusable latency models and a factory in nest_plugins_reference (constant, uniform, exponential, normal, pair_matrix, zero, plus optional jitter) that can be constructed from YAML specs.

Enhancements:
- Ensure latency modeling preserves determinism and legacy zero-latency traces when no configuration is provided.
- Extend documentation with a new section describing per-hop latency configuration and available models.
- Add helper utilities in the runner to build latency models from transport_config, including support for nested latency blocks.

Tests:
- Add simulator-level integration tests covering latency behavior, determinism, interaction with drop rate, and pair-matrix topologies.
- Add unit tests for all latency model primitives and the factory construction logic.
- Add end-to-end runner tests verifying YAML-driven latency configuration, backward compatibility, and deterministic traces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,9 +296,36 @@ Two things to know that aren't obvious:
    transport delivers at `time = now`, so the virtual clock stays at
    `0.0` and `mean_latency` / `duration` will both be `0.0` in your
    trace. The event queue still orders events deterministically, so
-   *correctness* tests are fine. Latency *numbers* become meaningful
-   only when agents use `ctx.schedule(delay, ...)` or you write a
-   transport plugin that introduces per-hop delay.
+   *correctness* tests are fine. To get meaningful latency *numbers*,
+   add a `transport_config:` block to your scenario YAML (see below) or
+   use `ctx.schedule(delay, ...)` from your agent code.
+
+### Per-hop latency
+
+Add this to any scenario YAML to make the in-memory transport delay
+each delivery by a sampled per-hop latency. The samples are drawn from
+a dedicated RNG so the trace stays byte-identical under a fixed seed.
+
+```yaml
+transport_config:
+  kind: exponential   # constant | uniform | exponential | normal | pair_matrix | zero
+  mean: 0.020         # seconds
+  jitter: 0.005       # optional uniform jitter ±0.005 s, clamped at 0
+```
+
+Available models:
+
+| `kind`         | Parameters                              | When to use it |
+|----------------|-----------------------------------------|----------------|
+| `constant`     | `mean`                                  | Every hop costs the same. |
+| `uniform`      | `low`, `high`                           | Bounded jitter, easy to reason about. |
+| `exponential`  | `mean`                                  | Long-tail latencies typical of real packet networks. |
+| `normal`       | `mean`, `stddev`, `min_delay`           | Tight distributions around a target RTT. |
+| `pair_matrix`  | `matrix: {"from,to": delay, ...}`, `default` | Heterogeneous topologies (fast intra-DC, slow inter-DC). |
+| `zero`         | —                                       | Explicit zero-latency (the legacy default). |
+
+`mean_latency` and `duration` in your trace become meaningful as soon
+as `transport_config` is present.
 
 ---
 
--- a/packages/nest-core/nest_core/runner.py
+++ b/packages/nest-core/nest_core/runner.py
@@ -47,7 +47,9 @@ def _build_latency_model_from_config(
     #     transport_config:
     #       kind: exponential
     #       mean: 0.02
-    spec = raw_latency if isinstance(raw_latency, dict) else transport_config
+    spec: dict[str, Any] = (
+        cast("dict[str, Any]", raw_latency) if isinstance(raw_latency, dict) else transport_config
+    )
     if not spec:
         return None
     try:
--- a/packages/nest-core/nest_core/runner.py
+++ b/packages/nest-core/nest_core/runner.py
@@ -28,6 +28,35 @@ def _parse_partition_groups(raw: object) -> list[list[str]] | None:
     return result if result else None
 
 
+def _build_latency_model_from_config(
+    transport_config: dict[str, Any] | None,
+) -> Any:
+    """Build a ``LatencyModel`` from scenario ``transport_config`` or return None.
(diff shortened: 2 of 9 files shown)


JamesCarnley:
Thank you @mariagorskikh. Reviewing this submission preserved an important fail-closed requirement: malformed network and fault configuration must be rejected rather than silently changing the meaning of a test. That current gap has now been independently implemented, tested, reviewed, and merged in #255 (merge dfca1903bd22da9309629feda38b02273298c188), with explicit credit to this source. The replacement validates the existing nth, delay, and drop-rate domains before a run starts. We are closing this archive/legacy PR now that the retained requirement has landed. Constant, stochastic, pair-matrix, and jitter latency models plus the legacy transport wiring were not ported. No contributor code from this PR was executed or merged.

PR #11, opened 2026-05-26
[Hackathon] google-staff: netem transport + tail-latency metrics + SLO validators

mariagorskikh:
## Piece picked

**Layer 1 — Transport.** A `netem`-style alternative transport plugin, plus the observability and assertion machinery a real SRE would need to use it: tail-latency metrics and per-scenario SLO validators.

## Why

The README is honest about it:

> The default transport is zero-latency. ... `mean_latency` / `duration` will both be `0.0` in your trace. Latency *numbers* become meaningful only when ... you write a transport plugin that introduces per-hop delay.

The mean is the least interesting number in a distributed system anyway. **Real protocols fail at the tail.** Without a way to inject realistic per-link delay, jitter, bandwidth, and reorder, and without p50/p95/p99 metrics and pass/fail SLO checks, NEST can verify *correctness* but cannot answer "will this still satisfy 250ms p99 at 1000 agents?" — which is the question every protocol team eventually has to answer in production.

## Core idea

1. **`DelayModel`** (`nest_core.sim.delay_model`) — a netem-style per-link emulator: constant / uniform / lognormal-tailed latencies, jitter, bandwidth-based serialization delay, per-link drops, reproducible per-link FIFO. Pure with respect to a caller-supplied `random.Random`, so the simulator stays byte-deterministic.
2. **Simulator wiring** — `InMemoryTransport` accepts an optional `delay_model`. When `None` (the default), behaviour is byte-identical to today. When attached, every `send`/`broadcast` consults the model and the delivery event is scheduled at `now + delay`. Drops surface as `dropped` trace events with `reason: netem`.
3. **`netem` plugin** — `nest_plugins_reference.transport.netem` exposes a `StandaloneNetemTransport` for Tier-2 / shell use (delays via `asyncio.sleep`) and a `make_delay_model` helper. Registered as a built-in `("transport", "netem")` in the plugin registry.
4. **Scenario YAML** — new optional `transport:` block (parsed into `TransportConfig`) and new optional `slo:` block. The runner builds a `DelayModel` only when `layers.transport == "netem"`, seeded from `scenario.seed XOR transport.seed_salt` so reruns are reproducible and the transport's randomness can be perturbed independently of agent behaviour.
5. **Metrics** — added `p50_latency`, `p95_latency`, `p99_latency`, `max_latency`, computed from corr-paired send/receive events using nearest-rank (matches what netem and most prod tooling report).
6. **SLO validators** — `validate_latency_slo` checks `p50/p95/p99/max_latency` budgets and `min_delivery_rate`. Wired into the runner: scenarios with an `slo:` block get pass/fail results on `runner.validations` automatically.

## How to test

```bash
uv sync
uv run ruff check . && uv run ruff format --check .
uv run pyright
uv run pytest -q
```

All green on my machine: **293 tests pass** (up from 259 baseline; +34 new), zero ruff findings, zero pyright errors.

Try the bundled example:

```bash
cp examples/netem-transport/marketplace-netem.yaml /tmp/
uv run nest run /tmp/marketplace-netem.yaml -o /tmp/netem.jsonl
```

Or programmatically:

```python

cfg = ScenarioConfig.from_yaml("examples/netem-transport/marketplace-netem.yaml")
runner = ScenarioRunner(cfg)
await runner.run()
print(runner.metrics)       # p50/p95/p99/max all > 0
print(runner.validations)   # pass/fail per SLO budget
```

A representative run from this branch on the bundled example:

```
mean_latency      0.034125 s
p50_latency       0.021009 s     (target 0.020)
p95_latency       0.111038 s
p99_latency       0.189972 s     (target 0.200, budget 0.250 -> PASS)
max_latency       0.820940 s     (heavy tail from lognormal)
slo_p99_latency           PASS  observed 0.189972s vs budget 0.250000s over 1000 messages
slo_min_delivery_rate     PASS  observed 1.0000 vs target 0.9900
```

Tightening the budget to `p99_latency: 0.050` immediately flips to `FAIL`, which is exactly what you want from a regression signal.

## Key assumptions

- **Determinism is non-negotiable.** Same `seed` + same YAML → byte-identical trace. The delay model uses its own seeded RNG so swapping `seed_salt` perturbs only the transport.
- **Backwards-compatible by default.** No existing scenario, validator, or metric changes behaviour. `delay_model=None` is the same code path as before.
- **Unit-sized messages.** This is a netem-style model for agent-protocol messages, not a packet-level DPDK emulator. Serialization delay is `bytes * 8 / kbps`.
- **Tier 1 only for the wired plugin path.** `StandaloneNetemTransport` exists for Tier 2 / shell use but is not deterministic by design (it sleeps in real time).
- **Nearest-rank percentiles**, since that is what netem and most production telemetry report and it never returns a value that wasn't actually in the distribution.

## Persona

Google staff engineer (Spanner / Borg DNA): metrics before features, SLOs over feelings, deterministic reproducibility over clever heuristics, and tail latency is the only latency that matters.

## Future work

- Per-link bandwidth queues (model HOL blocking under burst load, not just per-message serialization).
- A `chaos:` scenario hook that flips `seed_salt` mid-run for chaos-style regression sweeps.
- Hook the HTML report to plot the per-message latency CDF and color rows by SLO violation.
- A `validate_latency_slo` variant that operates on a *rolling window* so long-tail spikes show up even when the overall p99 stays under budget.
- Extend the model to Byzantine link behaviour (per-pair partition timers) so it composes cleanly with `failures.network_partition`.
- A `nest dashboard` panel that surfaces SLO pass/fail next to the metrics it depends on.

---

### DIFF
--- a/examples/netem-transport/README.md
+++ b/examples/netem-transport/README.md
@@ -1,0 +1,95 @@
+# `netem` transport: realistic latency in NEST
+
+The bundled `in_memory` transport delivers messages at `time = now` so every
+trace shows zero latency. That is fine for correctness work — and useless
+for SLO / tail-latency / capacity work. `netem` is a sim-aware drop-in that
+injects per-message latency, jitter, bandwidth-bound serialization delay,
+reordering, and per-link drops, all driven by the seeded simulator RNG.
+
+## Use it
+
+In any scenario YAML:
+
+```yaml
+layers:
+  transport: netem
+
+transport:
+  latency:
+    kind: lognormal    # constant | uniform | lognormal
+    p50_ms: 20
+    p99_ms: 200        # heavy-tail by construction
+  jitter_ms: 2
+  bandwidth_kbps: 1000 # optional — adds size-based serialization delay
+  reorder_prob: 0.0
+  drop_prob: 0.0
+  seed_salt: 0         # XORed with the scenario seed for the model rng
+```
+
+## Why this matters
+
+Real systems fail under tail latency, not average latency. With `netem`:
+
+- `mean_latency` becomes meaningful.
+- New `p50_latency`, `p95_latency`, `p99_latency`, `max_latency` metrics
+  reflect the actual distribution.
+- A scenario-level `slo:` block lets you assert latency / availability
+  budgets as part of the validators run, e.g.:
+
+  ```yaml
+  slo:
+    p99_latency: 0.250      # 250ms, in seconds
+    min_delivery_rate: 0.99
+  ```
+
+  Pass-or-fail per budget shows up in `runner.validations` and via
+  `nest_core.validators.validate_trace(..., slo=...)`.
+
+## Determinism
+
+`netem` uses an independent PRNG seeded by `scenario.seed XOR transport.seed_salt`
+so:
+
+- The same scenario YAML reruns produce byte-identical traces.
+- You can perturb only the transport (e.g. for chaos tests) by changing
+  `seed_salt` without changing agent behaviour.
+
+## Standalone (Tier 2)
+
+`StandaloneNetemTransport` lives in `nest_plugins_reference.transport.netem`
+and uses `asyncio.sleep` for delays — useful for shell / LLM agents.
+
+```python
+from nest_plugins_reference.transport.in_memory import InMemoryNetwork
+from nest_plugins_reference.transport.netem import (
+    StandaloneNetemTransport,
+    make_delay_model,
+)
+
+network = InMemoryNetwork()
+model = make_delay_model({
+    "latency": {"kind": "lognormal", "p50_ms": 20, "p99_ms": 200},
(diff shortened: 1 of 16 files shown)


JamesCarnley:
Thank you, @mariagorskikh. A future performance profile should declare its latency distribution, seed, topology, serialization and queue rules, sample count, percentile definition, warm-up, and SLO basis, then distinguish a healthy control from a threshold mutation and report insufficient evidence when samples are absent. Current Town's MemoryTransport provides deterministic fixed latency plus explicit drop, duplicate, and delay faults; it does not claim production-network performance or SLOs. The narrow malformed-fault input gap is being repaired separately from #8. I am closing this legacy PR as a future transport profile while retaining its hand-computed timing and trace-recomputed percentile cases. No contributor code or network model was executed during this review.

PR #29, opened 2026-06-21
[Hackathon] feat(auth): implement (Problem 4) delegatable capability tokens with cascading revocation 

anshuS1310:
## Overview

I implemented Macaroon-style delegatable capability tokens for the auth layer (Problem 4).

Basically, the old JWT-only setup was pretty static — there was no way to delegate a token to another agent with fewer permissions than you have, and if you revoked a parent token there was no way for that to automatically kill off any child tokens that came from it.

This PR adds a `delegatable` auth plugin that handles delegation chains, scope narrowing, and cascading revocation, all offline (no central DB needed).

**Persona:** writing this from more of an applied-crypto angle — I care less about "trust me it's secure" and more about being able to point at a specific attack and say exactly which check stops it.

Thanks for the review on v1, really helpful. Fixed all 4 points below + rebased onto main.

## What changed since the last review

- **`Auth.verify()` doesn't break anything anymore.** Made `presenter` an optional kwarg (`presenter: AgentId | None = None`) instead of a required positional arg. So old code calling `verify(token)` still just works — if you don't pass a presenter it just skips that check. Because of this I didn't even need to touch the 3 existing JWT tests anymore, so that diff basically disappeared.
- **YAML duplication — turns out both copies are actually needed, so I documented that instead of just picking one.** Checked `packages/nest-core/pyproject.toml` and only stuff under `nest_core/` actually gets packaged into the wheel, so `scenarios_builtin/` is the copy that makes `nest run delegated_auth` work if someone installs the package instead of cloning the repo. The root copy is just there so devs can actually find it in the repo. Instead of a test that just yells "these don't match!", I added a `make sync-scenarios` command that actually fixes it, and the test now just checks they match and tells you to run that command if they don't.
- **Reverted the `gossip_registry.yaml` comments.** My bad, those slipped in from a different thing — that scenario's already merged in #24 so it didn't belong here. Happy to open a tiny separate PR for those docs if they're actually wanted.
- **Rebased onto current main.**

## What I did

1. **HMAC chaining for delegation** — in `delegatable.py`, a parent token can delegate a child token, and the child's key is derived like `child_key = hmac(parent_key, child_token_id)`. So you can verify the whole chain back to the root secret without needing a database somewhere.

2. **Scope narrowing** — every time a token gets delegated (or verified), we check the child's scopes are a subset of the parent's scopes (`set.issubset()`). If something down the chain tries to sneak in extra permissions, it throws a `ValueError`.

3. **Cascading revocation** — when verifying, we walk up the whole chain. If any parent/grandparent token was revoked, the whole thing is invalid — raises a `RevokedAncestorError` (subclass of `ValueError`).

4. **Optional presenter binding** — if you call `verify(token, presenter)`, it checks the presenter actually matches who the token was meant for (`last_aud`). Mismatch = `ValueError`. It's optional so old callers that don't pass a presenter aren't affected.

## Threat model — what each validator actually stops

Wanted to be specific here instead of just saying "more secure," so here's the actual attack each one blocks and why plain JWT wouldn't catch it:

1. **Scope escalation (`validate_no_scope_escalation`)**
   *Attack:* some leaf agent (like `leaf-4`) only has `["read"]` scope but tries to pass itself off as having `["read", "write"]`.
   *Why JWT baseline fails:* plain JWT just checks a signature on whatever claims are in the token — no concept of a delegation chain, so nothing stops someone from just claiming a bigger scope.
   *Why `delegatable` catches it:* scopes get checked as subsets at every single hop, both when the token is made and when it's verified. Any escalation → `ValueError`.

2. **Stale/revoked ancestor (`validate_no_stale_parent`)**
   *Attack:* `leaf-0` tries to use a grandchild token even though a parent or the root got revoked.
   *Why JWT baseline fails:* JWT revocation is usually just "is this exact token on a blacklist." Nothing links parent and child tokens together, so revoking the parent does nothing to kids that were already handed out.
   *Why `delegatable` catches it:* each child's key is derived from its parent's signature, so verifying walks the *entire* chain and checks every ancestor's revoke status, not just the token itself. Any revoked ancestor → `RevokedAncestorError`.

3. **Audience confusion / replay (`validate_no_audience_confusion`)**
   *Attack:* `leaf-2` grabs a token meant for `leaf-1` and tries to use it like it's theirs.
   *Why JWT baseline fails:* normal JWT verification just checks the signature, not who's actually holding it — so if you intercept a valid token you can just replay it.
   *Why `delegatable` catches it:* when you pass a presenter into `verify()`, it's checked against the token's actual intended audience. Mismatch → `ValueError`.

## Testing

Ran everything locally, all green:

- **Unit tests:** 8 tests in `test_delegatable_auth.py` — scope subsets, parent/grandparent revocation, audience checks. Also confirmed this whole file still passes fine with the new optional `presenter` arg.
- **Integration scenario:** the `delegated_auth` scenario + the 3 validators above. `test_delegated_auth_scenario.py` shows:
  - Running on `jwt` → validators fail (attack goes through).
  - Running on `delegatable` → validators pass (attack gets blocked).
- **Full CI:** ran `make ci-local`, all 5 stages passed (`uv sync`, `ruff check`, `ruff format --check`, `pyright`, `pytest -v`, 555+ tests).
- **Backward compat:** re-ran the original 3 `JwtAuth` tests unchanged (no presenter arg) — still pass fine.

Lmk if you want more detail anywhere on the threat model stuff, happy to add more!


Skyrider3:
## Review: Request changes 

Thanks for this PR,  the structure is genuinely strong (clean separation, good docstrings, `hmac.compare_digest`, positional key-chaining so frame **reordering** is caught, scope-subset + TTL-≤-parent enforcement on `delegate()`, and a real adversarial jwt-vs-delegatable differential harness). All 10 of your tests pass for me locally.

But there's a **critical design flaw in the token format** that defeats both headline properties (no scope escalation, cascading revocation) through the standard `Auth.verify(token)` path, and the test suite doesn't catch it. Details below, with a reproducible PoC.

---

###  CRITICAL — Truncation attack: any delegated token strips back to its fully-privileged root

A delegated token is `json.dumps(parent_frames + [child_frame])` (`delegatable.py:188`), so **every ancestor frame is embedded verbatim with its own valid signature.** Because `verify` re-derives each frame's key from the master secret and checks each signature independently, `frames[0]` (the root) is a complete, independently-valid token on its own. A holder just deletes the trailing frames.

This breaks the macaroon model. Real macaroons are truncation-proof because only the *final* chained signature is ever transmitted (`sig_n = HMAC(sig_{n-1}, caveat_n)`); intermediate signatures never leave the issuer, so a shortened chain can't be presented. Here, all intermediate signatures ship inside the token.

The only thing preventing exploitation is the audience check at `delegatable.py:261`, which is **`presenter is not None`-gated and not part of the `Auth` protocol** (`layers/auth.py:37` is `async def verify(self, token)` — no `presenter`). So any code verifying through the standard interface is fully exploitable. The scenario only escapes this because `CoordinatorAgent._verify` introspects the signature and passes `presenter` — which is exactly why the validators pass while the real hole goes untested.

**PoC (drop in `packages/nest-plugins-reference/tests/`, runs against the package as-is):**

```python
import asyncio, json
from nest_core.types import AgentId, Token
from nest_plugins_reference.auth.delegatable import DelegatableAuth

async def main():
    auth = DelegatableAuth(secret=b"server-master-secret", clock=100.0)
    root = await auth.issue(AgentId("alice"), ["read", "write", "admin"])
    bob  = await auth.delegate(root, AgentId("bob"), ["read"], ttl=100.0)  # low-priv

    # Bob strips the delegation frame, keeping only the validly-signed root frame:
    forged = Token(json.dumps([json.loads(str(bob))[0]]))

    ctx = await auth.verify(forged)            # Auth-protocol call: no `presenter`
    print("scopes:", ctx.scopes, "subject:", ctx.subject)
    assert ctx.scopes == ["read"], f"ESCALATION: got {ctx.scopes} as {ctx.subject}"

asyncio.run(main())
```

Output — Bob escalates `["read"]` → `["read","write","admin"]` **and** impersonates `alice` (subject is read from the surviving frame):

```
scopes: ['read', 'write', 'admin'] subject: alice
AssertionError: ESCALATION: got ['read', 'write', 'admin'] as alice
```

**Same trick defeats cascading revocation** — the feature this PR is about. After `await auth.revoke(bob)` (which blacklists the child's `token_id`), `await auth.verify(forged)` still returns full scopes because the root frame's `token_id` was never revoked. Revoking a parent does **not** stop a holder who keeps the root frame.

Residual even if `presenter` were mandatory: with same-audience re-delegation `alice → bob[read,write] → bob[read]`, Bob truncates to the middle frame (`audience == bob`) and re-escalates to `[read,write]`. The audience check is a band-aid, not the fix.

**Fix direction:** chain the signatures so a prefix can't stand alone — each frame's MAC must cover the parent's MAC, and only the tail signature should be transmitted (true macaroon construction). Then make the presenter/audience binding a **mandatory part of the `Auth.verify` contract**, not an optional kwarg, so generic callers can't skip it.

---

###  Medium — `verify` violates the `ValueError` error contract

The `try/except` at `delegatable.py:199-204` wraps only `json.loads`, not the subsequent field accesses. Malformed-but-valid-JSON tokens raise raw `KeyError`/`TypeError` instead of `ValueError` (which is how `JwtAuth` and the rest of the codebase signal auth failure):

- `[{}]` → `KeyError: 'token_id'`
- `[123]` → `TypeError: 'int' object is not subscriptable`
- `[{"token_id":"x"}]` → `KeyError: 'expires_at'`

A caller catching `ValueError` won't catch these → uncaught exception / stack-trace leak on hostile input. Validate frame shape up front and raise `ValueError`.

###  Medium — Description overclaims the trust model

The description says delegation works "entirely offline … without needing a central database." It doesn't — `delegate()` requires the master secret (via `_token_keys` or `_derive_key_chain`). An edge node with the wrong secret produces a child the real server rejects. This is a centralized, symmetric capability scheme: only the secret-holder can mint or delegate. That's fine for a simulation, but please correct the description (the *verify*-without-DB claim is true; the peer-to-peer offline *delegation* claim is not).

###  Minor
- The audience check inspects only `frames[-1]`; combined with the critical issue, chain integrity rests on one optional comparison.
- `_token_keys` grows unbounded (every issued/delegated/verified key cached forever) — fine at sim scale, worth a note/TTL.
- `scenarios/delegated_auth.yaml` and `scenarios_builtin/yaml/delegated_auth.yaml` are byte-identical duplicates — matches repo convention but is copy-paste drift risk.

anshuS1310:
Ohk will implement this changes by today itself

anshuS1310:
@Skyrider3 check I made the changes according to the review

dhve:
Thanks @anshuS1310. The delegatable-tokens design is genuinely interesting. HMAC-chained parent/child keys, transitive scope narrowing, and cascading revocation is the right shape for this problem, and the plugin implementation itself (`delegatable.py` + tests) reads well.

**1. The change to `Auth.verify()` is a breaking API change.**

You changed `Auth.verify(token: Token) -> AuthContext` to `verify(token: Token, presenter: AgentId) -> AuthContext`. Adding a required positional argument breaks every current caller and every alternative auth implementation. The presenter-binding logic you added to `JwtAuth.verify` is a legitimate security improvement in its own right, but the protocol-level API change has much wider blast radius than your delegatable plugin needs.

Two cleaner options:

- Make `presenter` an optional keyword parameter with `None` default. Existing callers keep working; delegatable-aware callers pass it; `JwtAuth` treats `None` as "skip the presenter check" or as an implicit trust-the-subject.
- Add a new method (`verify_for(presenter, token)` or similar) alongside `verify`. Delegatable requires it; existing callers keep calling `verify` unchanged.

Either avoids breaking other plugins and lets your work land cleanly. The changes to `test_plugins.py` (updating three JWT tests to pass an extra arg) will disappear if you go the optional-kwarg route.

**2. YAML duplication pattern feels off.**

`delegated_auth.yaml` exists in two places: `packages/nest-core/nest_core/scenarios_builtin/yaml/` and top-level `scenarios/`. Confirmed byte-identical. The new `test_yaml_sync.py` is a linter that enforces they stay identical. Keeping two copies in sync via a test is a symptom, not a fix.

Pick one canonical location. If both are actually needed for different reasons, document that reasoning and consider generating one from the other rather than a byte-equality test.

**3. `gossip_registry.yaml` changes are out of scope.**

You added 14 lines of documentation comments to `nest_core/scenarios_builtin/yaml/gossip_registry.yaml`. That scenario ships with PR #24 (already merged). If those docs are worth having, please pull them into a separate small PR so this one stays focused on Problem 4.

**4. Persona and threat model in the PR body.**

The description explains what the plugin does but doesn't state the persona you're writing from, and doesn't say what specific attacks the three validators catch. The judge rubric weights adversarial framing heavily, so it's worth spelling out:

- Which persona you're presenting as (applied cryptography? distributed systems?).
- For each of your three validators: the specific attack it catches, and what the failure mode looks like on the `jwt` baseline that `delegatable` defends against.

That framing makes the "passes on us, fails on baseline" pattern legible in review, and lines up with the demo bar for July 11.

**Suggested path forward:**

1. Change `verify()` to accept `presenter` as an optional kwarg (or add a separate method); update `JwtAuth` and `test_plugins.py` to preserve existing behavior when `presenter` is not supplied.
2. Delete the `gossip_registry.yaml` doc-comment changes from this branch. Open a separate docs PR later if you want.
3. Pick one canonical location for `delegated_auth.yaml`, delete the other, delete `test_yaml_sync.py` unless there's a separate justification for keeping two copies.
4. Expand the PR body's threat-model section as described above.
5. Rebase onto current main. Your branch is behind by several merged PRs and there may be conflicts to reconcile.
6. Run `make ci-local` locally and confirm all 5 stages green before requesting re-review.

Leaving this open. Ping me on the hackathon channel if the API-change reshaping is unclear.

anshuS1310:
@dhve @Skyrider3 can you check now  I've successfully rebased the branch onto the current main, resolved the conflicts, and addressed all your requested changes. I can't message you through the channel I'm in my college rn and my college's slow wi-fi firewall does not allow discord

anshuS1310:
@dhve
@Skyrider3

JamesCarnley:
Thank you, @anshuS1310. The reusable cases are whole-chain verification before delegation or revocation, mandatory attack execution with typed rejection evidence, presenter binding, and truncation or middle-frame deletion attacks. Current Town's controller-signed Run Grants are flat and already test their selected one-hop contract. This alternate legacy rewrite duplicates the delegation family and, at the frozen head, revoke still mutates from an unverified identifier while validators can treat missing attacks or broad exceptions as success. I am closing it as a duplicate and retaining those cases under the consolidated future delegation profile rather than porting this implementation. No contributor code was merged or executed during this review.

PR #32, opened 2026-06-22
feat(examples): Agent Action Capsule trust layer + Stripe payment demo

StevenMih:
## Summary

Lands a working `capsule-emit-nanda` plugin package so NANDA users can
actually install and use the capsule trust + payments layers, with demos
underneath.

**Gap in the original PR:** `pip install capsule-emit[nanda]` didn't
exist, and `nest run scenarios/receipt_reputation_capsule.yaml` would
fail for anyone who cloned the repo.

### What's new: `examples/capsule-emit/`

Installable `capsule-emit-nanda` package with two real NANDA layer
plugins registered via entry points:

```bash
pip install -e examples/capsule-emit
nest plugins list | grep -E "trust|payments"
# capsule_emit   (trust)
# stripe_capsule (payments)
```

**`trust: capsule_emit`** — `CapsuleEmitTrust`
- Drop-in for `agent_receipts`; zero agent-code changes
- Seals every corroborated receipt to an Agent Action Capsule ledger
- Ring-severance validators still pass (ring mean ≈ 0.0, honest > 0.1)
- Third-party verifiable: `agent-action-capsule verify --store capsule_ledger.jsonl`

**`payments: stripe_capsule`** — `StripeCapsuledPayments`
- Sandbox/mock by default — no `STRIPE_SECRET_KEY`, no real charges
- Real Stripe with `pip install -e "examples/capsule-emit[stripe]"` + `STRIPE_SECRET_KEY=sk_test_...`
- Every payment sealed; `agent_input_digest` commits to amount + payer/payee at call time
- Dispute proof: re-derive digest from disputed values; mismatch = tamper evidence

### Updated

- `examples/capsule-trust/README.md` — capability-first, install points at real package
- `examples/stripe-capsule-payment/README.md` — capability-first with code pattern; sandbox vs real-Stripe; IETF slug fixed
- `scenarios/receipt_reputation_capsule.yaml` — install comment matches real package path
- IETF slug corrected everywhere: `draft-steele-agent-action-capsule` → `draft-mih-scitt-agent-action-capsule`

## Test plan

- [ ] `pip install -e examples/capsule-emit` → no errors
- [ ] `nest plugins list | grep -E "trust|payments"` shows `capsule_emit` and `stripe_capsule`
- [ ] `nest run scenarios/receipt_reputation_capsule.yaml` → validators pass, `capsule_ledger.jsonl` produced
- [ ] `agent-action-capsule verify --store capsule_ledger.jsonl` → exit 0
- [ ] `pip install -e "examples/capsule-emit[stripe]"` works; `StripeCapsuledPayments` sandbox runs without a real key
- [ ] IETF links resolve: `datatracker.ietf.org/doc/draft-mih-scitt-agent-action-capsule/`

## Background

The [Agent Action Capsule](https://datatracker.ietf.org/doc/draft-mih-scitt-agent-action-capsule/)
is an IETF draft (Apache-2.0 reference library) that seals agent decisions
as tamper-evident records anchored to a SCITT transparency log.

This PR shows two NANDA integration points — a one-line YAML swap that
adds an independently-verifiable ledger to any scenario, and a payments
layer that commits to payment amounts at execution time.

### DIFF
--- a/examples/capsule-emit/README.md
+++ b/examples/capsule-emit/README.md
@@ -7,7 +7,7 @@ by any third party without replaying the scenario.
 | Layer | Plugin name | What it adds |
 |-------|-------------|--------------|
 | `trust` | `capsule_emit` | Drop-in for `agent_receipts`; seals every receipt to a capsule ledger. Ring-severance identical; adds third-party auditability. |
-| `payments` | `stripe_capsule` | Stripe (or sandbox) payments; every payment sealed with `agent_input_digest` committing to amount + payer/payee at call time. |
+| `payments` | `stripe_capsule` | **Demo** — Stripe (or sandbox) payments; every payment sealed with `agent_input_digest` committing to amount + payer/payee at call time.  Not a conforming NANDA Payments layer (see [`stripe-capsule-payment/README.md`](../stripe-capsule-payment/)). |
 
 ## Install
 
--- a/examples/capsule-emit/README.md
+++ b/examples/capsule-emit/README.md
@@ -1,0 +1,78 @@
+# capsule-emit-nanda — verifiable, anchored records for NANDA agents
+
+Two NANDA layer plugins that seal every agent interaction and payment into
+a tamper-evident [Agent Action Capsule][] ledger — independently verifiable
+by any third party without replaying the scenario.
+
+| Layer | Plugin name | What it adds |
+|-------|-------------|--------------|
+| `trust` | `capsule_emit` | Drop-in for `agent_receipts`; seals every receipt to a capsule ledger. Ring-severance identical; adds third-party auditability. |
+| `payments` | `stripe_capsule` | Stripe (or sandbox) payments; every payment sealed with `agent_input_digest` committing to amount + payer/payee at call time. |
+
+## Install
+
+```bash
+pip install -e examples/capsule-emit   # from the nandatown repo root
+nest plugins list | grep -E "trust|payments"
+# capsule_emit   (trust)
+# stripe_capsule (payments)
+```
+
+For real Stripe payments add `[stripe]`:
+
+```bash
+pip install -e "examples/capsule-emit[stripe]"
+STRIPE_SECRET_KEY=sk_test_... nest run ./scenario.yaml
+```
+
+## Use in any scenario
+
+```yaml
+layers:
+  trust: capsule_emit        # was: agent_receipts
+  payments: stripe_capsule   # was: prepaid_credits
+```
+
+Run any scenario — all existing validators still pass, and a
+`capsule_ledger.jsonl` appears alongside the trace:
+
+```bash
+agent-action-capsule verify --store capsule_ledger.jsonl   # exit 0
+```
+
+Each capsule's `agent_input_digest` is independently verifiable without
+re-running the scenario. A tampered record produces a digest mismatch
+that `verify` catches.
+
+## Trust layer: `capsule_emit`
+
+Identical ring-severance logic to `agent_receipts` plus an anchored
+ledger gate: only receipts whose interactions were sealed to the capsule
+ledger count toward an agent's reputation score. See
+[`examples/capsule-trust/`](../capsule-trust/) for the full walkthrough.
+
+## Payments layer: `stripe_capsule`
+
+Sandbox by default (no `STRIPE_SECRET_KEY` required, no real charges).
+Seals the Stripe `payment_intent_id`, amount, and outcome — if a client
+later disputes the amount, re-derive the digest from the disputed values;
+a mismatch is proof the amount was altered post-payment. See
+[`examples/stripe-capsule-payment/`](../stripe-capsule-payment/) for the
+full walkthrough.
+
+## Demos
+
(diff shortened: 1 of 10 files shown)


Skyrider3:
## Review: Request changes (leaning approve) 

Nice contribution!  this closes a real gap (`pip install` of the capsule layers actually existing), and the trust integration is the genuine value here. No security concerns: the Stripe key is set on the client and never logged, capsule output carries only the intent ID, and sandbox-by-default is the right call. I verified the install premise too — `capsule-emit` (0.1.1) and `agent-action-capsule` (0.0.3) are live on PyPI, so `pip install -e examples/capsule-emit` resolves.

This is opt-in `examples/` code (standalone, not in the workspace/CI), so nothing here can break `main`. I'd just like three things tightened before merge — all small.

###  What's solid
- `CapsuleEmitTrust` correctly implements the `Trust` protocol (`score`/`attest`/`report`/`stake` match `layers/trust.py`) and is the plugin actually wired into the shipped scenario (`receipt_reputation_capsule.yaml` → `trust: capsule_emit`). The valid → corroborated → anchored gating is reasonable, and the fallback-to-heuristic path keeps it a true drop-in for stock scenarios.
- No secret leakage; clean separation; good docstrings.

###  The payments plugin's "drop-in layer" framing is inaccurate
`StripeCapsuledPayments` is described as a real `payments: stripe_capsule` layer, but it doesn't conform to the `Payments` protocol:

- Protocol: `pay(self, to: AgentId, amount: Money, ref: PaymentRef) -> Receipt`
- This PR: `pay(self, payer, payee, amount: float, currency="usd", **metadata) -> dict`

`quote`/`verify_payment`/`refund` diverge too (dicts/bools vs `Quote`/`PaymentStatus`/`Receipt`). `@runtime_checkable` only checks method *names*, so registration "passes," but a real `await payments.pay(to, amount, ref)` would bind `to→payer, amount→payee, ref→amount` and break. No scenario wires `stripe_capsule` (confirmed via grep), so this never surfaces — which means it's demo code, not a working layer.

**Ask:** either make it conform to `Payments`, or relabel it plainly as a standalone demo (drop the "drop-in layer" wording). Relabeling is the cheap, honest fix since nothing depends on it.

Also, on the real-Stripe path: `payee` is never used in the `PaymentIntent` (no destination/transfer), so the capsule "commits to payer/payee" while the actual charge enforces no payee — the dispute-proof story is weaker than the README implies. Worth a one-line caveat in the docstring.

###  Money rounding bug (real-Stripe path)
`_real_stripe_pay` uses `int(amount * 100)`, which truncates float error and **undercharges by a cent**: `$19.99 → 1998` ($19.98), `$0.29 → 28`, `$4.10 → 409`. One-line fix:

```diff
-    intent = stripe.PaymentIntent.create(
-        amount=int(amount * 100),
+    intent = stripe.PaymentIntent.create(
+        amount=round(amount * 100),
```

(Decimal would be even better for money, but `round()` removes the silent undercharge.)

###  No tests + a phantom sentinel test
- The test plan is all unchecked and the diff adds no test files.
- `trust.py` comments `# Sentinel test: tests/test_plugins.py::test_private_import_still_works`, but that test doesn't exist in the branch. So the deliberate coupling to private `nest_plugins_reference.trust.agent_receipts` internals (`_verify_receipt`, `_counterparty`, `_effective_receipts`, `_normalize`, `_raw_reputation`, …) has no guard despite the comment claiming one — any refactor of `agent_receipts` silently breaks this.

**Ask:** add a smoke test (instantiate both plugins, run one `report`/`score` and one sandbox `pay`), and either add the referenced sentinel test or remove the comment.

StevenMih:
@Skyrider3 — thank you, this is a genuinely useful review, and you're right on all three. Addressing each:

Payments framing. Agreed — stripe_capsule doesn't conform to the Payments protocol, and since nothing
wires it, I'll take the honest path and relabel it as a standalone demo (dropping the "drop-in layer"
wording in the README and docstrings) rather than imply a working layer. 

Good catch on the @runtime_checkable name-only check masking the signature mismatch. I'll also add the caveat you flagged: on the real-Stripe path payee isn't enforced in the PaymentIntent (no destination/transfer), so the capsule commits to payer/payee but the charge doesn't — worth being explicit about in the docstring.

Rounding. Fixing — int(amount*100) → round(amount*100) so we stop silently undercharging by a cent.
(Will note Decimal as the better long-term move.)

Tests + the phantom sentinel. Fair — the test_private_import_still_works comment references a test that
isn't there, which is worse than no comment. I'll add a real smoke test (instantiate both plugins, one
report/score and one sandbox pay) and either land the referenced sentinel test guarding the
agent_receipts coupling or drop the comment.

The trust plugin being the load-bearing piece (and the one actually wired into the scenario) matches how I think
about it too — appreciate you separating that from the payments demo. I'll push the changes and re-request your
review. Thanks again for the careful read.

StevenMih:
@Skyrider3 - I've updated the PR per your review. Re-requesting your review.

StevenMih:
closing this PR #32, re-opened with PR #54:

PR #35, opened 2026-06-23
[Hackathon] provenance-engineer: content-addressed datafacts with provenance chains & signed freshness proofs

Ngoga-Musagi:
## Problem

Problem 08 (datafacts). The default `datafacts_v1` plugin addresses datasets by **publisher-chosen name** (`df://<name>`), grants access unconditionally, and "proves" freshness with a bare `time.time()` check. That makes three classes of provenance attack silently undetectable in any trace:

- **Substitution** — republish *different* bytes under an existing name.
- **Stale-freshness** — claim a dataset is current by re-touching a timestamp without re-publishing.
- **Broken provenance** — a derived dataset whose ancestry references a hash nobody ever published.

## Design

A new `content_addressed` datafacts plugin where **the address is the content** and **freshness is a signature, not a timestamp**:

- `publish()` returns `df://sha256-<hex>` over the canonical content (volatile timestamps excluded so re-publishing identical content is idempotent). Different bytes ⇒ different address ⇒ **substitution is impossible by construction** — you cannot overwrite.
- `DatasetMetadata.parents` (new **optional** field, explicitly permitted by the problem) is folded into the child's content hash → a **Merkle-DAG**: a forged ancestor changes every descendant's address. `verify_provenance()` walks the DAG and rejects dangling parents; `publish()` raises `BrokenProvenanceError` on an unpublished parent.
- A `FreshnessProof` binds *(content-hash, tick, signer)* under a signature from the identity layer (`did_key`, taken via constructor — the same pattern as the EigenTrust PRs). A verifier validates it **without trusting wall-clock time**.

## API fit & tradeoffs

- Implements the `DataFacts` protocol **exactly** (`isinstance(plugin, DataFacts)` passes). The signed-proof machinery is added through **new** methods (`issue_freshness_proof`, `verify_freshness_proof`, `content_id`, `verify_provenance`) so `verify_freshness() -> bool` keeps the reference return type. The "proof object" the brief mentions is the new method, deliberately, so the plugin stays drop-in.
- Per-agent instances share one content-addressed **store** (substitution resistance is only meaningful swarm-wide) but each signs with its **own** identity — mirroring the merged `identity_rotation` provisioning pattern. Capability-gated so the same scenario degrades honestly under `datafacts_v1`.

## Adversarial validators (the charter's bar)

Three validators for scenario type `provenance_supply_chain`, each catching one attack class. From the **same** scenario YAML they **FAIL against `datafacts_v1`** and **PASS against `content_addressed`**:

| Validator | Catches |
|---|---|
| `validate_datafacts_content_addressing` | substitution / name-addressing |
| `validate_datafacts_freshness_proofs` | stale, unsigned freshness claims |
| `validate_datafacts_provenance_chain` | dangling ancestry, unverified chains |

## Verify

```bash
# Adversarial discrimination — same YAML, two plugins:
uv run nest run scenarios/provenance_supply_chain.yaml -o ./traces/prov.jsonl
  [print('PASS' if r.passed else 'FAIL', r.name, '—', r.detail) \
   for r in validate_trace(Path('traces/prov.jsonl'),'provenance_supply_chain')]"
# -> 3x PASS on content_addressed; flip the YAML to `datafacts: datafacts_v1` -> 3x FAIL
```

## Persona — `provenance-engineer`

The persona is in the code, not the label: content addressing as substitution resistance, parents folded into the hash to make the provenance DAG tamper-evident, a freshness proof that binds the content hash under a signature, and a validator that anchors freshness to *externally observed* ticks rather than attacker-claimed ones.

## Determinism & secrets

Pure-Python (`hashlib` / `json` + simulation-crypto `did_key`), **no new dependencies, no secrets/API keys**. Logical-tick clock, seeded only — `test_trace_is_byte_identical_under_replay` asserts byte-identical traces, and the validators pass under seeds 42 / 7 / 1337. Tier 1 fully deterministic.

**Local CI:** all five gates green — `ruff check`, `ruff format --check`, `pyright` (0 errors), `pytest` (588 passed, 1 skipped).

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -37,6 +37,9 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("memory", "lww_register"): f"{_REF}.memory.lww_register:LwwRegisterMemory",
     ("privacy", "noop"): f"{_REF}.privacy.noop:NoopPrivacy",
     ("datafacts", "datafacts_v1"): f"{_REF}.datafacts.datafacts_v1:DataFactsV1",
+    ("datafacts", "content_addressed"): (
+        f"{_REF}.datafacts.content_addressed:ContentAddressedDataFacts"
+    ),
 }
 
 
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -97,3 +97,9 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("receipt_reputation", receipt_reputation_factory)
+    elif name == "provenance_supply_chain":
+        from nest_core.scenarios_builtin.provenance_supply_chain import (
+            provenance_supply_chain_factory,
+        )
+
+        register_scenario("provenance_supply_chain", provenance_supply_chain_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/provenance_supply_chain.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/provenance_supply_chain.py
@@ -1,0 +1,475 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Content-addressed supply-chain provenance scenario with byzantine forgers.
+
+Persona: ``provenance-engineer``. This is the supply-chain scenario re-cast as a
+*provenance* test. Four honest hops (supplier -> manufacturer -> distributor ->
+retailer) each publish a content-addressed dataset derived from the previous
+hop's, building an end-to-end provenance DAG that the retailer verifies. A
+fraction of byzantine agents attempt the three attacks the
+``content_addressed`` plugin is built to defeat:
+
+* **substitution** — publish *different* content while trying to occupy an
+  existing dataset's address;
+* **stale-freshness** — claim a dataset is current without re-issuing a signed
+  freshness proof;
+* **broken provenance** — publish a derived dataset whose ancestry references a
+  hash nobody ever published.
+
+Every publish, freshness attestation, verification, and attack is emitted into
+the trace in a ``:``-delimited line protocol that the ``provenance_supply_chain``
+validators parse (see
+``nest_core.validators.validate_datafacts_*``). Behaviour is *capability-gated*
+via ``hasattr``: run the same scenario with ``datafacts: datafacts_v1`` and the
+hops fall back to name-addressed publishes with no signed proofs and no
+provenance checks — the validators then fail, which is the honest demonstration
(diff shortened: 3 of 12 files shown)


dhve:
Thanks @Ngoga-Musagi — this is real work on Problem 08 and the design instincts are good. The `content_addressed` plugin, the three-attack framing (substitution, stale-freshness, broken-provenance), the signed freshness proofs anchored to logical ticks — all solid.

Unfortunately this landed at the same time as #31 (also Problem 08, content-addressed datafacts with provenance and signed freshness), and #31 was merged a couple of hours ago. The two PRs create the exact same scenario file at the same paths (`packages/nest-core/nest_core/scenarios_builtin/provenance_supply_chain.py` and `scenarios/provenance_supply_chain.yaml`), so they can't coexist as-is — one has to be the canonical version and #31 got there first.

Closing this one as a duplicate. Not a reflection on the work itself.

**Two pieces of your work look like clean follow-up PRs worth landing independently:**

**1. First-class `parents: list[DataFactsUrl]` on `DatasetMetadata`.**

Your `types.py` change adding `parents` as a typed field on the model is a genuine API improvement over #31's `metadata["parents"]` dict-based approach. Cleaner, statically typed, more discoverable. If you want to open a small focused PR that:

- Adds the `parents` field to `DatasetMetadata` (your existing diff)
- Updates `nest_plugins_reference/datafacts/cid_facts.py` (the merged plugin from #31) to read from `metadata.parents` instead of `metadata.metadata["parents"]`, falling back to the dict for backward compat

…that would be a real quality-of-life win for the datafacts layer and I'd fast-track it.

**2. Linear supply-chain scenario as a separate scenario.**

Your supplier → manufacturer → distributor → retailer chain is a different testing angle from Ricardo's diamond (`source → {refine-a, refine-b} → aggregate → verify`). Different scenario name — `provenance_supply_chain_linear` or similar — different YAML file, different validators registry key. Both scenarios exercising the same plugin from different angles is genuinely useful. This one could be a scoped-down PR that only touches:

- `packages/nest-core/nest_core/scenarios_builtin/provenance_supply_chain_linear.py` (renamed from your current file)
- `scenarios/provenance_supply_chain_linear.yaml` (renamed)
- Registry entries in `plugins.py`, `scenarios.py`, `test_validators.py` for the new scenario name only (leave `provenance_supply_chain` untouched — that's #31's)

That would also be a fast-track review.

Not asking you to do both — either one alone would be a real contribution and land cleanly. If you want to demo at the Nanda Summit on July 11, either of those PRs open by July 5 is more than enough runway to be merge-ready.

Closing now to keep the queue clean. Ping on the hackathon channel if either of those directions is unclear.

PR #36, opened 2026-06-24
Add sealedbox privacy plugin (real authenticated encryption)

rahulsharma-sys:
## Summary

The default `noop` privacy plugin is a passthrough — `encrypt`/`decrypt` return the input unchanged and proofs always verify. This PR adds **`sealedbox`**, a real implementation of the Privacy layer: multi-recipient sealed-box encryption with a fresh ChaCha20-Poly1305 content key per message, wrapped to each recipient via X25519 ECDH + HKDF-SHA256. The Privacy layer now provides genuine confidentiality (non-audience agents can't read a payload; traces carry only ciphertext) and AEAD tamper-evidence (a modified ciphertext fails to decrypt rather than silently yielding altered bytes).

Design notes that keep it aligned with NANDA Town:

- **Replay-friendly by default.** Keys/nonces derive deterministically from `(seed, agent_id)`, so traces stay byte-for-byte reproducible — consistent with the framework's replay-first identity model. This is protocol/trace-level confidentiality, not secrecy against a holder of the shared seed (documented as such). `deterministic=False` switches to random per-agent keys for genuine secrecy at the cost of reproducibility.
- **Honest stubs.** `prove`/`verify_proof` raise `NotImplementedError` rather than reproducing `noop`'s forged always-valid proof. ZK proofs are noted as future work.
- **Registered** in the plugin registry as `("privacy", "sealedbox")`, so `nest plugins list` shows it and any scenario can select `privacy: sealedbox`.

Files: `nest_plugins_reference/privacy/sealedbox.py` (plugin), one line in `nest_core/plugins.py` (registry), `tests/test_sealedbox.py` (tests).

## Test Plan

- `make ci-local` passes — ruff, ruff format --check, pyright (strict), pytest (549 passed, 1 skipped).
- 8 new tests: encrypt/decrypt round-trip, multi-recipient (each opens, outsider rejected), tamper detection, malformed envelope, honest `NotImplementedError` stubs, deterministic replay (identical bytes for same seed/sender/counter), random mode (non-reproducible but round-trips), and registry discovery of `sealedbox`.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -36,6 +36,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("memory", "blackboard"): f"{_REF}.memory.blackboard:Blackboard",
     ("memory", "lww_register"): f"{_REF}.memory.lww_register:LwwRegisterMemory",
     ("privacy", "noop"): f"{_REF}.privacy.noop:NoopPrivacy",
+    ("privacy", "sealedbox"): f"{_REF}.privacy.sealedbox:SealedBoxPrivacy",
     ("datafacts", "datafacts_v1"): f"{_REF}.datafacts.datafacts_v1:DataFactsV1",
 }
 
--- a/packages/nest-plugins-reference/nest_plugins_reference/privacy/sealedbox.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/privacy/sealedbox.py
@@ -1,36 +1,98 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Real authenticated encryption for the Privacy layer (a sealed-box alternative
-to the ``noop`` passthrough).
+"""Nonce-misuse-resistant hybrid encryption for the Privacy layer.
 
-Multi-recipient sealed-box encryption: a fresh ChaCha20-Poly1305 content key per
-message, wrapped to each recipient via X25519 ECDH + HKDF-SHA256. This gives the
-Privacy layer genuine confidentiality (non-audience agents cannot read a payload
-through the API; the trace carries only ciphertext) and tamper-evidence (a
-modified ciphertext fails to decrypt rather than silently yielding altered bytes).
+``sealedbox`` is a full Problem 09 privacy plugin (hybrid encryption + selective
+disclosure + broadcast revocation), but its **load-bearing contribution** is a
+property the merged ``hybrid_x25519`` plugin openly concedes it lacks:
+**deterministic mode that is safe under key/nonce reuse.**
 
-Keys derive deterministically from ``(seed, agent_id)`` by default so simulation
-traces replay byte-for-byte, consistent with NandaTown's replay-first identity
-model. That is protocol/trace-level confidentiality, *not* secrecy against a
-holder of the shared seed. Pass ``deterministic=False`` for random per-agent keys
-(genuine secrecy against a seed-holder, at the cost of non-reproducible traces).
+Why that matters
+----------------
 
-``prove`` / ``verify_proof`` are honest stubs that raise ``NotImplementedError``
-— unlike the ``noop`` plugin's forged proof that always verifies. Zero-knowledge
-proofs are future work.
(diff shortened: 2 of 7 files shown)


dhve:
Thanks @rahulsharma-sys. The code quality here is genuinely good: the versioned binary envelope with strict truncation checks in `_parse_envelope`, the clean error taxonomy (`NotInAudienceError` / `TamperError` / `MalformedEnvelopeError`), and the docstring's honest framing of deterministic mode as trace-level confidentiality rather than secrecy against a seed-holder all show real care. Deriving keys from `(seed, agent_id)` by default so traces replay byte-for-byte is exactly the right instinct for this codebase.

I have to be upfront about the biggest issue before anything else, though. Not ready to merge as-is. Four things:

**1. This re-implements the already-merged `hybrid_x25519` plugin (PR #28), and delivers a subset of it.**

Look at the line directly below the one you added in `nest_core/plugins.py`: `("privacy", "hybrid_x25519")`. That plugin, merged from #28, uses the identical construction: fresh ChaCha20-Poly1305 content key per message, wrapped per recipient via X25519 ephemeral-static ECDH + HKDF-SHA256, with a deterministic seed-derived mode gated on a monotonic counter. On top of that it ships the two Problem 09 success criteria sealedbox lacks: a real selective-disclosure `prove`/`verify_proof` (salted Merkle commitment) and epoch-based broadcast revocation, plus replay protection by binding a `msg_id` into the AEAD associated data. Your `NotImplementedError` stubs are more honest than `noop`'s forged proofs, but the problem statement requires a real disclosure flow, so as submitted this meets neither the novelty bar (charter: no re-issuing existing work) nor the problem's own success criteria.

**2. The two mandatory charter deliverables are missing.**

Every submission needs an adversarial validator that catches an attack class the reference plugin would fail, plus a scenario YAML under `scenarios/` demonstrating your plugin passing under it. This PR ships plugin + unit tests only, so the judge panel has nothing to run it against.

**3. Deterministic mode has a real nonce-reuse hazard.**

In `_seal`, the content key and payload nonce derive from `(seed, sender, counter)` only, and `self._counter` lives in memory on the instance. Any run that reconstructs the plugin for the same agent (restart, or two instances in one scenario) resets the counter to 0 and reuses the same key/nonce pair on different plaintexts, which for ChaCha20-Poly1305 leaks the XOR of the plaintexts and enables forgeries. Your docstring flags that the counter must be unique, but nothing enforces it. Concrete fix: fold a digest of the plaintext and sorted audience into the HKDF info for the cek and nonce (SIV-style), so a counter collision degrades to revealing message equality instead of breaking the AEAD. Binding the envelope header as associated data instead of passing `None` would also harden the format.

**4. Registration is half-wired.**

`hybrid_x25519` and `trust_gated` both register in two places: the `_BUILTINS` dict (which you did) and the `[project.entry-points."nest.plugins.privacy"]` table in `packages/nest-plugins-reference/pyproject.toml` (which this PR does not touch). The rubric's api_fit dimension checks the entry point.

**Suggested path forward:**

1. Given #28 is merged, the highest-value pivot is the validator side of the privacy layer: build an adversarial validator targeting something `hybrid_x25519` openly concedes (its docstring states pre-revocation traffic stays readable forever; a forward-secrecy validator, or a nonce-reuse / trace-plaintext-leak detector, would be genuinely new), plus the scenario YAML. That satisfies the charter without re-issuing merged work.
2. If you keep sealedbox instead, state the differentiator vs `hybrid_x25519` in the PR body, make it load-bearing in tests, and add the selective-disclosure and revocation pieces Problem 09 requires.
3. Either way: fix the deterministic key/nonce derivation per point 3, and add the pyproject entry point.
4. Also note CI has not reported on this branch yet, and the branch pattern is `hackathon/<handle>-<theme>` (your handle is `rahulsharma-sys`, the branch says `rahul`). Run `make ci-local` and confirm green before requesting re-review.

Leaving this open. Ping me on the hackathon channel if the pivot options are unclear.

rahulsharma-sys:
Superseded by #176. Renaming the branch to match my charter handle (`rahul` → `rahulsharma-sys`) auto-closed this cross-repo PR — the reworked plugin, addressing all four points with `make ci-local` green, is there. @dhve re-review appreciated when you have a moment.

PR #39, opened 2026-06-25
[Hackathon] hiten: Layer 12 DataFacts — content-addressed provenance …

Hiten0305l:
## Problem

The existing DataFacts layer could store and retrieve datasets, but it could not reliably answer a critical trust question:

> Where did this data come from, and can its complete lineage be verified?

Datasets were identified by publisher-chosen names, freshness relied on timestamps, and provenance relationships were limited. Agents could verify that data existed, but not reliably verify its origin, integrity, freshness, or derivation history.

### cid_facts
A content-addressed DataFacts implementation.

URL = SHA-256 of semantic dataset content: owner, description, schema_version, tags, checksum, size_bytes, access_tier, metadata, and normalized parent URLs — not the dataset name and not timestamps.
Same content resolves to the same address, different content resolves to a different address, and republishing identical content is idempotent.
Tags are normalized before hashing, so equivalent tag sets resolve to the same address regardless of order. Parent URLs are also normalized before hashing, so `[A,B]` and `[B,A]` resolve to the same address.

### Provenance DAGs

Provenance is modeled as a DAG rather than a simple chain.
Datasets declare parents through `metadata["parents"]`. A dataset may reference multiple parents, enabling joins and derived datasets with multiple upstream sources. Publishing rejects unknown parent references, so lineage can only be built from datasets already known to the system.

The supply-chain scenario was generalized into:
```text
N suppliers → M manufacturers → distributor → retailer
```
allowing larger provenance graphs without code changes.

### Freshness Proofs
Freshness is a signed proof rather than a timestamp check.
Each publish operation signs:
```text
(dataset_url, logical_tick)
```
using the publisher's identity key. Verification only accepts proofs signed by the dataset's declared owner. No wall-clock time is used, preserving deterministic Tier-1 execution.

### Full Lineage Verification
Retailers verify the complete provenance graph rather than only immediate parents.
Verification uses a full breadth-first traversal (BFS), visits every reachable ancestor, deduplicates shared ancestors, validates freshness proofs throughout the lineage, and detects broken provenance chains. A first-parent-only walk would miss part of the graph; the multi-parent supply-chain DAG makes full traversal load-bearing.

### Provenance Audit Events
After successful verification, retailers emit:
```text
provenance_audit:<url>:nodes=<count>:ancestors=<urls>
```
This gives the trace an explainable record of exactly what lineage was discovered and verified.

## Security & Validation

The implementation targets the three main DataFacts failure modes: substitution, stale freshness claims, and broken provenance.

Validator tests use synthetic traces to prove the difference between weak traces and content-addressed provenance traces:

* name-based or substituted URLs fail substitution checks
* traces without valid signed freshness proofs fail freshness validation
* derived datasets with unpublished parents fail provenance-chain validation

Provenance tests exercise the real supply-chain DAG: suppliers feed manufacturers, manufacturers feed the distributor, and the retailer verifies the final dataset by walking the full lineage. Shared ancestors are intentionally exercised so the traversal proves they are visited exactly once.
Scenario/factory tests verify configurable topology: different supplier/manufacturer counts can be created from config, manufacturers wait for all required supplier URLs, distributors wait for all required manufacturer URLs, and retailers emit `provenance_audit` after successful verification.

## Verification

```bash
uv sync
uv run ruff check .
uv run ruff format --check .
uv run pyright
uv run pytest -v
```

Results:

```text
58 passed in test_cid_facts.py
599 passed, 1 skipped, 1 deselected
```
## Future Work

* Graph fingerprints for entire provenance DAGs.
* Provenance risk scoring based on lineage quality.
* Multi-distributor supply-chain topologies.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -37,6 +37,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("memory", "lww_register"): f"{_REF}.memory.lww_register:LwwRegisterMemory",
     ("privacy", "noop"): f"{_REF}.privacy.noop:NoopPrivacy",
     ("datafacts", "datafacts_v1"): f"{_REF}.datafacts.datafacts_v1:DataFactsV1",
+    ("datafacts", "cid_facts"): f"{_REF}.datafacts.cid_facts:CidFacts",
 }
 
 
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -97,3 +97,9 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("receipt_reputation", receipt_reputation_factory)
+    elif name == "provenance_supply_chain":
+        from nest_core.scenarios_builtin.provenance_supply_chain import (
+            provenance_supply_chain_factory,
+        )
+
+        register_scenario("provenance_supply_chain", provenance_supply_chain_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/provenance_supply_chain.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/provenance_supply_chain.py
@@ -1,0 +1,563 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Provenance supply-chain scenario — content-addressed DataFacts end-to-end.
+
+Each participant in the supply-chain DAG publishes a content-addressed dataset
+via the `cid_facts` plugin and emits structured trace events that the
+`provenance_supply_chain` validators can inspect:
+
+* ``publish:df://sha256-<hex>:<agent_id>``
+* ``freshness_proof:df://sha256-<hex>:<tick>:<sig_hex>``
+* ``derived:df://sha256-<child>:df://sha256-<parent1>[,df://sha256-<parent2>]``
+* ``provenance_audit:<url>:nodes=<n>:ancestors=<url1>,<url2>,...``
+
+The topology is configurable via ``num_suppliers`` (default 2) and
+``num_manufacturers`` (default 2) in the task config block.  Agent IDs are
+``supplier-0 … supplier-(N-1)`` and ``manufacturer-0 … manufacturer-(M-1)``.
+
+Flow (default 2×2)::
+
+```
+supplier-0 and supplier-1
+  └─ each publishes raw-material datasets
+  └─ each forwards dataset URLs to all manufacturers
+
+manufacturer-0 and manufacturer-1
+  └─ each waits until ALL supplier URLs are received
+     for the same round/item
+  └─ publishes a manufactured-goods dataset with all supplier URLs as parents
+  └─ emits a derived trace event
+  └─ forwards the dataset URL to Distributor
+
+Distributor
+  └─ waits until ALL manufacturer URLs are received
+     for the same round/item
+  └─ publishes a shipment dataset with all manufacturer URLs as parents
+  └─ emits a derived trace event
+  └─ forwards the dataset URL to Retailer
+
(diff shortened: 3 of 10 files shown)


dhve:
Thanks @Hiten0305l — this is real work on Problem 08 and there are several genuinely good ideas in it that we can land separately even though the PR as a whole can't merge.

Unfortunately this collides directly with PR #31 (Ricardo, merged a couple of hours ago), which also solved Problem 08. Same plugin name (`cid_facts`), same scenario name (`provenance_supply_chain`), same file paths:

- `packages/nest-plugins-reference/nest_plugins_reference/datafacts/cid_facts.py`
- `packages/nest-core/nest_core/scenarios_builtin/provenance_supply_chain.py`
- `scenarios/provenance_supply_chain.yaml`

The two implementations can't coexist since they occupy the same slot in the plugin registry. #31 landed first, so I'm closing this one — not a reflection on the quality of your work.

**Four pieces of your submission look like clean focused follow-up PRs worth landing independently:**

**1. Typed `FreshnessProof` model class in `types.py`.**

Your addition of `FreshnessProof(BaseModel)` with `url`, `publisher`, `tick`, `signature` fields is a proper typed replacement for the dict-based proof shape #31 uses internally. Statically typed, discoverable via the API, and makes verifier code cleaner. Small focused PR: add the model + update the merged `cid_facts.py` to use it. Fast-track review.

**2. First-class `parents: list[DataFactsUrl]` field on `DatasetMetadata`.**

PR #35 (Ngoga, closed as duplicate of #31) also proposed this. Both of you independently arrived at the same conclusion, which is a strong signal it's the right design. #31 uses `metadata["parents"]` dict-embedded, which is untyped and undiscoverable. A typed field is a real improvement. One PR from either of you (with a note to the merged `cid_facts.py` to read from `metadata.parents` with backward-compat fallback to the dict) would land clean. Worth coordinating with @Ngoga-Musagi so you don't submit the same thing twice.

**3. `agent_id` property and `has_peer_key(agent)` helper on `DidKeyIdentity`.**

The `did_key.py` change is genuinely useful outside the scope of this PR. `has_peer_key` is a much cleaner test than "try to sign and catch". `agent_id` as a property is the sort of thing a lot of downstream code needs. Tiny focused PR to `packages/nest-plugins-reference/nest_plugins_reference/identity/did_key.py` plus a test for each helper would land cleanly.

**4. Configurable N×M supply-chain scenario shape.**

Your `num_suppliers × num_manufacturers → distributor → retailer` topology is a genuinely different testing angle from Ricardo's diamond (`source → {refine-a, refine-b} → aggregate → verify`). Both scenarios exercising the same plugin from different graph shapes strengthens coverage. Scoped-down PR: rename to `provenance_supply_chain_multi` (or similar), give it its own YAML at the new name, register the new key in the VALIDATORS registry alongside `provenance_supply_chain`. That way both scenarios coexist and #31's diamond stays as-is.

Not asking you to do all four — any one alone is a real contribution and lands cleanly. If you want a demo-ready piece for the Nanda Summit on July 11, the `FreshnessProof` types PR or the `did_key` helpers PR are the fastest to green CI. The N×M scenario is a bigger piece but stands well on its own.

Closing now to keep the queue clean. Ping me on the hackathon channel if any of these directions is unclear, and please coordinate with @Ngoga-Musagi if you both intend to submit the `parents` field PR.

Hiten0305l:
Thanks for the detailed review. I understand why the full PR could not merge after #31 landed.

I’ll split the work into smaller focused follow-up PRs. I’ll start with the typed FreshnessProof model since it looks like the smallest clean piece, then I can follow with the did_key helpers and the N×M supply-chain scenario as a separate provenance_supply_chain_multi scenario.

I’ll also coordinate before touching DatasetMetadata.parents so it does not duplicate work from @Ngoga-Musagi.

PR #40, opened 2026-06-25
platform: audit remediation (trace schema, CI, ADRs, telemetry)

abhishekeb211:
## Summary
- Enterprise audit doc + simulation plugin UserWarnings (jwt, did_key, noop)
- Trace schema v1: trace_header with schema_version, JSON Schema docs (ADR-003)
- Windows scripts/ci-local.ps1; CI matrix Python 3.12/3.13; CycloneDX SBOM job
- Formal ADRs (001-005) + RFC-001 network transport
- LLM token telemetry (nest_shell.telemetry, judge log_judge_usage)
- Audit doc refresh with 552-test baseline and gap table updates

## Test plan
- [x] ci-local.ps1 — 552 passed locally
- [x] nest doctor 7/7
- [x] nest run marketplace — first JSONL line is trace_header schema 1.0
- [x] GitHub Actions green on 3.12 + 3.13
- [x] SBOM artifact produced

## Out of scope (60-90 day)
- TCP/gRPC transport implementation (follow-up PR)
- OpenTelemetry / production observability
- SOC2 / formal compliance

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,3 +76,17 @@ jobs:
           name: sbom-cyclonedx
           path: sbom.json
           if-no-files-found: error
+
+  sast:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v4
+        with:
+          version: "latest"
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - run: uv sync
+      - name: Bandit SAST (packages + scripts)
+        run: uv run bandit -r packages scripts -ll -q
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,9 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.12", "3.13"]
     steps:
       - uses: actions/checkout@v4
       - uses: astral-sh/setup-uv@v4
@@ -17,13 +20,16 @@ jobs:
           version: "latest"
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.12"
+          python-version: ${{ matrix.python-version }}
       - run: uv sync
       - run: uv run ruff check .
       - run: uv run ruff format --check .
 
   typecheck:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.12", "3.13"]
     steps:
       - uses: actions/checkout@v4
       - uses: astral-sh/setup-uv@v4
@@ -31,11 +37,27 @@ jobs:
           version: "latest"
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.12"
+          python-version: ${{ matrix.python-version }}
       - run: uv sync
       - run: uv run pyright
 
   test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.12", "3.13"]
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v4
+        with:
+          version: "latest"
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - run: uv sync
+      - run: uv run pytest -v
+
+  sbom:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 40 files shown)


dhve:
Thanks @abhishekeb211. There is real engineering care in here: `scripts/ci-local.ps1` is a faithful Windows mirror of `make ci-local` and a genuinely useful gap-filler, the ADR write-ups show you actually understood the codebase's decisions (ADR-004 correctly captures the zero-latency logical-time behavior of the default transport), and `log_judge_usage` reading `input_tokens`/`prompt_tokens` off either provider's response shape is a tidy provider-agnostic touch.

Honest verdict though: as a hackathon submission this is off-brief, and as a platform PR it bundles too many unrelated workstreams to review or merge as one unit mid-hackathon. Five things to work through:

**1. This does not map to the charter, and the scope crosses lines the charter draws.**

The charter asks for exactly one problem from `docs/hackathon/problems/`, one PR per participant, containing a plugin or validator plus a mandatory adversarial validator plus a scenario YAML, on a branch named `hackathon/<handle>-<short-theme>`. This PR maps to none of the ten problems, carries no adversarial validator and no scenario YAML, and `platform/audit-remediation` does not match the branch pattern the charter auto-closes on. It also touches CI, `nest-core` internals, four reference plugin families, and `scripts/judge`, which the charter puts out of bounds (do not modify code outside the layer/scenario/validator you are working on). You also have #145 open, which is similarly platform-shaped. Which one is your hackathon entry? Side note: marking all five new ADRs `Status: Accepted` is a call maintainers make, not something a PR can self-declare.

**2. You removed the `authenticated` comms entry point.**

In `packages/nest-plugins-reference/pyproject.toml` the diff replaces the `[project.entry-points."nest.plugins.comms"]` table (which registers `AuthenticatedComms`) with your new transport table instead of adding a second table. The `_BUILTINS` fallback in `nest_core/plugins.py` masks it at runtime, but the documented install-time discovery path for `authenticated` is gone. I assume this was an accidental replace; both tables should coexist.

**3. The judge token budget is never enforced in the real flow.**

In `scripts/judge/limits.py`, `record_tokens` raises `JudgeLimitError` when `tokens_used` exceeds the budget, but the only production caller (`log_judge_usage` in `judge_pr.py`) wraps it in `contextlib.suppress(JudgeLimitError)`, and `before_call` only checks the call count. So with `NEST_JUDGE_TOKEN_BUDGET` set and `NEST_JUDGE_MAX_CALLS` unset, nothing is ever blocked. The unit test passes because `test_token_budget_blocks` calls `record_tokens` directly rather than going through the judge path. Fix: check `tokens_used` against the budget inside `before_call`, let `record_tokens` just accumulate, and drop the suppress. (`complete_call` is also defined but never called from anywhere.) Bigger picture: `scripts/judge` is the scoring instrument for the event you are competing in, so even with the bug fixed, this piece should land as an organizer-owned change, not inside a participant PR.

**4. The trace header breaks a shared format in the middle of the hackathon.**

Versioning traces is a good idea, and `docs/trace-schema.json` plus ADR-003 are solid write-ups. But `TraceWriter._write_header` now emits a `trace_header` line at the top of every trace, and you had to patch `test_scenario.py` and `test_sim.py` to skip it. That is exactly the breakage every open PR with per-line trace assertions will hit on rebase, and there are roughly ninety open PRs right now. Also, `analyze_trace` in `inspect.py` now counts the header in `total_events` and in the kind counter, so `nest inspect` output changes for every trace. This workstream needs its own PR and organizer sign-off on timing, probably opt-in until the hackathon window closes.

**5. `tcp_loopback` is TCP in name only.**

In `tcp_loopback.py`, the socket does a one-shot registration handshake and is then closed; `send` goes through `TcpLoopbackHub.deliver`, which puts straight onto an in-process `asyncio.Queue`. Message bytes never traverse TCP, yet every test run now binds a real loopback socket. And nothing here wires it into a run: no scenario or factory constructs a `TcpLoopbackHub`, and `test_plugin_registry_resolves` only checks `resolve()`, not construction from a YAML. Your own PR body lists TCP transport implementation as a follow-up, which I agree with, so shipping a half-TCP plugin now cuts against your own scoping. I would keep it inside RFC-001 (correctly marked Draft) until a transport that actually routes over the wire exists.

One more note: no CI checks have run on this PR in this repo, so the "GitHub Actions green" line in your test plan presumably refers to your fork. The judge panel does not score broken PRs, so getting checks green here is what counts.

**Suggested path forward:**

1. State on both #40 and #145 which one (if either) is your hackathon submission; the charter is one problem, one PR.
2. Split the platform work into independent PRs proposed through issues so maintainers can sequence them. `ci-local.ps1` on its own is the easiest piece to review; the trace schema and judge changes need discussion first.
3. Restore the `nest.plugins.comms` entry-point table alongside the new transport one.
4. If organizers pick up the judge budget idea, fix the enforcement gap from point 3.
5. For a competing entry: one problem, branch `hackathon/abhishekeb211-<theme>`, plugin plus adversarial validator plus scenario YAML.

Leaving this open. Ping me on the hackathon channel if you want help carving this up.

PR #48, opened 2026-07-01
[Hackathon] hiten: add typed FreshnessProof model

Hiten0305l:
## Summary
This is a focused follow-up to my previous hackathon PR (#39), implementing the reviewer-suggested `FreshnessProof` API improvement after the DataFacts work from #31 was merged.
This PR promotes `FreshnessProof` from a plugin-local implementation in `cid_facts.py` to a first-class shared type in `nest_core.types` and updates the merged `cid_facts` implementation to use the shared model.

## What
- Add `FreshnessProof` as a shared model in `nest_core.types`.
- Update the merged `cid_facts` implementation to use the shared model instead of maintaining a plugin-local definition.
- Update the existing tests to import and validate the shared model from `nest_core.types`.

## Why
Previously, `FreshnessProof` was defined only inside `cid_facts.py`, making it an implementation detail of a single plugin.
Promoting it to `nest_core.types` makes it:
- a first-class public API type,
- statically typed and IDE-discoverable,
- reusable across plugins, validators, scenarios, and future extensions, and
- simpler for verifier implementations to consume consistently.

This follows the reviewer suggestion to promote `FreshnessProof` into a shared typed model while keeping the change small and focused.

## Scope
This PR is intentionally limited to the `FreshnessProof` API improvement. 

## Verification
- ✅ `make ci-local`
```bash
uv sync
uv run ruff check .
uv run ruff format --check .
uv run pyright
uv run pytest -v

### DIFF
--- a/packages/nest-core/nest_core/types.py
+++ b/packages/nest-core/nest_core/types.py
@@ -547,6 +547,31 @@ class AccessGrant(BaseModel):
     expires_at: float | None = None
 
 
+class FreshnessProof(BaseModel):
+    """Signed proof that a content-addressed dataset was (re)published at a given logical tick.
+
+    The ``signature`` covers a canonical serialization of the dataset URL and
+    logical tick, and must have been produced by the agent identified by
+    ``publisher``.
+    Proofs signed by anyone other than the dataset's declared owner are rejected
+    by ``CidFacts.verify_freshness``.
+
+    Example::
+
+        proof = FreshnessProof(
+            url=DataFactsUrl("df://sha256-abc123"),
+            publisher=AgentId("supplier-0"),
+            tick=3.0,
+            signature=Signature(signer=AgentId("supplier-0"), value=b"sig"),
+        )
+    """
+
+    url: DataFactsUrl
+    publisher: AgentId
+    tick: float
+    signature: Signature
+
+
 # ---------------------------------------------------------------------------
 # Transport capabilities
 # ---------------------------------------------------------------------------
--- a/packages/nest-plugins-reference/nest_plugins_reference/datafacts/cid_facts.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/datafacts/cid_facts.py
@@ -42,8 +42,13 @@ import hashlib
 import json
 from typing import TYPE_CHECKING, Any, cast
 
-from nest_core.types import AccessGrant, AgentId, DataFactsUrl, DatasetMetadata, Signature
-from pydantic import BaseModel
+from nest_core.types import (
+    AccessGrant,
+    AgentId,
+    DataFactsUrl,
+    DatasetMetadata,
+    FreshnessProof,
+)
 
 if TYPE_CHECKING:
     from nest_core.layers.identity import Identity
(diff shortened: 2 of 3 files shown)


dhve:
Thanks @Hiten0305l. This is exactly the follow-up I asked for when closing #39: the typed model lands in `nest_core.types`, the merged `cid_facts.py` consumes it, and re-importing `FreshnessProof` into `cid_facts` keeps the old import path working for downstream code. `TestFreshnessProofModel` pinning the `SharedClock` semantics (`ticks == [1.0, 2.0, 3.0, 4.0]`) is a nice touch, and CI is green across lint, typecheck, and test.

**1. `publisher` is attacker-controlled and never verified. Make it load-bearing.**

In `publish()` you set `publisher=dataset.owner`, but `_freshness_payload` still signs only `{"url", "tick"}`, and `verify_freshness` checks `proof.signature.signer` against `meta.owner` without ever reading `proof.publisher`. Two consequences:

- A valid proof can have its `publisher` swapped to any value and still verify, because the field sits outside the signed bytes.
- In the existing `test_forged_freshness_claim_is_rejected` scenario, the attacker republishes with `owner=AgentId("owner")`, so the forged proof stored in the shared registry now carries `publisher == "owner"` even though the attacker produced it. Verification correctly fails on the signer check, but any validator that reads `proof.publisher` as "who published this" (which is the whole point of promoting the type for reuse) gets the attacker's chosen answer.

The field was in my suggested shape in the #39 writeup, so this one is partly on me, but as implemented it is decorative. Fix: include the publisher in the signed payload (`{"url": ..., "publisher": ..., "tick": ...}` in `_freshness_payload`) and add `if proof.publisher != meta.owner: return False` to `verify_freshness`. Then add the test the model deserves: rebuild a valid proof with a different `publisher` over the same signature, store it, and assert `verify_freshness` returns False. As the code stands today that test would pass verification, which is the clearest demonstration the field is not yet doing anything. This is where the correctness and test rigor of the PR live.

**2. The `types.py` docstring promises more than the code enforces, and core now documents a plugin.**

"must have been produced by the agent identified by `publisher`" is not enforced anywhere (see item 1), and the sentence naming `CidFacts.verify_freshness` means a `nest_core` type is describing the behavior of one reference plugin. Fixing item 1 makes the first claim true; for the second, phrase it plugin-neutrally, e.g. "verifiers are expected to reject proofs whose signature was not produced by the dataset's declared owner." Also worth carrying over the content-binding rationale from the docstring you deleted in `cid_facts.py` (the proof binds to the content hash, not to a name), since that is the load-bearing design fact for anyone reusing this type.

One note for anyone else reviewing: #48, #49, and #60 are the follow-up split I explicitly suggested when closing #39, so the one-problem-one-PR rule is not in play here. Cross-linking the three PRs in each body would still help other reviewers see that at a glance.

**Suggested path forward:**

1. Bind `publisher` into `_freshness_payload` and reject `proof.publisher != meta.owner` in `verify_freshness`.
2. Add the spoofed-publisher rejection test, and confirm `test_forged_freshness_claim_is_rejected` still passes with the new payload shape.
3. Reword the `types.py` docstring plugin-neutrally and restore the content-binding note.
4. Re-run `make ci-local` and confirm all five stages green.

Leaving this open. Ping me on the hackathon channel once the publisher binding is in.

JamesCarnley:
Thank you, @Hiten0305l. A real freshness record must bind dataset identity, publisher/controller identity, observation time, validity basis, and signature in canonical bytes, then match the publisher to authoritative ownership; a signature alone does not prove freshness. Current Town's evidence records bind one observer, subject, claim, value, and logical time but make no dataset-currentness claim. In this frozen model, publisher is outside the signed payload and can be attacker-controlled, so the wrapper cannot provide the advertised trust. I am closing the legacy PR as a future-profile requirement and retaining publisher, URL, time, owner, key, window, and signature mutation cases. No contributor code was merged or executed during this review.

PR #49, opened 2026-07-01
[Hackathon] hiten: add DidKeyIdentity helper accessors

Hiten0305l:
## Summary
This is a focused follow-up to my previous hackathon PR (#39). Following the merge of the DataFacts implementation in #31, this PR contributes an independent identity-layer API improvement by adding two small helper accessors to `DidKeyIdentity`.
The change exposes the identity's agent ID through a public property and provides a helper for checking whether a peer's public key has been registered, without changing existing identity semantics.

## What
- Add the `agent_id` property.
- Add the `has_peer_key(agent)` helper.
- Add focused unit tests covering both helper APIs.

## Why
`DidKeyIdentity` already stores the identity's agent ID and maintains a registry of known peer public keys internally. Exposing these operations through dedicated helper APIs improves usability while preserving existing identity behavior.
The `agent_id` property provides a stable public accessor instead of requiring downstream code to depend on private implementation details.
The `has_peer_key(agent)` helper provides a clear, self-documenting way to determine whether a peer's public key has been registered, avoiding direct inspection of internal state and making downstream code more readable.

These additions are intentionally lightweight, reusable, and do not modify signing, verification, key generation, or identity semantics.

## Scope
This PR is intentionally limited to the `DidKeyIdentity` helper APIs and their associated unit tests.

## Verification
- ✅ `make ci-local`
```bash
uv sync
uv run ruff check .
uv run ruff format --check .
uv run pyright
uv run pytest -v
```

### DIFF
--- a/packages/nest-plugins-reference/nest_plugins_reference/identity/did_key.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/identity/did_key.py
@@ -55,6 +55,25 @@ class DidKeyIdentity:
         self._known_keys[agent_id] = _decode_public_key(public_key)
 
     @property
+    def agent_id(self) -> AgentId:
+        """This agent's identifier.
+
+        Example::
+
+            aid = ident.agent_id
+        """
+        return self._agent_id
+
+    def has_peer_key(self, agent: AgentId) -> bool:
+        """Return ``True`` if the given agent's public key is registered.
+
+        Example::
+
+            ident.has_peer_key(AgentId("a2"))
+        """
+        return agent in self._known_keys
+
+    @property
     def public_key(self) -> bytes:
         """This agent's public key.
 
--- a/packages/nest-plugins-reference/tests/test_plugins.py
+++ b/packages/nest-plugins-reference/tests/test_plugins.py
@@ -140,6 +140,25 @@ class TestDidKeyIdentity:
         with pytest.raises(ValueError, match="public keys only"):
             ident.register_peer(AgentId("a2"), ident.public_key, private_key=b"secret")
 
+    def test_agent_id(self) -> None:
+        from nest_plugins_reference.identity.did_key import DidKeyIdentity
+
+        ident = DidKeyIdentity(AgentId("a1"), seed=b"seed")
+        assert ident.agent_id == AgentId("a1")
+
+    def test_has_peer_key_self_and_unknown(self) -> None:
+        from nest_plugins_reference.identity.did_key import DidKeyIdentity
+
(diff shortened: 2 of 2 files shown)


dhve:
Thanks @Hiten0305l. The two accessors are clean and match the house style of `did_key.py` exactly: typed signatures, `Example::` docstrings, and `test_has_peer_key_self_and_unknown` correctly covers the construction-time self-registration in `__init__`, which tells me you actually read the class before extending it.

Honest verdict though: as a hackathon entry this is off-brief, and I want to save you from spending effort where the judges cannot score it. Three things to work through:

**1. This does not map to a hackathon problem, so there is nothing for the panel to score.**

The charter asks for exactly one problem from `docs/hackathon/problems/`, with the plugin/scenario/validator it specifies, plus an adversarial validator, plus a scenario YAML demonstrating it. Two helper accessors on the reference plugin meet none of those criteria. The identity problem on offer (05, Ed25519 key rotation) is also already merged as #25, so identity is a tough place to land novelty regardless.

**2. One problem, one PR.**

You have #48, #49, and #60 open, and the charter is explicit: single PR, single problem per participant. Please pick your primary (from a quick read, #60 is the most substantial and closest to the brief) and say on the others how they relate, or close them.

**3. The new surface is concrete-class-only and partly redundant.**

The `Identity` Protocol in `packages/nest-core/nest_core/layers/identity.py` defines only `sign`, `verify`, and `resolve`, so any caller typed against the Protocol never sees `agent_id` or `has_peer_key`. And `resolve()` already reports registration state (unknown agents come back with an empty `public_key`). Fine polish, but no new capability.

Also: no CI checks have reported on this branch yet, so I cannot confirm the `make ci-local` run from the body.

**Suggested path forward:**

1. Pick one primary PR as your hackathon submission and consolidate.
2. Make sure that primary has the adversarial validator and scenario YAML the charter requires.
3. If you still want these accessors, fold them into the primary only if a scenario actually uses them; standalone, this would be reviewed as ordinary maintenance, not scored as an entry.

Leaving this open. Ping me on the hackathon channel if the consolidation plan is unclear.

JamesCarnley:
Thank you, @Hiten0305l. The useful behavior is already available through current public operations: identity creation returns the agent ID and key material, is idempotent for the same name, and file/chain resolvers distinguish present from absent identities. The proposed agent_id and has_peer_key helpers target a retired concrete DidKeyIdentity rather than the protocol and would duplicate current reads. I am closing this legacy convenience PR as already covered. No contributor code was merged or executed during this review.

PR #50, opened 2026-07-01
datafacts: first-class typed parents field on DatasetMetadata (follow-up to #35)

Ngoga-Musagi:
Follow-up #1 to #35 (closed as a duplicate of #31), the typed-lineage-field piece @dhve flagged as a clean, independent improvement.

## What
Adds a first-class, statically typed provenance field to `DatasetMetadata`:

```python
parents: list[DataFactsUrl] = Field(default_factory=lambda: list[DataFactsUrl]())
```

…and updates the merged `cid_facts` plugin (#31) to read it, keeping the old dict spelling working:

```python
def parents_of(dataset: DatasetMetadata) -> list[DataFactsUrl]:
    if dataset.parents:                       # typed field wins
        return list(dataset.parents)
    raw = dataset.metadata.get("parents", []) # legacy fallback, backward compatible
    ...
```

## Why
#31 declares lineage via `metadata["parents"]` — a free-form, untyped, undiscoverable dict entry. A typed field is statically checkable, IDE-discoverable, and self-documenting. Both @Hiten0305l and I independently proposed it on our closed PRs, which @dhve noted is a strong signal it's the right design.

## Non-breaking by construction
- `content_hash()` is **untouched**. Because `publish()` still folds parents into the content hash via `metadata`, a dataset declared through the typed field and one declared through the dict **content-address to the identical URL** — verified by a new test.
- #31's entire 44-test `cid_facts` suite passes unchanged.

## Tests
`packages/nest-plugins-reference/tests/test_cid_facts_parents_field.py` (6 tests): default empty, typed field read, legacy dict still works, address parity between the two spellings, typed-takes-precedence, and phantom-parent rejection via the typed field.

## Verify
```
uv run ruff check . && uv run ruff format --check . && uv run pyright && uv run pytest -q
# 740 passed, 1 skipped
```

## Coordination
@dhve — on #39 you asked @Hiten0305l and me to coordinate on this so it isn't submitted twice. You listed it as item #1 on my #35, and it originated in my closed PR, so I'm carrying it here. Happy to adjust if you'd prefer otherwise.

PR #51, opened 2026-07-01
scenarios: provenance_supply_chain_linear — deep linear provenance chain (follow-up to #35)

Ngoga-Musagi:
Follow-up #2 to #35 (closed as a duplicate of #31), the distinct-topology scenario @dhve flagged as worth landing independently.

## What
A new `provenance_supply_chain_linear` scenario — a deep **linear** dataset chain:

```
supplier-0 -> manufacturer-0 -> distributor-0 -> retailer-0 -> auditor-0
```

Each producer declares exactly the previous hop's dataset as its single provenance parent, so the lineage is one unbranched spine four hops deep. The auditor walks the whole chain back to the supplier's root, then runs the three attacks (substitution, forged freshness, broken provenance) as an outsider.

## Why it complements #31
#31's `provenance_supply_chain` is a **diamond** — it stresses lineage *fan-out* (a join whose ancestry forks back to a shared root). This one stresses lineage *depth* (a long single-parent chain). Notably #31's own history shows the scenario started linear and moved to a diamond; this restores the linear angle as a first-class, separate scenario rather than losing it.

## Minimal and #31-safe
- Exercises the **merged `cid_facts` plugin** as-is — no plugin changes.
- Emits the **same `|`-delimited provenance wire protocol** as the diamond, so it reuses #31's four validators under a new registry key `provenance_supply_chain_linear` — **zero new validator code**, and `provenance_supply_chain` is left completely untouched.
- Touch points: new scenario module + YAML, one `scenarios.py` dispatch branch, one `VALIDATORS` key, one line in the registry test.

## Adversarial discrimination (same YAML, two plugins)
```
uv run nest run scenarios/provenance_supply_chain_linear.yaml
# cid_facts  -> 4x PASS (chain resolved to depth 4)
# datafacts_v1 -> chain_integrity PASS, substitution/freshness/provenance FAIL
```

## Determinism & tests
Tier 1, seeded only, logical-tick clock. `test_provenance_supply_chain_linear.py` asserts all-pass on `cid_facts`, the three adversarial FAILs on `datafacts_v1`, lineage depth == 4, and **byte-identical replay under seeds 42 / 7 / 1337**.

## Verify
```
uv run ruff check . && uv run ruff format --check . && uv run pyright && uv run pytest -q
# 740 passed, 1 skipped
```

Persona — provenance-engineer: content addressing as substitution resistance, lineage folded into the content hash so a deep chain is tamper-evident end to end, freshness bound under a signature, and a verifier anchored to externally observed ticks.

### DIFF
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -109,6 +109,12 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("provenance_supply_chain", provenance_supply_chain_factory)
+    elif name == "provenance_supply_chain_linear":
+        from nest_core.scenarios_builtin.provenance_supply_chain_linear import (
+            provenance_supply_chain_linear_factory,
+        )
+
+        register_scenario("provenance_supply_chain_linear", provenance_supply_chain_linear_factory)
     elif name == "bft_hotstuff":
         from nest_core.scenarios_builtin.bft_hotstuff import bft_hotstuff_factory
 
--- a/packages/nest-core/nest_core/scenarios_builtin/provenance_supply_chain_linear.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/provenance_supply_chain_linear.py
@@ -1,0 +1,323 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Content-addressed provenance scenario: a deep *linear* supply chain, then attacks.
+
+Where ``provenance_supply_chain`` (#31) is a **diamond** that stresses lineage
+*fan-out* (a join whose ancestry forks back to a shared root), this scenario is a
+**linear spine** that stresses lineage *depth* -- the canonical physical
+supply-chain shape::
+
+    supplier-0 -> manufacturer-0 -> distributor-0 -> retailer-0 -> auditor-0
+
+Each producer hop publishes a *dataset* through ``plugins["datafacts"]`` and
+declares exactly the previous hop's dataset as its single provenance parent, so
+the lineage is one unbranched chain four hops long. ``auditor-0``:
+
+1. **Verifies** -- walks the whole parent chain from the retailer's dataset back
+   to the supplier's root and reports how many distinct datasets it found. A
+   single broken hop anywhere along a deep chain must surface as a break.
+2. **Attacks** -- as an *outsider* (signing with its own identity, never the
+   supplier's), it runs the three provenance attacks against the root:
+
+   * *Substitution*: republish different content under the supplier's exact
+     ``name`` -- does it land on the supplier's URL?
+   * *Forged freshness*: republish identical content claiming the supplier as
+     ``owner`` -- does the supplier then read as freshly attested under an
+     outsider's signature?
+   * *Broken provenance*: publish a dataset whose declared parent was never
+     published -- is it rejected?
+
+Every step is reported as a ``|``-delimited trace message (``:`` collides with
+the ``df://`` URL scheme) using the *same* wire protocol as the diamond
+scenario, so the two share one validator set: point ``layers.datafacts`` at
+``cid_facts`` and every adversarial validator passes; point it at
+``datafacts_v1`` and they fail, because that reference plugin has no
+content-addressing, no signed freshness, and no provenance concept at all.
+
+Example::
+
+    agents = provenance_supply_chain_linear_factory(config, plugins)
+"""
+
+from __future__ import annotations
+
+from typing import Any, cast
+
+from nest_core.scenario import ScenarioConfig
+from nest_core.sim.agent import AgentContext, StateMachineAgent
+from nest_core.types import AgentId, DatasetMetadata
+
+# The producer hops of the linear chain, upstream-to-downstream. Kept as data so
+# the spine is easy to read and to lengthen; the auditor is appended separately.
+_CHAIN: tuple[str, ...] = ("supplier", "manufacturer", "distributor", "retailer")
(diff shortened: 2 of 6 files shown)


dhve:
Thanks @Ngoga-Musagi. This is exactly the scoped follow-up I asked for when closing #35: just the new module with its test file, the new YAML, and the three registry touch points (the `scenarios.py` dispatch branch, the `VALIDATORS` key, one line in `test_validators.py`), with `provenance_supply_chain` left completely untouched. The `_CHAIN` tuple keeping the spine as data, the byte-identical replay test across seeds 42/7/1337, and the plugin-swap discrimination (cid_facts all-pass, datafacts_v1 failing the three adversarial checks) are all the right shape, and the module docstring on fan-out vs depth is genuinely good reading.

**1. The depth claim is never actually pinned by the trace or the tests.**

`chain_ok|{leaf}|{len(seen)}` reports a distinct-node count, and that count is 4 in both scenarios: the diamond's `test_diamond_walk_visits_all_four_nodes` asserts `"4"` (source + refine-a + refine-b + aggregate), and your `test_linear_walk_visits_all_four_hops` asserts the same `"4"`. So the one thing this scenario exists to add (a long single-parent spine instead of a shallow fan-out) is invisible at the trace level; a diamond trace would satisfy your depth test verbatim. Suggested fix: track distance from the leaf inside `_verify_chain` (a `(url, dist)` stack is enough for an unbranched chain) and emit one extra message, say `chain_depth|{leaf_url}|{max_depth}`. The shared validators match exact prefixes (`chain_ok|`, `chain_broken|`, `attack_...`), so an extra message passes straight through them and validator reuse is unaffected. Then assert depth 4 in the test, with a comment noting the diamond's longest path is 3. That single assertion makes the scenario's reason for existing verifiable, which is exactly what test rigor and novelty scoring look at.

**2. Set docstrings aside and about 135 of the module's roughly 165 code lines are verbatim copies of the diamond module.**

`AuditorAgent` is `VerifyAndAttackAgent` with only the docstrings reworded, `ChainHopAgent` is `RefineAgent` with `aggregator` renamed to `downstream`, `SupplierAgent` is `SourceAgent` with the fan-out loop collapsed to a single send, and `_parents_of`, `_PHANTOM_PARENT`, and `_build_datafacts_handles` are identical code (only docstring wording differs). I suggested starting from your #35 file so I own part of this, but seeing it landed: both scenarios feed one shared validator set keyed to the same `|`-delimited protocol, so two private copies of the emitters can drift apart silently and quietly break the reuse this PR is built on. Once you add depth tracking (point 1), `AuditorAgent` earns its keep as genuinely different code. For the rest, import `SourceAgent`, `RefineAgent`, `_parents_of`, `_PHANTOM_PARENT`, and `_build_datafacts_handles` from `provenance_supply_chain` instead of copying them. No scenario module cross-imports today, but these two files are twins by design and this is the right first case. `SupplierAgent` then just becomes `SourceAgent(source_id, refiners=[downstream], ...)` in the factory.

Two smaller notes, neither blocking:

- "No checks reported" on this PR is on our side: the CI run for your head commit is sitting at action_required waiting for a maintainer to approve workflow runs. I will kick it off so the checks report against the branch.
- Good call writing parents via the `metadata["parents"]` dict spelling; it keeps this PR merge-order independent from your #50. Once #50 lands, switching both provenance scenarios to the typed `parents` field is a natural follow-up, not something to do here.

**Suggested path forward:**

1. Add depth tracking to the walk, emit `chain_depth|...`, and assert 4 in `test_linear_walk_visits_all_four_hops`.
2. Replace the copied helpers and producer agents with imports from `provenance_supply_chain`; keep `AuditorAgent` local since it now genuinely differs.
3. Re-run `uv run ruff check . && uv run ruff format --check . && uv run pyright && uv run pytest -q` and push; I will approve the workflow run so CI reports on the PR.

Leaving this open. Ping me on the hackathon channel if the import reshuffle collides with anything.

Ngoga-Musagi:
Thanks for the detailed review — both points addressed in 84a44ba.

**1. Depth is now pinned in the trace.** The auditor tracks each node's distance from the leaf during the walk (a `(url, dist)` stack) and emits `chain_depth|{leaf}|{max_depth}` — the longest root-path length, **4** for this spine — separate from `chain_ok`'s distinct-node count (4 in both scenarios). `test_linear_walk_visits_all_four_hops` now asserts depth 4, with a comment noting the diamond's longest path is only 3, so a diamond trace no longer satisfies it. The new prefix passes straight through the shared validators (they match exact prefixes) — confirmed: all four still PASS on `cid_facts` and the three adversarial ones FAIL on `datafacts_v1`.

**2. ~135 copied lines removed.** `SourceAgent`, `RefineAgent`, `_parents_of`, `_PHANTOM_PARENT`, and `_build_datafacts_handles` are now imported from `provenance_supply_chain` (supplier = `SourceAgent(refiners=[downstream])`, middle hops = `RefineAgent`). Only `AuditorAgent` stays local, since depth tracking now makes it genuinely different code — which also removes the second copy of the attack/`chain_*` emitters that could have drifted from the shared protocol. `provenance_supply_chain` is untouched.

JamesCarnley:
Thank you, @Ngoga-Musagi. A provenance-depth claim must be independently reconstructed from emitted parent links, reach the declared root through the configured number of distinct nodes, and fail on a broken, forged, cyclic, or missing link. Current Town evidence is flat and signed; it has no dataset object, parent graph, or claimed depth. Adding a provenance DAG would be a new data model, and the shared legacy validator does not itself bind the configured depth. I am closing this future-profile PR while retaining configurable-depth and single-link mutation cases for any later DAG implementation. No contributor code was merged or executed during this review.

PR #52, opened 2026-07-01
[Hackathon] varuni7: delegatable capability tokens with cascading, partition-tolerant revocation

varuni7:
# [Hackathon] varuni7: delegatable capability tokens with cascading, partition-tolerant revocation

**Problem 4 (auth).** The default `jwt` plugin cannot delegate a subset of a
capability, and revokes by exact token string with no parent/child relationship.
This ships `delegatable`: macaroon-style tokens where a holder mints a strictly
narrower, shorter-lived sub-token for another agent **offline** (no issuer
round-trip), and revoking a token severs its entire subtree.

## Design

- **Macaroon HMAC chain.** A child seal is `HMAC(parent_seal, caveat)`, so a
  holder can only *add* restrictions (subset scopes, shorter TTL, bound
  audience) — never widen them. `delegate()` also checks the scope subset and
  clamps child TTL ≤ parent.
- **Cascade by construction.** Revocation is keyed on the seal, not the token
  string. `verify()` recomputes the chain, and a revoked ancestor seal severs
  every descendant — no per-child revocation list.
- **Partition-tolerant revocation.** Revocation state is a grow-only set
  (G-Set CRDT). `merge` is the set union — commutative, associative,
  idempotent — so replicas converge by gossip regardless of order, duplication,
  or loss. Safety is monotone (a seal once observed revoked is never accepted
  again), and a verifier that has not yet heard a revocation still accepts until
  it merges one.
- **Determinism.** No `uuid`, no `time.time()`. Token identity is the seal (a
  pure function of content + secret); all time comes from an injected clock.
  Same seed → byte-identical trace.

## Attacks defeated (adversarial validator)

The `delegated_auth` scenario + validators PASS on `delegatable` and FAIL on
`jwt`. Three attack classes are blocked, each a typed error:

1. **Scope escalation** — child requests a scope the parent lacks → `ScopeEscalationError`.
2. **Stale / revoked parent** — delegate or verify through a revoked ancestor → `RevokedAncestorError`.
3. **Audience confusion** — token presented by an agent that is not its audience → `AudienceMismatchError`.

Plus a **convergence** check: after the revoked seal is gossiped, the revoked
intermediary can no longer mint tokens while its siblings still can — proof the
revocation propagated via CRDT merge, not just on the coordinator.

## Verify it

```bash
uv sync
nest run scenarios/delegated_auth.yaml -o traces/delegated_auth.jsonl
python -c "
from pathlib import Path
for r in validate_trace(Path('traces/delegated_auth.jsonl'), 'delegated_auth'):
"
# Swap `auth: delegatable` -> `auth: jwt` in the YAML and the same four validators FAIL.
```

## Tests

- 51 new tests: example-based + `hypothesis` property tests (scope-subset,
  cascade over random chains, CRDT merge laws, determinism) + the 3 attacks +
  partition convergence, for both the plugin and the validators.
- Full suite: **592 passed, 1 skipped**. `make ci-local`: all 5 checks pass
  (`ruff`, `ruff format`, `pyright` strict, `pytest`).

## Files

- `packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable.py` — plugin
- registered in `plugins.py` `_BUILTINS` **and** `nest-plugins-reference/pyproject.toml` entry point
- `packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py` + `scenarios/delegated_auth.yaml`
- 4 validators in `nest_core/validators.py`
- `docs/layers/auth.md`
- tests: `test_delegatable_auth.py`, `test_delegated_auth_scenario.py`

— with Claude Code


dhve:
Thanks @varuni7, this is a genuinely strong submission. The macaroon construction is right: `delegate()` clamps the child with `min(now + ttl, effective_expiry)` and checks the scope subset before minting, `_chain_seals` is a single source of truth for both signature and cascade, and keying revocation on the seal rather than the token string is exactly what makes the subtree severance free. Two details I especially liked: `verify(token, presenter=None)` adds delegation binding as an optional kwarg so it stays drop-in with the base `Auth.verify(token)` contract (the base protocol in `nest_core/layers/auth.py` and the `jwt_auth` revoke-by-exact-string model both confirm you read the surface correctly), and the `test_revoking_any_ancestor_severs_all_at_or_below` hypothesis property is a real invariant test, not a happy-path stub.

This is close. A mechanical blocker plus two things to tighten before it's scoreable:

**1. Rebase onto `main`, then confirm CI goes green.**

The PR is currently `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`, and there are no CI checks reported on the branch yet, so the green run you describe in the body cannot be confirmed from here. The judge panel does not score broken or unmergeable PRs. The files most likely to conflict are the shared registration points every auth/scenario PR touches: the `VALIDATORS` dict in `validators.py`, `_BUILTINS` in `plugins.py`, the `_try_load_builtin` dispatch in `scenarios.py`, and the expected set in `test_all_scenario_types_registered` in `test_validators.py`. Reconcile those additively (keep your `delegated_auth` entries alongside whatever merged since you branched) and push so a fresh CI run attaches.

**2. Make the CRDT merge load-bearing in the scenario, not just the unit tests.**

The partition-tolerant G-Set is your differentiator, and it matters: Problem 04 is the single most crowded problem in this round (roughly twenty open PRs), so the convergence angle is what sets you apart. But in the scenario, `IntermediaryAgent._merge_and_probe` folds the gossiped revocation in with `auth.revocations.revoke(seal)`, a single-element add, not the `RevocationSet.merge` union you documented. So the commutative/idempotent merge laws are proven in `test_delegatable_auth.py` but never actually exercised at the scenario level, and `validate_delegated_auth_convergence` is really reading the three per-intermediary `converge:` probe lines rather than watching replicas reconcile. Have the scenario call `merge()` between replicas (or gossip a `RevocationSet` snapshot and merge it) so the convergence claim in `docs/layers/auth.md` and the scenario docstring is demonstrated end to end, not asserted. That is the cheapest way to make the novelty dimension land.

**3. The cascade is verified entirely on the coordinator's replica.**

Every `verify:` line in the trace comes from `DelegationCoordinator._verify_and_record`, i.e. the same node that called `revoke()`. That correctly proves seal-chain cascade, but it means `validate_delegated_auth_cascade` never observes a verifier that learned the revocation by gossip rather than by revoking it itself. If a leaf (or a second verifier replica) merged the gossiped seal and then rejected a descendant, the cascade and the convergence stories would be the same story, and both would be distributed. Optional, but it would tighten the persona: a swarm verifies at the edge, not at the issuer.

Minor: `set_now` rebuilds a fresh `LogicalClock(tick)` each call, which drops any injected `VirtualClock`. Harmless here since the agents push `ctx.time` in before every op, but worth a comment so the next reader does not expect a live clock to survive.

**Suggested path forward:**

1. Rebase onto `main`, reconciling `validators.py`, `plugins.py`, `scenarios.py`, and `test_validators.py` additively.
2. Swap the scenario's direct `.revoke()` for an actual `RevocationSet` / `merge()` exchange between replicas so convergence is demonstrated, not asserted.
3. Optionally add one edge-verifier that rejects a descendant after merging the gossip, unifying the cascade and convergence proofs.
4. Run `make ci-local` and confirm all 5 stages green before requesting re-review.

Leaving this open. Ping me on the hackathon channel if any of the rebase reconciliation is unclear.

varuni7:
Thanks @dhve!!  really appreciate the detailed review. Rebased onto main  

Since #138 merged the cascading-macaroon version of Problem 4 while this was open, I renamed mine to delegatable_crdt (plugin) / convergent_auth (scenario) so it's purely additive alongside #138 and leaned it into the partition-tolerant angle you mentioned as the differentiator. All three points addressed:

1. The CRDT merge is now load-bearing in the scenario  the gossiped revocation arrives as a RevocationSet and is folded in via RevocationSet.merge() (the commutative/idempotent union), not a single-element add.
2. Added an edge verifier : a non-issuer replica that learned the revocation only by gossip rejects a descendant, so cascade + convergence are proven off the issuer.
3. set_now now documents that it replaces the clock (a live VirtualClock won't survive the call).

varuni7:
Hi @dhve , requesting you to approve the workflow

JamesCarnley:
Thank you, @varuni7. The valuable future requirement is monotonic distributed revocation: after a partition heals, every verifier that observes a valid revocation must converge to rejection and never re-accept, while the pre-merge partition window remains explicitly distinguishable. Current Town has flat grants and no replica, gossip, delegation-chain, or revocation-consistency claim. This frozen scenario can also treat unrelated exceptions as successful attack blocking, which is not acceptable evidence. I am closing the legacy PR as a future profile and retaining the two-replica before/after-union, provenance, and never-reaccept cases for later work. No contributor code was merged or executed during this review.

PR #54, opened 2026-07-01
[Hackathon] action-state-group: capsule-emit trust layer — verifiable, anchored records for NANDA agents

StevenMih:
## What this adds

`capsule-emit-nanda` — a drop-in NANDA trust plugin backed by the [Agent Action Capsule](https://github.com/action-state-group/agent-action-capsule) spec (IETF draft). Every corroborated receipt triggers a `capsule_emit.emit()` call, sealing the reputation event to a tamper-evident ledger that any third party can independently verify — with zero changes to existing NANDA agent code.

A companion demo (`StripeCapsuledPayments`) shows the same capsule pattern applied to payment events.

## How it works

```
NANDA scenario  →  ctx.plugins.get("trust").report(agent, evidence)
                →  CapsuleEmitTrust.report()
                →  capsule_emit.emit(action, operator, agent_input, agent_output, anchor=True)
                →  anchored to https://anchor.agentactioncapsule.org/v1/digest (live public log)
                →  capsule sealed to local JSONL ledger
                →  agent-action-capsule verify --store capsule_ledger.jsonl  ✔
```

**Three gates before a receipt contributes to reputation:**
1. Valid — Ed25519 issuer signature verifies
2. Corroborated — distinct counterparty co-signed the same interaction
3. Anchored — a capsule was emitted and is present in the ledger

Gate 3 is the novel contribution: an agent whose interactions are never anchored gets **zero reputation score** even if their receipts are individually valid and corroborated. The capsule ledger is the authoritative, externally-auditable record.

## Installable package

```
pip install -e examples/capsule-emit
```

```yaml
# scenario.yaml
layers:
  trust: capsule_emit
```

## Smoke tests

```
$ pytest examples/capsule-emit/tests/ -v
tests/test_smoke.py::test_trust_report_and_score  PASSED
tests/test_smoke.py::test_sandbox_pay             PASSED
```

Both plugins instantiate cleanly, exercise a happy path each, and sandbox mode runs without any Stripe credentials.

## Files changed

| Path | Role |
|------|------|
| `examples/capsule-emit/capsule_emit_nanda/trust.py` | `CapsuleEmitTrust` — NANDA Trust protocol implementation |
| `examples/capsule-emit/capsule_emit_nanda/payments.py` | `StripeCapsuledPayments` — standalone payment demo |
| `examples/capsule-emit/pyproject.toml` | Installable package config, entry-points |
| `examples/capsule-emit/README.md` | Plugin docs |
| `examples/capsule-emit/tests/test_smoke.py` | Smoke tests |
| `examples/stripe-capsule-payment/README.md` | Stripe demo docs with protocol-conformance caveats |

## Demo clarifications

- `StripeCapsuledPayments` is a **standalone demo**, not a conforming NANDA Payments protocol implementation. The NANDA Payments protocol requires `pay(to, amount, ref) -> Receipt`; this class uses a different signature. `@runtime_checkable` only checks method names, so an `isinstance` check will falsely pass — documented in the module docstring.
- On the real-Stripe path, `payee` is **not enforced at the Stripe level** (no `destination`/`transfer_data`) — the capsule commits to the payer/payee pair by digest, but the charge does not route to the payee.


StevenMih:
@Skyrider3 
all three review points addressed in 0c81bc7 (relabel + payee caveat, rounding, smoke tests); ready for re-review
Note - i renamed this PR to join the Hackathon. So thsi PR #54superset of PR #32  which is closed.

dhve:
Thanks @StevenMih. The drop-in ergonomics here are genuinely nice: `CapsuleEmitTrust` delegating severance to `AgentReceiptsTrust` while keeping the plain-string fallback means stock scenarios run unchanged, and the caveat docs in `StripeCapsuledPayments` are more honest than most submissions manage.

**1. The mandatory adversarial validator is missing, and Gate 3 cannot actually fire in a run.**

The charter requires an adversarial validator catching an attack the reference plugin misses, plus a scenario passing under it. `scenarios/receipt_reputation_capsule.yaml` is the stock `receipt_reputation` scenario with the trust layer swapped, which demonstrates parity with `agent_receipts`, not an attack it fails. And as wired, the anchor gate is inert: `report()` calls `_emit_capsule()` for every receipt that passes `_verify_receipt`, so the only way a valid, corroborated receipt ends up unanchored is if `capsule_emit.emit` itself raises. No adversary inside the sim can produce the "valid but unanchored" state the PR body says gets zero score. The tax-audit demo that shows the deterrent lives in your external capsule-emit repo; judges score what is in this PR. Build the attack in-repo: a scenario plus validator where `agent_receipts` accepts a mutated record and `capsule_emit` flags the digest mismatch.

**2. The plugin ships twice, with different scoring semantics, under the same entry-point name.**

`examples/capsule-emit/capsule_emit_nanda/trust.py` scores directly from anchored, severance-filtered receipts (importing six underscore-private helpers like `_effective_receipts` from `agent_receipts`), while `packages/nest-plugins-reference/.../trust/capsule_emit_trust.py` delegates to `AgentReceiptsTrust` and multiplies `score` and `confidence` by `anchor_ratio`. Those return different numbers for the same ledger. Both pyprojects register `nest.plugins.trust` -> `capsule_emit` (and `stripe_capsule` likewise), so following your README's `pip install -e examples/capsule-emit` creates duplicate entry points and which class loads is undefined. Pick one home and delete the other copy.

**3. New hard dependencies and a demo plugin inside the shared reference package.**

The diff adds `capsule-emit` and `agent-action-capsule` to `nest-plugins-reference` hard dependencies, so every participant's `uv sync` now installs your external packages, and it registers `stripe_capsule` there too, a plugin you document as not conforming to the Payments Protocol (`pay(payer, payee, amount, currency) -> dict` vs the protocol's `pay(to, amount, ref) -> Receipt`). The charter says existing extras only, and no touching code outside your layer. Keep everything under `examples/capsule-emit/` and revert the reference-package files and pyproject changes.

**4. Determinism red flags.**

`_sandbox_pay` seeds the intent id with `time.monotonic_ns()`, so the same seed gives a different ledger every run. Separately, `_USE_REAL_STRIPE` is computed at import time, so the `no_real_stripe` fixture's `monkeypatch.delenv` cannot force sandbox mode: on any machine with `STRIPE_SECRET_KEY` exported, `test_sandbox_pay` calls `stripe.PaymentIntent.create(confirm=True)` for real. Read the env inside `pay()`. Also: does `capsule_emit.emit` stamp wall-clock time into the capsule? If yes, the ledger is not byte-reproducible even with `anchor=False`.

**5. Process items.**

The branch conflicts with `main` and no CI checks are reported on it; the judge panel does not score broken PRs, so rebase first. The branch name `hackathon/capsule-trust` is missing your handle (pattern is `hackathon/<handle>-<short-theme>`, non-matching branches are subject to auto-close). And the charter is one problem per participant, single PR: you also have #55 open targeting Problem 09, while this PR names no problem at all. Say which one is your submission and how the other relates.

**Suggested path forward:**

1. Decide whether #55 or this PR is your submission; if this one, name the problem it targets in the body.
2. Collapse to the `examples/capsule-emit/` package only and revert all `packages/nest-plugins-reference/` changes.
3. Ship the adversarial validator plus a scenario where `agent_receipts` misses a record-mutation attack and `capsule_emit` catches it, so Gate 3 is reachable and load-bearing in tests.
4. Fix the determinism items and add a test using `sign_receipt`/`cosign_receipt` so at least one test exercises a real anchored receipt instead of only the plain-string fallback.
5. Rebase on `main`, fix the branch name, and get `uv sync`, `ruff check`, `ruff format --check`, `pyright`, and `pytest -v` green.

Leaving this open. Ping me on the hackathon channel if the validator shape is unclear.

StevenMih:
Addressed issues #2, #3, and #4 from @dhve review (commit 53fa950):

**#2 — Duplicate plugin removed**
Deleted `packages/nest-plugins-reference/trust/capsule_emit_trust.py` and `payments/capsule_emit_payments.py`. Single implementation lives under `examples/capsule-emit/` only.

**#3 — Reference-package changes reverted**
Removed `capsule-emit` and `agent-action-capsule` from `nest-plugins-reference` hard deps, removed `stripe_capsule` and `capsule_emit` entry points from the reference package. `pyproject.toml` restored to pre-PR state. All new code stays in `examples/capsule-emit/`.

**#4 — Determinism fixed**
- `_USE_REAL_STRIPE` removed from module level; `os.environ.get("STRIPE_SECRET_KEY")` now called inside `pay()` and `verify_payment()` so `monkeypatch.delenv` in tests can force sandbox mode.
- `_sandbox_pay` now seeds from `hash(payer:payee:amount:currency)` — deterministic, no `time.monotonic_ns()`.

**#5 — Branch rename**
Pushed to `hackathon/stevenmih-capsule-trust` (correct handle format). This PR still tracks `hackathon/capsule-trust` — can close+reopen from the new branch or retarget manually.

**#1 — Adversarial validator (in progress)**
We have `sign_receipt` / `cosign_receipt` helpers available in `agent_receipts.py` and tamper-detection patterns from the `capsule-emit` and `agent-action-capsule` test suites. Building a test where `agent_receipts` accepts a mutated receipt but `CapsuleEmitTrust` drops it to zero (via ledger re-verification in `score()`). Will follow up in the next commit.

StevenMih:
**#1 adversarial validator complete** (commit 117d395):

**What was missing and why:** `score()` only checked `self._anchored` (capsule was *emitted*) but never re-verified the sealed digest against the current receipt content. An adversary mutating a receipt in-memory after sealing still got full reputation — identical to `agent_receipts`, so Gate 3 was a no-op.

**Fix — ledger re-verification in `score()`:**
`score()` now calls `capsule_emit.read_ledger()` + `capsule_emit.verify_input_digest(capsule, receipt)` for each anchored receipt. A receipt whose in-memory content no longer matches its `agent_input_digest` is excluded from scoring. `agent_receipts` has no ledger reference and cannot detect this.

**New test `test_gate3_tampered_receipt_excluded_from_score`:**
```
1. sign_receipt() + cosign_receipt() → valid, corroborated receipt
2. CapsuleEmitTrust.report() → capsule sealed with agent_input_digest = hash(receipt)
3. score() before tamper → score > 0, confidence > 0  ✓
4. adversary mutates receipt["action"]["category"] in-memory
5. score() re-runs → verify_input_digest detects mismatch → score = 0.0  ✓
```
`agent_receipts` would return score > 0 at step 5 (no ledger check).

**Also added:**
- `test_sandbox_pay_is_deterministic`: confirms same payer/payee/amount → same `pi_sandbox_*` (covers the `monotonic_ns` fix)
- Type annotations on `monkeypatch` and `tmp_path` pytest fixtures

Run locally: `pip install -e examples/capsule-emit && pytest examples/capsule-emit/tests/ -v`

StevenMih:
Closing to reopen from correctly-named branch `hackathon/stevenmih-capsule-trust` (review item #5). All commits and history carry over — this PR's thread is the reference. New PR incoming.

PR #55, opened 2026-07-01
[Hackathon] Problem-09 privacy: SD-JWT capsule selective-disclosure plugin

StevenMih:
## Summary

Implements the SD-JWT salted-hash commitment model from
`draft-mih-scitt-agent-action-capsule-sel-disc` as a NANDA Town privacy plugin.
Addresses Problem-09 (hybrid encryption + selective disclosure + 4-attack adversarial validator).

**Distinct from `hybrid_x25519` (PR #28):** that plugin uses Merkle authentication
paths.  This plugin uses per-field salted-hash commitments — no tree traversal,
decoy digests, algorithm-agile `_sd_alg` header, wire-compatible with
capsule-emit SD-Capsule output.

### Deliverables

- **`nest_plugins_reference/privacy/capsule_selective_disclosure.py`** — `CapsuleSelDiscPrivacy` plugin
  - `commit_credential` / `build_witness` / `verify_disclosure` standalone API (capsule-emit compatible)
  - X25519 ephemeral-static ECDH + HKDF-SHA256 + ChaCha20-Poly1305 AEAD, per-recipient key-wrap
  - Broadcast revocation via epoch counter; anti-replay via `(sender, msg_id)` seen-set
  - Tier-1 deterministic mode

- **`tests/test_capsule_selective_disclosure.py`** — 41 tests
  - `TestAdversarialValidator::test_capsule_plugin_passes_all_attacks` — all 4 attacks defeated
  - `test_noop_fails_*` — confirms noop has no protections (required by Problem-09)
  - ruff-clean, pyright-clean

- **`scenarios/sealed_bid_capsule_privacy.yaml`** — 1 auctioneer + 7 bidders, 3 rounds, Tier 1

### Commitment scheme

```
digest = BASE64URL(SHA-256(UTF8(JCS([salt_b64url, name, value]))))
```

- `_sd` array sorted lexicographically; `_sd_alg: "sha-256"` in every SD-object
- Decoy digests hide the count of concealed fields

### 4 attacks defeated

| Attack | Mechanism |
|--------|-----------|
| Eavesdropper | No wrap entry → `NotInAudienceError` before any AEAD |
| Replay | `msg_id` bound in AAD + seen-set → `ReplayError` |
| Field injection | SHA-256 of tampered triple ∉ `_sd` → `verify_proof` returns `False` |
| Stale revocation | Revoked agent excluded from wrap set → `NotInAudienceError` |

### Test run

```
41 passed in 0.24s
```

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -40,6 +40,9 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("memory", "lww_register"): f"{_REF}.memory.lww_register:LwwRegisterMemory",
     ("privacy", "noop"): f"{_REF}.privacy.noop:NoopPrivacy",
     ("privacy", "hybrid_x25519"): f"{_REF}.privacy.hybrid_x25519:HybridX25519Privacy",
+    ("privacy", "capsule_selective_disclosure"): (
+        f"{_REF}.privacy.capsule_selective_disclosure:CapsuleSelDiscPrivacy"
+    ),
     ("datafacts", "datafacts_v1"): f"{_REF}.datafacts.datafacts_v1:DataFactsV1",
     ("datafacts", "cid_facts"): f"{_REF}.datafacts.cid_facts:CidFacts",
 }
--- a/packages/nest-plugins-reference/nest_plugins_reference/privacy/capsule_selective_disclosure.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/privacy/capsule_selective_disclosure.py
@@ -1,0 +1,863 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Capsule selective-disclosure privacy plugin.
+
+Unlike :mod:`nest_plugins_reference.privacy.hybrid_x25519` — which uses a Merkle
+authentication-path construction for selective disclosure — this plugin implements
+the **SD-JWT salted-hash commitment model** from the Agent Action Capsule
+selective-disclosure profile (``draft-mih-scitt-agent-action-capsule-sel-disc``):
+
+* **Hybrid encryption (HPKE-shaped).** X25519 ephemeral-static ECDH + HKDF-SHA256
+  key-agreement, ChaCha20-Poly1305 AEAD per message, one symmetric content-key
+  wrap per recipient.  Identical broadcast cost model to the Merkle-path plugin.
+* **Selective disclosure — SD-JWT salted-hash model.**  :meth:`commit_credential`
+  commits each field as ``digest = BASE64URL(SHA-256(UTF8(JCS([salt, name, value]))))``,
+  inserts the digest into a per-object ``_sd`` array, and optionally adds **decoy
+  digests** to hide the count of concealed fields.  A holder reveals a subset by
+  sharing the ``[salt, name, value]`` triple; a verifier checks
+  ``BASE64URL(SHA-256(UTF8(JCS(triple))))`` is present in the ``_sd`` commitment
+  set.  No authentication path is required — the commitment is per-field.
+* **Broadcast revocation.** :meth:`revoke` advances an epoch and excludes the
+  revoked agent from future key-wrap sets without touching any live member's key.
+
+Differences from the Merkle-path approach (``hybrid_x25519``)
+--------------------------------------------------------------
+
+``hybrid_x25519`` commits all fields to a single Merkle root and proves individual
+fields via their authentication paths (sibling hashes up the tree).  **This plugin
+uses per-field salted-hash commitments** — the same construction the AAC sel-disc
+I-D (``draft-mih-scitt-agent-action-capsule-sel-disc``) applies to Capsule payloads:
+
+* **No tree traversal.** Verifying a disclosure is one SHA-256 comparison; there
+  is no ``_root_from_path`` computation.
+* **Decoy digests.** Extra commitment digests with no associated disclosure hide
+  the count of concealed fields, providing ``k``-anonymity over field count.
+* **Algorithm-agile.**  The ``_sd_alg`` member (``"sha-256"``) permits future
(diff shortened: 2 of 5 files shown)


dhve:
Thanks @StevenMih. Implementing your own I-D as a Nanda Town plugin is a great fit for this problem, and the core code is well built: `CapsuleSelDiscPrivacy`'s constructor mirrors `HybridX25519Privacy` exactly (`agent_id`, `seed`, `deterministic`), the AAD in `_Envelope.aad` binds scheme, msg_id, sender, epoch, and the sorted recipient key ids, and the standalone `commit_credential` / `build_witness` / `verify_disclosure` trio is a clean capsule-emit-compatible surface. The `_tamper_disclosure` test helper doing a real decode-modify-reencode field injection (instead of byte mangling) is exactly right. Docstrings, SPDX headers, and entry-point wiring are all in order.

**1. The adversarial validator cannot demonstrate the charter's bar: same validator FAILS on `noop`, PASSES on yours.**

`_adversarial_validate(privacy: CapsuleSelDiscPrivacy)` in `test_capsule_selective_disclosure.py` never uses its `privacy` argument. Every attack constructs fresh `CapsuleSelDiscPrivacy` instances internally, so the validator literally cannot be pointed at `NoopPrivacy`. The `test_noop_fails_*` tests are separate hand-written assertions of noop behavior (and `test_noop_fails_eavesdropper` does not even model an outsider, it round-trips through the same instance), not the validator failing on the baseline. The `except Exception: results[...] = True` catch-alls compound this: a plugin that raises on every decrypt would "defeat" attacks 1 and 4. Note the repo already ships plugin-agnostic checks in `nest_plugins_reference/validators/privacy_validators.py` (from #28) covering these exact four attacks against the `Privacy` protocol surface. Running those against your plugin would prove protocol-surface compatibility and satisfy the fails-on-noop requirement with minimal new code. This is the biggest correctness and test-rigor gap.

**2. Rebase needed, and CI has not run.**

`mergeable` is CONFLICTING: your branch predates #115, which added `trust_gated` to both `nest_core/plugins.py` and the `[project.entry-points."nest.plugins.privacy"]` table in `packages/nest-plugins-reference/pyproject.toml`, right where your new entries land. Both resolutions are trivial (keep both entries). No CI checks have reported on this branch yet; the judge panel does not score broken PRs, so after rebasing please confirm the full local pipeline (`uv sync`, `ruff check`, `ruff format --check`, `pyright`, `pytest -v`) is green.

**3. Nothing exercises `scenarios/sealed_bid_capsule_privacy.yaml`, and its header comment overstates the run.**

The YAML is a near-copy of `sealed_bid_with_privacy.yaml` with the layer swapped, which is fine, but #28 paired its scenario with `test_hybrid_x25519_scenario.py`: resolve the layer through `PluginRegistry`, boot the scenario at two seeds via `ScenarioRunner`, and assert two seed-42 runs produce byte-identical traces. Your PR has no equivalent, so the Tier-1 claim for this scenario is unverified. Also, unless I am misreading `scenarios_builtin/auction.py`, the state-machine auction factory never calls the privacy layer, so the comment's claim that "each bid is encrypted with X25519+ChaCha20-Poly1305 so only the auctioneer can see the amount" does not describe what this run does. #28's scenario header was explicit that sealing mechanics live in the plugin tests and the scenario proves discoverability plus determinism; I would mirror that framing.

**4. Novelty: make the SD-JWT delta load-bearing, since Problem-09 is already merged as #28.**

Your differentiation pitch (per-field salted-hash commitments vs Merkle paths, decoys, `_sd_alg` agility, capsule-emit wire compat) is credible, but the four attacks you validate are the same four #28 validates, so the tested surface shows no delta. Two cheap wins: (a) the duplicate-disclosure guard in `verify_proof` (the `consumed` set, your SD-4 step 4) is implemented but has zero tests, and it is an attack class the Merkle plugin does not even discuss; (b) decoy coverage is thin (`test_decoy_count_padding` only checks array length, and `test_decoy_has_no_disclosure` uses a hand-built `"A" * 43` string rather than a real `_decoy_digest`). A test that a decoy digest admits no valid disclosure, plus a hypothesis property or two over arbitrary field maps (as #28 did in `test_hybrid_x25519_properties.py`), would move both novelty and test rigor.

Two housekeeping notes: the charter's branch format is `hackathon/<handle>-<short-theme>` and yours omits the handle (mismatched names are subject to auto-close, so worth noting in the PR body). And the charter asks for one problem per participant with a single PR; you also have #54 open on the trust layer. Please state which is your primary submission or how the two relate.

**Suggested path forward:**

1. Rebase onto main and resolve the two small conflicts in `nest_core/plugins.py` and `packages/nest-plugins-reference/pyproject.toml`.
2. Rework the validator to accept any `Privacy` instance (or reuse the `check_*` functions from `validators/privacy_validators.py`), then assert it fails on `NoopPrivacy` and passes on `CapsuleSelDiscPrivacy` through the same code path.
3. Add duplicate-disclosure and decoy-indisclosability tests, plus a hypothesis property over field maps.
4. Add a scenario boot + replay-determinism test mirroring `test_hybrid_x25519_scenario.py`, and soften the scenario header comment.
5. Clarify the #54 / #55 relationship in both PR bodies.
6. Run the full local CI pipeline and confirm all stages green before requesting re-review.

Leaving this open. Ping me on the hackathon channel if the validator reshaping is unclear.

StevenMih:
**PR #54 is the primary hackathon submission** (capsule-emit trust layer, Problem addressed: receipt tamper-detection via Gate 3 ledger re-verification).

This PR (#55, Problem-09 SD-JWT selective disclosure) is a secondary submission. Per charter one-per-participant rule, please score #54 as the primary entry.

Cross-reference: primary PR reopened as #177 (`hackathon/stevenmih-capsule-trust`) — same commits as #54, branch renamed to match required naming convention. Review history is in #54.

StevenMih:
closing this pr as we are rebuilding in pr 197

PR #56, opened 2026-07-02
[Hackathon] theritwik: delegatable capability tokens with cascading revocation

theritwik:
## Problem

Implements hackathon problem #04 (docs/hackathon/problems/04-auth-capability-delegation.md):
delegatable capability tokens with cascading revocation for the auth layer.

## What this adds

- **Plugin** `auth/delegatable.py` — macaroon-style HMAC chaining (Birgisson et al., 2014).
  `delegate(parent_token, audience, scopes_subset, ttl)` mints a narrower, time-bounded
  sub-token **without the issuer's secret**; child scopes must be a subset of the parent's
  (else `ScopeEscalationError`), child expiry is clamped to the parent's.
  `verify_presented(token, presenter)` adds audience binding. Revoking any ancestor fails
  every descendant at the next verify (`RevokedAncestorError`) — cascading by construction.
- **Adversarial validator** `validators/auth_validators.py` — three pure checks: scope
  escalation, stale/revoked ancestor, audience confusion. **FAILS against the reference
  `jwt` plugin, PASSES against `delegatable`** (proven in tests).
- **Scenario** `scenarios/delegated_auth.yaml` — 16-agent delegation tree (1 coordinator,
  3 intermediaries, 12 leaves). Mid-run one intermediary is revoked; exactly its 4 leaves
  fail with `RevokedAncestorError`, the other 8 keep verifying. Deterministic under seeds
  42, 7, 1337.
- **Tests** — 30 new: unit, Hypothesis property tests, forged-token defenses (validly-signed
  scope-widening and ttl-extension rejected at verify), validator discrimination, and
  full-simulator integration.
- Registered as `("auth", "delegatable")` builtin + `nest.plugins.auth` entry point;
  docs in `docs/layers/auth.md`.

## Determinism

Token ids are content hashes, signatures are HMAC-SHA256 over canonical JSON, expiry uses
a logical tick clock — no wall time, no unseeded RNG.

## Verify

    make ci-local
    # or just the new tests:
    uv run pytest packages/nest-plugins-reference/tests/test_delegatable_auth.py \
                  packages/nest-plugins-reference/tests/test_delegated_auth_scenario.py -v

All five CI commands exit 0 locally (ruff check, ruff format --check, pyright strict,
pytest: 766 passed).

### DIFF
--- a/docs/layers/auth.md
+++ b/docs/layers/auth.md
@@ -21,6 +21,56 @@ shape (header.payload.sig), no claim validation beyond the signature.
 
 Source: [`nest_plugins_reference/auth/jwt_auth.py`](../../packages/nest-plugins-reference/nest_plugins_reference/auth/jwt_auth.py).
 
+## `delegatable` — macaroon-style delegation with cascading revocation
+
+`jwt` issues flat tokens and revokes by exact token string, so it cannot
+model *delegation* (an agent minting a narrower sub-capability for another
+agent without the issuer) or *cascading revocation* (revoke a parent →
+every descendant fails automatically). `delegatable` adds both, following
+the macaroon construction (Birgisson et al., 2014).
+
+A token carries its whole chain of links; each link's HMAC is keyed on the
+previous link's signature, so **delegation needs only the parent token, not
+the root secret**:
+
+```python
+auth = DelegatableAuth(secret=b"authority-secret")
+root  = await auth.issue(AgentId("coordinator-0"), ["read", "write", "admin"])
+child = await auth.delegate(root, AgentId("worker-3"), ["read"], ttl=100)
+ctx   = await auth.verify_presented(child, AgentId("worker-3"))   # audience-bound
+await auth.revoke(root)          # cascades: child now fails to verify
+```
+
+Extra surface on top of the base `Auth` protocol:
+
+- `delegate(parent_token, audience, scopes_subset, ttl) -> Token` — child
+  scopes must be a subset of the parent's (a superset raises
+  `ScopeEscalationError`); child expiry is clamped to never outlive the
+  parent.
+- `verify_presented(token, presenter) -> AuthContext` — everything `verify`
+  checks, plus that the presenter is the tip's declared audience.
+
+Three attacks it stops at verify time — an attacker holding a token can
+always *append* a validly-signed link, so every restriction is re-checked on
+the way in — that the `jwt` plugin silently allows:
+
+| Attack | `jwt` | `delegatable` |
+|---|---|---|
+| **Scope escalation** (child grants a scope the parent lacked) | accepted | `ScopeEscalationError` |
+| **Stale ancestor** (child verifies after parent revoked/expired) | accepted | `RevokedAncestorError` / `ExpiredTokenError` |
+| **Audience confusion** (token for B presented by C) | accepted | `AudienceConfusionError` |
+
+The shipped adversarial validator
+([`validators/auth_validators.py`](../../packages/nest-plugins-reference/nest_plugins_reference/validators/auth_validators.py))
+encodes these as pure checks over observed grants: it **fails** against `jwt`
+and **passes** against `delegatable`. Determinism is preserved — token ids are
+content hashes, signatures are HMAC-SHA256, and expiry is measured against a
+logical tick clock, never wall time.
+
+Sources:
+[`auth/delegatable.py`](../../packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable.py),
+scenario [`scenarios/delegated_auth.yaml`](../../scenarios/delegated_auth.yaml).
+
 ## Writing your own
 
 See [`writing-a-plugin.md`](../writing-a-plugin.md). Register under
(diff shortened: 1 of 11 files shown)


theritwik:
CI appears to be awaiting first-time-contributor approval — could a maintainer approve the workflow run? All five checks (ruff, format, pyright, pytest) pass locally via make ci-local.

dhve:
Thanks @theritwik. This is one of the stronger Problem 04 submissions in a very crowded field. The macaroon construction is done properly: `delegate` extends the parent's tip signature and never touches the root secret, and `test_delegate_needs_parent_token_not_root_secret` proves it by delegating through an instance holding the wrong secret. The `_forge_child` tests show exactly the right adversarial instinct (validly signed scope widening and ttl extension, caught by verify-time monotonicity checks rather than by the signature), and extending the surface with `verify_presented` instead of changing `Auth.verify`'s signature keeps the base contract intact, which is precisely where an earlier attempt at this problem went wrong.

**1. The branch conflicts with main, and the branch name is off-charter.**

GitHub reports this PR as CONFLICTING. Main has grown since you branched: all four shared registration files you touch (`nest_core/scenarios.py`, `nest_core/plugins.py`, `nest-plugins-reference/pyproject.toml`, `validators/__init__.py`) gained entries from other merged hackathon PRs. Rebase onto current main. Separately, the branch is `hackathon/theritwik-reputation` but this is an auth submission; the charter's pattern is `hackathon/<handle>-<short-theme>` and mismatched names are subject to auto-close. Renaming the branch in your fork through the GitHub UI carries the open PR along; something like `hackathon/theritwik-auth-delegation` fits. No CI checks have run on this branch yet, and the judge panel does not score broken PRs, so confirm all five stages green after the rebase.

**2. `revoke()` is unauthenticated, so any descendant can revoke the whole tree.**

In `auth/delegatable.py`, `revoke` decodes the token and adds `chain[-1]["jti"]` to the revoked set without ever checking the signature. Because `_mint_jti` derives ids purely from link content (no secret involved) and every child token carries its full ancestor chain, any leaf holder can lift the root link out of its own token, wrap it as `{"chain": [root_link], "sig": "junk"}`, and call `revoke`. That cascades a denial of service to every agent in the system, and in your scenario all 16 agents share the one authority instance, so a single malicious leaf could do it. Your own docstring says every restriction is re-checked on the way in; revocation should meet the same bar. Fix: `await self.verify(token)` (or at minimum the HMAC chain check) at the top of `revoke`, which limits callers to revoking tokens they hold a valid signature for. A descendant cannot reconstruct the root token's signature without the secret, so the attack dies there. Add an adversarial test for the crafted root-link attempt; it would make a strong addition to your forged-token section and matters for both correctness and persona fidelity.

**3. The adversarial validator never runs against the scenario output.**

`check_delegation_safety` is exercised in `test_delegatable_auth.py` over hand-built `GrantObservation` lists, and the jwt-fails/delegatable-passes discrimination there is convincing. But the scenario ledger only records `(agent_id, ok, error_name)` triples, so nothing in `test_delegated_auth_scenario.py` can feed the validator. The charter asks for a scenario that demonstrates the plugin passing under the adversarial validator, and right now that link exists in prose, not code. Have `LeafAgent._verify_into` log jti, parent jti, audience, and presenter (via `auth.describe`), have the coordinator log the jti it revoked, then run `check_delegation_safety` over the ledger at the end of the scenario test. That makes `DelegationAudit` load-bearing end-to-end instead of test-only.

**4. The authority's logical clock is frozen at 0 in the scenario.**

`delegated_auth_factory` instantiates `DelegatableAuth(root_ttl=...)` and nothing in the simulation path ever calls `advance()`, so `ExpiredTokenError` can never fire in the integrated run; expiry and ttl clamping are only exercised in unit tests. Either sync the authority clock to the simulator clock (`ctx.time`) before each verify, or say explicitly in the scenario docstring that expiry is out of scope for this run. Related nit: the factory imports `DelegatableAuth` directly instead of instantiating the resolved class from `plugins["auth"]`, which makes the YAML's `auth: delegatable` line decorative.

**5. Say what differentiates this from the other Problem 04 entries.**

Roughly twenty open PRs implement this same problem, and novelty is scored against the field. Your strongest cards are real: delegation that needs only the parent token (the problem statement explicitly warns against "delegation" that is really re-issuance by the central authority, and `test_delegate_needs_parent_token_not_root_secret` pins the difference), verify-time re-checking against validly signed forged links, and a non-breaking protocol extension. Add a short differentiators section to the PR body and keep each claim pinned by a named test.

**Suggested path forward:**

1. Rebase onto main and resolve the four registration-file conflicts; rename the branch to match the auth theme.
2. Verify before revoking, plus an adversarial test for the crafted root-link revocation.
3. Extend the ledger and run `check_delegation_safety` inside the scenario test.
4. Decide the clock story: sync to `ctx.time` or document the choice.
5. Add the differentiators section, run `make ci-local`, and confirm all five stages green.

Leaving this open. Ping me on the hackathon channel if the revoke hardening is unclear.

JamesCarnley:
Thank you, @theritwik. A future delegation profile must prove holder-only attenuation, verifier-side scope and expiry monotonicity, presenter binding, authenticated revocation of exactly the intended subtree, and attack results derived from actual emitted evidence rather than scenario labels. Current Town uses controller-signed, nondelegatable Run Grants and has no revocation tree. The frozen revoke path is unauthenticated and its adversarial validator is disconnected from real scenario results, so it is not a safe basis for extending current grants. I am closing this as a future-profile legacy PR while retaining forged-revocation, widened/longer child, presenter, cascade, and healthy-sibling cases. No contributor code was merged or executed during this review.

PR #57, opened 2026-07-02
[Hackathon] distributed-systems-architect: private_commerce — four-layer cross-plugin composition with joint validators

Abi5678:
# [Hackathon] distributed-systems-architect: private_commerce — four-layer cross-plugin composition with joint validators

## Problem picked

None of the ten listed problems — this PR targets the gap they leave behind. Every merged hackathon plugin (gossip registry #24, streaming payments #21, hybrid_x25519 privacy #28, agent_receipts trust #26) was designed, tested, and validated **in isolation**. Nobody has asked what happens when they run **together**, and no validator in the repo can even express a cross-layer invariant. This PR is a **scenario + validator** submission per the charter ("a layer plugin *or* a scenario *or* a validator"): a `private_commerce` scenario that composes four merged plugins into one workflow, plus four joint validators that test emergent properties spanning layer boundaries.

## What it does

**Scenario (12 agents):** 5 buyers discover sellers through per-agent gossip registries (partition-honest — buyers 0-1 are partitioned away for the first 3000 ticks and must wait for the heal), open per-tick streaming payments, and send bids wrapped in hybrid X25519 + ChaCha20-Poly1305 envelopes. 4 honest sellers decrypt, deliver, and issue Ed25519 cross-signed purchase receipts; they also settle among themselves in a directed receipt cycle, forming the strongly-connected honest anchor that collusion severance needs. The adversary — `shill_seller-0` — looks legitimate on the registry, decrypts the bid, **drains the stream, never delivers**, and covers itself by wash-trading mutually co-signed fake receipts with `shill-0`. An auditor ingests receipts and negatives into the configured trust plugin and emits final scores.

**Joint validators** (registered as `VALIDATORS["private_commerce"]`):

| Validator | Layers spanned | Invariant |
|---|---|---|
| `commerce_discovery_precedes_bid` | registry × privacy | every encrypted bid is preceded by an honest gossip discovery of that seller |
| `commerce_bid_opacity` | privacy × transport | the known bid plaintext never appears in any `bid:` wire message (send / receive / **dropped**) |
| `commerce_undelivered_penalized` | payments × trust | a seller that drained a stream without delivering ends below the trust threshold — *despite* wash-traded receipts |
| `commerce_delivery_rewarded` | payments × trust | every fulfilment is backed by a payment stream, and fulfilling sellers score above threshold |

## Adversarial discrimination (one YAML line each)

The charter asks for validators the reference plugin cannot satisfy. Here each cross-layer check is pinned to the specific plugin that makes it pass:

- `privacy: hybrid_x25519 → noop`: bid envelopes become plaintext; `commerce_bid_opacity` **FAILS** (the other three still pass — the failure is surgically attributable).
- `trust: agent_receipts → score_average`: the shill's wash-traded receipts *raise* its running average to 0.75; `commerce_undelivered_penalized` **FAILS**. Under `agent_receipts` the isolated mutual co-signing pair is severed by the SCC analysis and the shill collapses to 0.0 — the composition passes.

Both discriminations are integration tests (`test_noop_privacy_fails_exactly_the_opacity_validator`, `test_score_average_trust_fails_exactly_the_penalty_validator`) that assert the *exact* verdict vector, not just "something failed."

## Design decisions & tradeoffs

- **Ground-truth sidecar (`bidmeta:`)**: opacity is unfalsifiable unless the validator knows the plaintext. Buyers broadcast a `bidmeta:` marker declaring what they encrypted; the validator then proves that string never rides the wire in a `bid:` message. In production this sidecar would not exist; in a test rig, making the invariant checkable is the point.
- **Marker broadcasts vs. functional sends**: markers (`discovered:`, `stream:*`, `fulfilled:`, `score:`) are broadcasts — recorded in the trace at *send* time, so a dropped delivery can never make the trace lie about what an agent did. Functional messages (bid, ack, receipt) get 3× send redundancy with receiver-side dedup by ref/receipt-id; self-scheduled ticks get 5× redundancy (drop-chain death probability ≈ 0.05⁵). This is the same pattern the gossip scenario (#24) uses, extended to point-to-point traffic.
- **Deterministic pairing**: buyer *i* targets the *i*-th sorted discovered seller, so with 5 buyers and 5 sell-capable cards, the shill gets exactly one victim under every seed — the adversary is guaranteed to fire, and the penalty validator fails loudly if it ever doesn't (`"the adversary never fired"`), rather than passing vacuously.
- **Trust threshold 0.3**: an honest seller with one corroborated purchase receipt scores `1 − exp(−5/10) ≈ 0.39` under `agent_receipts`; a severed shill scores exactly 0.0; the reference neutral prior is 0.5. 0.3 separates every honest outcome from the severed adversary with margin on both sides, and sits below the 0.5 prior so `score_average`'s wash-trade inflation (0.75) is caught.
- **Per-agent plugin wiring** uses the runner's existing `_agent_plugins` override channel (same as #24): per-agent gossip views, per-buyer streaming ledgers, per-agent X25519 keypairs cross-registered at build time. No runner changes needed.

## Test rigor

25 tests in `packages/nest-plugins-reference/tests/test_private_commerce.py`:

- **17 unit tests** covering pass / fail / absent-marker / vacuous-pass paths for all four validators (e.g. zero-total stream is not a drain; a delivered stream is not an offense; unpaid fulfilment is caught).
- **2 Hypothesis property tests**: arbitrary junk markers interleaved with valid ones never crash any validator; N-fold marker duplication (the drop-redundancy scenario) never changes any verdict.
- **6 integration tests** booting the full simulator: all validators green under seeds 42 and 7; byte-identical traces under a repeated seed; different traces under different seeds; and the two exact-verdict adversarial discriminations above.

## How to verify

```bash
uv run nest run scenarios/private_commerce.yaml
uv run python -c "
from pathlib import Path
for r in validate_trace(Path('traces/private_commerce.jsonl'), 'private_commerce'):
    print('PASS' if r.passed else 'FAIL', r.name, '-', r.detail)
"
# Adversarial: edit scenarios/private_commerce.yaml, set privacy: noop → opacity FAILS
# Adversarial: set trust: score_average → undelivered-penalty FAILS
uv run pytest packages/nest-plugins-reference/tests/test_private_commerce.py -v
```

Full CI (`uv sync && uv run ruff check . && uv run ruff format --check . && uv run pyright && uv run pytest -v`) is green locally: 761 passed, 0 ruff, 0 pyright.

## Persona

`distributed-systems-architect`: the submission's whole thesis is that correctness of parts does not compose for free. The code reflects it — redundancy budgets computed from the drop rate, partition-aware pairing deadlines, an SCC-severance-aware threshold derivation, and validators that treat the trace as the only source of truth (markers recorded at send time so drops can't hide behavior). The novelty is not any one algorithm; it is the instrumented composition and the two one-line adversarial swaps that attribute each cross-layer failure to the exact layer that caused it.

### DIFF
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -175,3 +175,9 @@ def _try_load_builtin(name: str) -> None:
         from nest_core.scenarios_builtin.sybil_bond import sybil_bond_factory
 
         register_scenario("sybil_bond", sybil_bond_factory)
+    elif name == "private_commerce":
+        from nest_core.scenarios_builtin.private_commerce import (
+            private_commerce_factory,
+        )
+
+        register_scenario("private_commerce", private_commerce_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/private_commerce.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/private_commerce.py
@@ -1,0 +1,778 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Private commerce scenario — a four-layer cross-plugin composition.
+
+Every hackathon plugin so far was tested in isolation: gossip registry
+against a partition, streaming payments against a drain attack, hybrid
+encryption against an eavesdropper, receipt reputation against a collusion
+ring. This scenario is the first to run four of them **together** and
+instrument the emergent, cross-layer invariants none of the single-layer
+validators can see:
+
+1. **Discovery honesty × privacy.** Buyers may only learn sellers through
+   gossip that respects the partition; every encrypted bid must be preceded
+   by an honest discovery event (``registry`` × ``privacy``).
+2. **Bid opacity.** The bid amount travels the wire only inside a hybrid
+   X25519 envelope. A ground-truth ``bidmeta:`` sidecar marker lets the
+   validator assert the plaintext never leaks (``privacy`` × transport).
+3. **Undelivered streams are penalized.** A seller that drains a payment
+   stream and never delivers must end with a low receipt-based trust score
+   — even when it wash-trades fake receipts with a shill partner
+   (``payments`` × ``trust``).
+4. **Delivery is rewarded.** Sellers that fulfil their bid end with an
+   adequate trust score, and every fulfilment is backed by an earlier
+   payment stream (``payments`` × ``trust``).
+
+Cast (12 agents):
+
+* ``buyer-0..4`` — discover sellers via per-agent gossip registries, open a
+  payment stream, send a hybrid-encrypted bid, then verify delivery. An
+  unfulfilled bid triggers a ``negative:`` report to the auditor.
+* ``seller-0..3`` — honest sellers. Register a ``sell`` card, decrypt bids,
+  acknowledge delivery, and issue an Ed25519 cross-signed purchase receipt
+  co-signed by the buyer. They also settle among themselves in a directed
+  cycle of ``payment_received`` receipts, which makes the honest guild a
+  strongly-connected anchor for the trust layer's collusion severance.
+* ``shill_seller-0`` — the adversary. Registers a ``sell`` card, decrypts
+  the bid, drains the stream, never delivers — and covers its tracks by
+  wash-trading fake ``purchase`` receipts with ``shill-0``. Under the
+  ``score_average`` reference trust plugin the fake receipts *raise* its
+  score; under ``agent_receipts`` the isolated mutual pair is severed and it
+  collapses to zero. That asymmetry is exactly what validator 3 catches.
+* ``shill-0`` — the wash-trading partner. Issues the mirror-image fake
+  receipts.
+* ``auditor-0`` — owns the single trust-plugin instance, ingests receipts
+  and negatives, and emits one ``score:`` line per roster agent at the end.
+
+Trace-marker protocol (all markers are broadcasts, so they are recorded in
+the trace at send time and cannot be lost to message drops):
+
+* ``discovered:<buyer>:<seller>`` — first gossip lookup that returned the seller.
+* ``bidmeta:<buyer>:<seller>:<ref>:<amount>`` — ground-truth sidecar for the
+  opacity validator. Would not exist in production; exists so the validator
+  can bind the invariant to a known plaintext.
+* ``stream:open:<buyer>:<seller>:<ref>:<rate>`` / ``stream:close:...:<total>``
+* ``fulfilled:<seller>:<buyer>:<ref>`` — seller acknowledged delivery.
+* ``receipt:<issuer>:<json>`` — receipt en route to the auditor.
+* ``negative:<buyer>:<seller>:<ref>`` — buyer reporting non-delivery.
+* ``score:<agent>:<score>:<confidence>`` — auditor's final trust scores.
+
+The wire bid itself is ``b"bid:" + envelope`` sent point-to-point; with
+``privacy: hybrid_x25519`` the envelope is ciphertext, with ``privacy: noop``
+it is plaintext and the opacity validator fails — the charter's adversarial
+bar, demonstrated per layer by swapping exactly one YAML line.
+
+Determinism: every agent uses the simulator's seeded RNG and logical clock;
+the privacy plugin runs in ``deterministic=True`` mode; receipt identities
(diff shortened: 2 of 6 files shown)


dhve:
Thanks @Abi5678. This is one of the most interesting submissions in the queue. The thesis held up when I went looking: I could not find an existing validator that ties two plugin layers together, and your four joint validators are a genuinely new kind of artifact here. I checked the load-bearing details against main and they hold up: `_seed_for` matches the `sha256(str(agent))` derivation inside `AgentReceiptsTrust._did_of`, the 0.3 threshold sits correctly between the severed shill's 0.0 and the single-guild-receipt honest score of `1 - exp(-5/10) ~ 0.39`, and the `_agent_plugins` override wiring is exactly how the runner pops it. The `bidmeta:` ground-truth sidecar is an honest answer to "how do you falsify opacity", and the exact-verdict discrimination tests (`test_noop_privacy_fails_exactly_the_opacity_validator` and its `score_average` twin) are the strongest form of the adversarial bar I have seen in this hackathon: they assert failure attribution, not just failure.

**1. You picked none of the ten problems, and the charter says pick exactly one.**

You say this openly in the body, so I will answer it openly: I am treating scenario-plus-validator composition work as in scope. The rubric explicitly lists "novel layer compositions" under novelty, and this PR is the first to cash that in. Keep that framing in the body so the judges see the reasoning, but know it was a judgment call, not a free pass; the letter of the charter is against you.

**2. The branch conflicts with main and no CI has run.**

`mergeable` is CONFLICTING. Since you branched, four more scenario suites merged (`failure_detection`, `parc_migration`, `rogue_trusted_agent`, `sybil_bond`), and all three files you touch outside your new modules are append hotspots: the `_try_load_builtin` chain in `scenarios.py`, the `VALIDATORS` dict, and the expected set in `test_all_scenario_types_registered`. The rebase is mechanical, but until it lands GitHub reports no checks on the branch, and the judge panel does not score broken PRs.

**3. `SEND_REDUNDANCY = 3` leaves a real tail across the seed bank.**

Your own math gives ticks a 0.05^5 budget, but bids get 3 copies, so a bid dies entirely with p = 1.25e-4. When that happens the buyer's stream still drains (`open_stream` debits the first tick immediately, so every closed stream has total > 0), no `fulfilled:` marker ever exists, and the honest seller lands in the offender set of `commerce_undelivered_penalized` at score 0.39, above threshold, so the validator fails. Four honest-seller bids per run makes that roughly 1 in 2000 per seed. Your tests pin seeds 42 and 7, but judging re-runs the scenario under the seed bank. Cheapest fix: give bids (and acks) the same redundancy 5 as ticks, and note the budget in the docstring.

**4. The validators trust marker bodies but not marker provenance.**

`_pc_marker_events` never reads the event's `agent` field, so any agent can utter any marker. Most forgeries fail closed (a fake `fulfilled:` for the shill's ref just flips the penalty validator to "the adversary never fired"), but `score:` is last-write-wins in both trust validators: under the `score_average` swap, a byzantine agent broadcasting `score:shill_seller-0:0.000000:0.000000` after the auditor finalizes would overwrite the real 0.75 and false-pass `commerce_undelivered_penalized`. Nothing in this scenario exploits it, but the whole thesis of the PR is that byzantine parts do not compose for free. Broadcast trace events carry the sender, so bind each marker to its expected principal: `score:` only from the auditor, `fulfilled:<seller>` only when the broadcaster is that seller.

**Suggested path forward:**

1. Rebase onto main and re-append your three registry hunks.
2. Bump the bid/ack redundancy to 5 per finding 3.
3. Add sender-provenance checks to the four validators, plus one unit test with a forged `score:` line.
4. Run the five CI stages locally, push, and confirm the GitHub checks go green.

Leaving this open. Ping me on the hackathon channel if you want a second pair of eyes on the rebase.

JamesCarnley:
Thank you, @Abi5678. A future cross-layer private-commerce path should prove engine-attributed discovery before bidding, non-vacuous wire opacity, payment only for delivered units across multiple ticks, and reputation derived from independently attributed outcomes; swapping one owning layer should fail that exact stage. Current Town already composes small scoped scenarios, but the four legacy plugins used here are absent, and this frozen validator stack relies on participant-authored markers, can pass bid opacity without bid metadata, and observes only an initial payment tick. I am closing this obsolete implementation while retaining the cross-layer discriminators for a future current-code profile. No contributor code was merged or executed during this review.

PR #60, opened 2026-07-03
[Hackathon] hiten: add configurable provenance_supply_chain_multi scenario

Hiten0305l:
## Summary

This is a focused follow-up to my previous hackathon PR (#39). Following the merge of the DataFacts implementation in #31, this PR contributes the independent scenario enhancement suggested during review by introducing a configurable provenance supply-chain scenario.

The new `provenance_supply_chain_multi` scenario exercises the existing `cid_facts` implementation over a **parameterized multi-stage provenance DAG**, supporting configurable supplier and manufacturer stages while preserving the existing `provenance_supply_chain` (diamond topology) unchanged. Together, both scenarios exercise the same provenance implementation across complementary graph structures without modifying plugin behaviour.

## What
- Add a new built-in `provenance_supply_chain_multi` scenario supporting configurable `num_suppliers` and `num_manufacturers`.
- Add a dedicated `provenance_supply_chain_multi.yaml` for configuring parameterized graph sizes.
- Register the scenario independently alongside the existing provenance scenarios.
- Register a dedicated validator entry that reuses the existing provenance validator suite without modifying validator behaviour.
- Add comprehensive test coverage for configurable topologies, deterministic execution, provenance validation, invalid configuration handling, and regression protection.

### Topology
```text
                   supplier-0
                   supplier-1
                      ...
                supplier-(N-1)
                      │
        ┌─────────────┼─────────────┐
        ▼             ▼             ▼
  manufacturer-0 manufacturer-1 ... manufacturer-(M-1)
        └─────────────┬─────────────┘
                      ▼
                distributor-0
                      ▼
                  retailer-0
```
Each manufacturer aggregates provenance from **all supplier datasets**, producing a configurable multi-stage provenance DAG. The distributor aggregates all manufacturer outputs before forwarding the resulting lineage to the retailer for provenance verification and adversarial validation.

## Why
The existing `provenance_supply_chain` scenario validates provenance using a fixed diamond-shaped dependency graph. This PR introduces a complementary configurable graph topology that exercises the same `cid_facts` implementation under a different dependency structure without changing the underlying plugin.

Supporting configurable numbers of suppliers and manufacturers allows the framework to exercise the same provenance implementation across multiple graph sizes and dependency structures, expanding scenario coverage for:

- Provenance lineage construction and traversal
- Parent aggregation across multiple upstream datasets
- Provenance verification across configurable graph topologies
- Deterministic replay for configurable scenario configurations
- Adversarial validation using the existing validator suite

This expands scenario coverage while preserving backward compatibility and leaving the existing provenance implementation unchanged.

## Scope
This PR is intentionally limited to introducing the new `provenance_supply_chain_multi` scenario, its configuration, registry integration, and associated tests.

## Verification
```bash
make ci-local

uv sync
uv run ruff check .
uv run ruff format --check .
uv run pyright
uv run pytest -v
```

### DIFF
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -109,6 +109,12 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("provenance_supply_chain", provenance_supply_chain_factory)
+    elif name == "provenance_supply_chain_multi":
+        from nest_core.scenarios_builtin.provenance_supply_chain_multi import (
+            provenance_supply_chain_multi_factory,
+        )
+
+        register_scenario("provenance_supply_chain_multi", provenance_supply_chain_multi_factory)
     elif name == "bft_hotstuff":
         from nest_core.scenarios_builtin.bft_hotstuff import bft_hotstuff_factory
 
--- a/packages/nest-core/nest_core/scenarios_builtin/provenance_supply_chain_multi.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/provenance_supply_chain_multi.py
@@ -1,0 +1,465 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Content-addressed provenance scenario: a configurable N×M fan-in pipeline, then attacks.
+
+Topology (configurable via ``task.config`` in the YAML)::
+
+    supplier-0  supplier-1  ...  supplier-(N-1)
+         \\            |              /
+          \\           |             /
+       manufacturer-0  ...  manufacturer-(M-1)
+               \\       ...       /
+                \\               /
+               distributor-0   (join: parents = all manufacturer URLs)
+                      |
+                 retailer-0
+
+Each supplier publishes a raw dataset.  Every manufacturer waits to receive
+*all* supplier inputs (``num_suppliers`` messages), then publishes one derived
+dataset whose ``parents`` list every supplier URL, and forwards the result to
+the distributor.  The distributor waits for *all* manufacturers (``num_manufacturers``
+messages) and publishes a join dataset whose ``parents`` list every manufacturer
+URL.  The retailer acts as verifier-and-attacker: it walks the full provenance
+DAG and then runs the three canonical attacks.
+
+This gives ``validate_trace(..., "provenance_supply_chain_multi")`` a different
+graph shape from the fixed diamond in ``provenance_supply_chain``: fan-width is
+configurable from YAML, not hard-coded to 2.
+
+Example (from YAML ``task.config``)::
+
+    task:
+      type: provenance_supply_chain_multi
+      config:
+        num_suppliers: 3
+        num_manufacturers: 2
+
+Defaults to ``num_suppliers=2, num_manufacturers=2``.
+"""
+
+from __future__ import annotations
+
+from typing import Any, cast
+
+from nest_core.scenario import ScenarioConfig
+from nest_core.sim.agent import AgentContext, StateMachineAgent
+from nest_core.types import AgentId, DatasetMetadata
+
+_PHANTOM_PARENT = "df://sha256-" + "0" * 64
+
+# ---------------------------------------------------------------------------
+# Shared helper
+# ---------------------------------------------------------------------------
+
+
+def _parents_of(meta: DatasetMetadata) -> list[str]:
+    """Read declared provenance parents off a dataset as a plain list of URL strings.
+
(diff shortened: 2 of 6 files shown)


dhve:
Thanks @Hiten0305l. This is exactly the scoped-down follow-up from the #39 close-out, delivered to spec: `provenance_supply_chain_multi` gets its own YAML and its own key in the VALIDATORS registry, the merged diamond scenario stays untouched (and you prove it with the depth-4 regression test), and the test matrix is genuinely good: 2x2, 3x2, 1x3, and the degenerate 1x1 topologies, a `datafacts_v1` negative test that pins the exact validator split, and byte-identical determinism across three seeds. The NOTE in the YAML explaining that `agents.count` is not load-bearing for the state-machine path is the kind of documentation that saves the next reader real time.

This is close. One real bug to fix, plus two smaller things:

**1. The substitution attack can grade against the wrong supplier's URL.**

`MultiRetailerAgent._verify_chain` returns `sorted(roots)[0]` and `_attack_substitution` compares the attacker's republished URL against that. In the diamond there is exactly one root, so this was precise. In your scenario there are `num_suppliers` roots, and under `cid_facts` the lexicographically first content hash is an arbitrary supplier. I ran your 3x2 config at seed 42: the forge targets supplier-0 (`raw_material_0`), but the `source_url` in the `attack_substitution` trace line is supplier-1's URL.

Your tests still pass because both reference plugins get lucky: `cid_facts` hashes the tampered description so `collided` is 0 against any URL, and `datafacts_v1`'s name-addressed `df://raw_material_0` happens to sort first. But consider the plugin this check exists to catch: one that content-addresses only `name` and `owner` and ignores the rest. Its forged URL lands exactly on supplier-0's real URL (a successful substitution), and whenever supplier-0's hash is not the one that sorts first, the comparison runs against a different root, `collided` reports 0, and `provenance_substitution_resistant` passes on a vulnerable plugin. That is a false pass on the precise attack class the scenario exists to catch, so it goes straight to the correctness dimension.

The fix is small and makes the check strictly stronger than the diamond's single-root comparison: have `_verify_chain` hand back the full root set and check membership:

```python
collided = int(str(attacker_url) in roots)
```

For test rigor, the gold-standard regression is a tiny in-test fake datafacts plugin that addresses by name and owner only, asserting `provenance_substitution_resistant` now fails against it at `num_suppliers=3`.

**2. CI has never run on this branch.**

`gh pr checks 60` reports no checks at all, and the branch is cut from the #38-era main, 52 commits behind by now. Rebase onto (or merge) current main and push; that triggers the pipeline and picks up the four scenario names main has since added to the expected set in `test_validators.py` (`failure_detection`, `parc_migration`, `rogue_trusted_agent`, `sybil_bond`). The merge resolves cleanly, I checked. For what it's worth, I merged your branch into today's main locally and ran the full CI set: `ruff check`, `ruff format --check`, `pyright`, and `pytest` all green (1018 passed). So this is bookkeeping, not surgery, but the judge panel does not score PRs without green CI on record.

**3. Small test-file hygiene.**

In `test_provenance_supply_chain_multi.py`, the `type Event = dict[str, Any]` alias is never used, and the two `TestFactoryValidation` methods import `pytest` and re-import `ScenarioConfig` function-locally when both can live at module top. Also, the walk in `_verify_chain` is depth-first (stack plus pop) while the docstring and the 3x2 test comment call it breadth-first; the diamond has the same wording so no blame, but since you are editing anyway, "full traversal" sidesteps the question.

**Suggested path forward:**

1. Return the root set from `_verify_chain` and switch `_attack_substitution` to the membership check.
2. Add the adversarial regression test with a name-and-owner-addressed fake plugin.
3. Rebase onto current main, push, and confirm CI goes green.
4. Sweep the unused alias and the function-local imports.

Leaving this open. This is the most substantial of your three follow-ups from the #39 split, and once the substitution target is fixed it is in good shape. Ping me on the hackathon channel if anything is unclear.

JamesCarnley:
Thank you, @Hiten0305l. A future provenance DAG must bind each delivered artifact to the correct supplier root and parent chain, and root or edge substitution must fail regardless of lexical ordering in a multi-root graph. The frozen negative control compares the forged URL with sorted(roots)[0], which can be a different supplier's valid root and therefore pass for the wrong reason. Current Town has signed flat evidence records but no artifact DAG or multi-root traversal contract. I am closing this obsolete scenario while retaining the correct-supplier substitution case for any future DAG profile. No contributor code was merged or executed during this review.

PR #65, opened 2026-07-04
RoboAgent Guard for Safety and Privacy Validation

subhashchandra001:
## Summary [Updated]

RoboAgent Guard targets **NandaHack Phase 1 Problem 09: privacy**.

It adds a deterministic `privacy: sensor_redaction` plugin for robot-agent sensor traces, plus a `roboagent_guard` scenario and adversarial validators. The scenario demonstrates that the reference `privacy: noop` plugin leaks raw sensor payloads and fails validation, while `privacy: sensor_redaction` redacts sensitive sensor terms and passes.

## Motivation

Robot-agent traces may contain privacy-sensitive observations such as `raw_camera`, `camera_frame`, `lidar_scan`, `person_detected`, `private_zone`, and `map_save`.

A privacy layer should prevent those raw sensor terms from being shared directly in the trace. This PR adds a narrow, deterministic sensor-redaction privacy plugin that is different from the already merged hybrid encryption work: it focuses on auditable robot-sensor redaction policy, not general-purpose encryption.

## Changes

  - Adds `SensorRedactionPrivacy` as `privacy: sensor_redaction`.
  - Registers the plugin in the built-in plugin registry and entry points.
  - Adds the `roboagent_guard` scenario factory.
  - Adds runnable scenario YAML at `scenarios/roboagent_guard.yaml`.
  - Adds adversarial validators for:
    - robot command safety markers
    - prior supervisor approval before robot commands
    - privacy filtering on sensitive sensor messages
  - Updates validator matching to use exact tokens instead of substring matching.
  - Adds adversarial tests for:
    - `unsafe_action`
    - `not_supervisor_approved`
    - `risk_checked_skipped`
    - self-attested supervisor approval
    - `privacy: noop` leaking raw sensor payloads
  - Adds documentation at `docs/hackathon/roboagent-guard.md`.
  - Updates privacy-layer docs with `sensor_redaction`.

## Adversarial Path

The reference plugin fails:

  ```yaml
  privacy: noop

With noop, raw sensor terms such as raw_camera, person_detected, and private_zone remain in the trace without privacy markers, so the RoboAgent Guard privacy validator fails.

The new plugin passes:

  privacy: sensor_redaction

With sensor_redaction, sensitive sensor tokens are replaced with redaction markers and stamped with privacy_filtered=<action_id>, no_raw_storage, and redacted.

## Verification

Branch is cleanly based on latest upstream/main with one RoboAgent Guard commit.

  make ci-local

Result:

1298 passed, 1 skipped, 1 deselected, 1 warning
ci-local: all 5 checks passed

Focused checks also pass:

uv run pytest packages/nest-core/tests/test_validators.py::TestRoboAgentGuard packages/nest-core/tests/test_scenarios.py::TestRoboAgentGuardScenario packages/nest-plugins-reference/tests/test_sensor_redaction.py -q

Result:

  16 passed

## Scope

This PR is limited to Problem 09 privacy: deterministic robot-sensor redaction, the paired scenario, adversarial validators, docs, and tests.

### DIFF
--- a/docs/hackathon/roboagent-guard.md
+++ b/docs/hackathon/roboagent-guard.md
@@ -1,0 +1,441 @@
+# RoboAgent Guard
+
+RoboAgent Guard is a NANDA Town Phase 1 contribution targeting
+`docs/hackathon/problems/09-privacy-hybrid-encryption-with-selective-disclosure.md`.
+It adds a narrow privacy-layer differentiator for robot sensor traces:
+`privacy: noop` leaks raw sensor payloads, while `privacy: sensor_redaction`
+filters those payloads before they are shared.
+
+The goal is not to duplicate the already-merged `hybrid_x25519` work. This
+contribution is deliberately narrower: deterministic redaction of camera,
+person-detection, private-zone, lidar, and mapping tokens, plus trace validators
+that prove the reference `noop` privacy plugin fails this sensitive workflow.
+
+## What This Adds
+
+This contribution adds five pieces:
+
+- A scenario type named `roboagent_guard`.
+- A runnable YAML scenario at `scenarios/roboagent_guard.yaml`.
+- A privacy plugin named `sensor_redaction`.
+- Two validators registered under `roboagent_guard`.
+- Unit and end-to-end tests for safe and unsafe traces.
+
+The plugin and validators are deterministic. They do not call external services,
+require secrets, or use nondeterministic behavior.
+
+## Files
+
+| File | Purpose |
+| --- | --- |
+| `packages/nest-core/nest_core/scenarios_builtin/roboagent_guard.py` | Built-in scenario factory and simple state-machine agents. |
+| `scenarios/roboagent_guard.yaml` | Runnable scenario configuration. |
+| `packages/nest-core/nest_core/scenarios.py` | Registers the scenario factory loader. |
+| `packages/nest-core/nest_core/validators.py` | Adds RoboAgent Guard validators and registry entry. |
+| `packages/nest-plugins-reference/nest_plugins_reference/privacy/sensor_redaction.py` | Privacy plugin that redacts sensitive sensor payloads. |
+| `packages/nest-plugins-reference/tests/test_sensor_redaction.py` | Unit tests for the privacy plugin and registry wiring. |
+| `packages/nest-core/tests/test_validators.py` | Unit tests for validator pass/fail cases. |
+| `packages/nest-core/tests/test_scenarios.py` | End-to-end scenario tests. |
+| `docs/hackathon/roboagent-guard.md` | This guide. |
+
+## Architecture
+
+The flow is:
+
+1. `ScenarioConfig.from_yaml("scenarios/roboagent_guard.yaml")` loads the YAML.
+2. `ScenarioRunner` reads `task.type: roboagent_guard`.
+3. `get_scenario_factory("roboagent_guard")` loads the built-in factory from
+   `nest_core.scenarios_builtin.roboagent_guard`.
+4. The factory creates four deterministic state-machine agents:
+   - `planner-0`
+   - `vision-0`
+   - `robot-0`
+   - `mapper-0`
+5. The supervisor sends prior approval for the planner's action.
+6. The planner sends a robot-motion command to the robot.
+7. The vision agent passes raw sensor data through `ctx.plugins["privacy"]`.
+8. With `privacy: sensor_redaction`, the trace contains redacted sensor data.
+   With `privacy: noop`, the trace leaks raw sensor data.
+9. `validate_trace(..., "roboagent_guard")` runs both RoboAgent Guard validators
+   against that trace.
+
+## Scenario Agents
+
+The scenario is intentionally small so it is easy to audit.
+
+| Agent | Role |
+| --- | --- |
+| `supervisor-0` | Sends prior approval for robot action `nav-1`. |
+| `planner-0` | Sends a robot autonomy command such as `cmd_vel` and `navigation_goal`. |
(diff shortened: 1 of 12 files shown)


dhve:
Thanks @subhashchandra001. This is clean, careful work: the `roboagent_guard` scenario factory with its scripted `_ScriptedSender` / `_AckSink` agents is deterministic by construction, the wiring in `scenarios.py` and the `VALIDATORS` registry follows the existing pattern exactly, and the hygiene is all there (SPDX headers, `from __future__ import annotations`, `Example::` blocks, e2e tests covering both the safe YAML run and the unsafe `from_dict` run). The fail-on-empty-trace design in both validators is a genuinely nice touch; vacuous passes are a real failure mode in trace validators and most submissions miss it.

Not ready to merge as-is though. Three things to work through:

**1. It doesn't map to a hackathon problem, and the adversarial-validator requirement isn't met yet.**

The charter (docs/hackathon/charter.md) asks you to pick exactly one problem from `docs/hackathon/problems/` and ship an adversarial validator that catches an attack class the reference plugin would fail. Your own doc calls this a "warm-up contribution", and none of the 10 problems covers robot safety. Right now the unsafe mode is the same scripted agents omitting the strings the validator greps for, so it proves the validator reads text, not that any layer behavior is being guarded. Notice your YAML sets `privacy: noop`: the privacy layer never actually does anything in this scenario. The closest home is problem 09 (privacy). Reframed there, the guard becomes real: a privacy plugin that actually filters or redacts sensor payloads, with your validators as the adversarial check that the noop reference plugin fails and yours passes. One caution: a hybrid-encryption take on problem 09 already merged as #28, so the sensor-redaction policy itself would need to be your differentiator. Which problem were you targeting? Worth stating explicitly in the PR body either way.

**2. Marker matching is substring-based and spoofable (correctness).**

In `_roboagent_marker_matches_action`, when the message carries no `action_id` the fallback is a plain `marker in msg`. I ran these against your logic and all three pass the safety validator:

```
cmd_vel:0.9 unsafe_action            # "safe_action" is a substring of "unsafe_action"
cmd_vel:0.9 not_supervisor_approved
cmd_vel:0.9 risk_checked_skipped
```

Match tokens, not substrings: something like `marker in msg.replace(":", " ").split()`. Then add exactly these adversarial strings as failing tests. Test rigor is graded, and right now nothing in the test file feeds the validators a hostile string; the closest is the `risk_checked=nav-2` mismatch case.

**3. Markers are self-attested by the risky sender.**

The agent sending `cmd_vel` also asserts `supervisor_approved=nav-1` about itself. An adversarial agent just appends the marker; nothing in the trace has to show a supervisor ever spoke. A deterministic fix that stays within trace validation: require the approval to appear in an earlier message from a different agent (a supervisor role) referencing the same `action_id`, i.e. approval-before-execution ordering. Your Future Improvements list already names this; I would pull it into this PR, because it is the difference between a convention check and a validator with teeth.

One mechanical thing: GitHub reports the branch as conflicting with `main`, almost certainly the `VALIDATORS` registry and the `test_all_scenario_types_registered` set, since those lines are contested in nearly every hackathon PR. No CI checks have run on this branch yet either; rebase onto latest main so the pipeline can go green, since the judge panel only scores green PRs.

**Suggested path forward:**

1. Pick a problem from `docs/hackathon/problems/` (09 privacy looks like the natural fit) and state it in the PR body.
2. Move enforcement into a plugin (e.g. a privacy layer that filters `camera_frame` / `private_zone` payloads) and keep your validators as the adversarial check the noop reference fails.
3. Fix the substring matching to token-level and add the spoof strings above as failing tests.
4. Add supervisor-message ordering for approvals.
5. Rebase on main and confirm all five CI stages green locally before requesting re-review.

Leaving this open. Ping me on the hackathon channel if the reframing under problem 09 is unclear.

subhashchandra001:
Updated based on review.

This PR now explicitly targets hackathon Problem 09: privacy. It adds a deterministic `privacy: sensor_redaction` plugin for robot sensor traces, wires it into plugin discovery, and updates the `roboagent_guard` scenario so `privacy: noop` leaks raw sensor payloads and fails the adversarial validators while `privacy: sensor_redaction` passes.

Also fixed the validator issues:
  - marker matching is token-based, not substring-based
  - added tests for `unsafe_action`, `not_supervisor_approved`, and `risk_checked_skipped`
  - robot commands now require prior approval from a different supervisor agent
  - added safe/unsafe scenario tests and sensor-redaction plugin tests

Merged latest `upstream/main` and verified locally with `make ci-local`.

Result: 1298 passed, 1 skipped, 1 deselected, 1 warning. All 5 checks passed.

subhashchandra001:
Updated again to match the requested mechanical cleanup.

The branch is now cleanly based on latest `upstream/main` with one RoboAgent Guard commit, not a merge commit. I re-ran `make ci-local` after the clean rebase.

Result: 1298 passed, 1 skipped, 1 deselected, 1 warning. All 5 checks passed.

subhashchandra001:
The branch is now cleanly rebased onto latest `upstream/main`, and I ran `make ci-local` locally.

  Result: 1298 passed, 1 skipped, 1 deselected, 1 warning. All 5 checks passed.

  Could you please approve the pending GitHub Actions workflow when you get a chance?

JamesCarnley:
Thank you, @subhashchandra001. The adopted invariant is already covered in rebuilt Town: fields declared private are recursively redacted before export, and the evaluator fails on a literal nested leak rather than merely observing that a redactor ran. The RoboAgent token-scanning policy inside opaque sensor strings is a stronger domain-specific claim that Town does not currently make. I am closing this legacy PR as covered for the declared-field behavior, while retaining token-in-opaque-payload as a future robotics/privacy profile case. No contributor code was merged or executed during this review.

PR #53, opened 2026-07-01
[Hackathon] moltpass.dev — bonded_trust — a Sybil-resistant trust root with a pluggable scarcity anchor

moltpass:
[Hackathon] Moltpass.dev — bonded_trust — a Sybil-resistant trust root with a pluggable scarcity anchor

## Summary

**Stop fake-identity swarms from buying reputation for free.** In an agent marketplace a Sybil can spin up 10,000 accounts that vouch for each other and ride the reputation they manufacture. `bonded_trust` makes a trust *root* cost a scarce resource you can't fake by minting — credits, proof-of-work, or a consensus grant — so free identities stay worthless no matter how many you create. A squirrel doesn't trust another squirrel because it *says* it has nuts; it trusts the one that actually stashed some.

## The gap being filled here

The merged/open trust plugins (score_average, EigenTrust-style plugins, agent_receipts) ration influence among identities that already exist; none stops the free *minting* of identities. The EigenTrust submissions concede the boundary:

> "if NEST's auth/identity layers fail upstream, no trust layer can save you."

NEST's `did_key` mints identities for free, so upstream *has* failed by default. `bonded_trust` covers that seam. It does **not** claim to beat Douceur's Sybil impossibility result (2002) — and it takes Douceur seriously: the resource must be *genuinely scarce and verified*, not self-asserted. So the plugin never trusts a self-declared bond.

## Mechanism

- **Ledger-backed self-bond (defeats free minting).** An identity's trust root is pinned at the untrusted floor (score 0.0) until it *reserves* a bond, and the reservation goes through a `StakeLedger`. If the agent can't afford it, no bond is recorded. A broke Sybil that sends `bond:1000000` gets nothing.
- **Reporter-weighting (defeats wash-trading).** Reports count in proportion to the reporter's bond; unbonded reporters carry zero weight; self-reports are ignored. Splitting a fixed budget across K identities buys the same influence as holding it in one.

## The pluggable scarcity anchor (payments · PoW · consensus)

The plugin is agnostic about *what makes the bond scarce* — that's a `StakeLedger` you inject. Two ship in `trust/stake_ledgers.py`:

- `CreditBackedLedger` — scarcity = the **payments layer's** finite credits.
- `ProofOfWorkLedger` — scarcity = **CPU**: each bond unit needs a sha256 proof-of-work, so minting K identities costs K× the hashing. No payments layer, no trusted mint. Deterministic mining keeps traces reproducible.

[...]

PR #67, opened 2026-07-05
[Hackathon] vladimirputkov: verified_registry — signature-verified registration with typed rejections and a Sybil scenario

VladimirPutkov:
# [Hackathon] vladimirputkov: verified_registry — signature-verified registration with typed rejections and a Sybil scenario

**Branch:** `hackathon/vladimirputkov-registry-integrity`  
**Base:** `main`  
**Layer:** registry  
**Scope:** self-scoped registry improvement because Problem 06 was already completed by #24

## Problem

The default `in_memory` registry trusts every call to `register(card)`.

An attacker can therefore:

- register an `AgentCard` under another agent’s `agent_id`;
- silently overwrite the authentic card;
- advertise capabilities it has not proved it owns;
- mutate a valid card after signing it.

The identity layer already provides signing plugins such as `did_key` and `ed25519_rotating`, but the default registration path does not use them.

This PR adds an authenticated registration boundary to the registry layer.

### `registry: verified`

`VerifiedRegistry` is a drop-in, non-replicated `Registry` implementation registered through the `nest.plugins.registry` entry-point group.

A card is admitted only if it carries a valid identity signature from the same agent named by its `agent_id`.

The signature covers canonical card bytes containing:

- `agent_id`;
- `name`;
- sorted, order-independent `capabilities`;
- `endpoint`.

`metadata` is deliberately excluded because it carries the signature and optional identity fields such as `key_id` and `signed_at`.

Cryptographic verification remains delegated to the configured `Identity` plugin. The registry itself implements no cryptography.

### Typed rejection reasons

Rejected registrations raise `RegistrationRejectedError` with a stable reason:

| Attack or failure | Reason |
|---|---|
| Missing or malformed signature | `missing_signature` |
| Signature created by an agent other than the claimed `agent_id` | `signer_mismatch` |
| Forged signature, unknown signer, or post-signing card mutation | `bad_signature` |

A rejected registration is atomic: it does not modify the store or notify subscribers.

## Adversarial scenario

`scenarios/registry_integrity.yaml` runs seven deterministic agents with seed `42`:

- three honest agents sign and register their own cards;
- one Sybil agent attempts impersonation;
- one Sybil agent submits an unsigned card;
- one Sybil agent signs its own card and then mutates it to claim an honest capability;
- one auditor checks the resulting discovery state.

The auditor records four properties for every honest capability:

- whether the authentic owner remains present;
- whether the owner’s card still points to its authentic endpoint;
- whether another `agent_id` is squatting on the capability;
- whether each attack was accepted or rejected with the correct typed reason.

## Validators

`VALIDATORS["registry_integrity"]` contains four implementation-independent checks:

- `registry_unauthenticated_rejected` — every adversarial attempt must be rejected and the trace must contain real attacks;
- `registry_honest_admitted` — honest signed registrations must still succeed;
- `registry_rejection_reasons` — each attack must produce the correct typed reason;
- `registry_authentic_discovery` — every honest owner must remain discoverable and authentic, with no capability squatting.

The validators judge observable trace behavior rather than inspecting `VerifiedRegistry` internals.

## Differential proof

The same scenario is tested against both registries.

### `registry: verified`

- all honest registrations are accepted;
- all three attacks are rejected;
- every rejection carries the expected typed reason;
- authentic discovery is preserved;
- `squatting=0`;
- all four validators pass.

### `registry: in_memory`

The scenario still completes without crashing, but:

- all three attacks are accepted;
- the victim’s card is overwritten;
- the poisoned lookup reports `present=1, authentic=0`;
- the tampered Sybil card claims the victim’s capability;
- the audit reports `squatting=1`;
- the adversarial validators fail.

This is the baseline failure required by the hackathon charter: the test fails without the improvement and passes with it.

## Positioning relative to #148

PR #148 introduced `byzantine_gossip`, which protects a replicated gossip registry by re-verifying signed writes on every hop, detecting signed equivocation, quarantining publishers, and reducing eclipse risk.

This PR addresses a narrower and complementary boundary:

| | `verified` — this PR | `byzantine_gossip` — #148 |
|---|---|---|
| Scope | Local registration | Replicated gossip mesh |
| Primary threat | Unauthenticated registration | Forged or equivocated gossip writes and eclipse attempts |
| Signature verification | At `register()` | Across gossip propagation |
| Replication | None | Gossip |
| Equivocation detection | Out of scope | Detection and quarantine |
| Rejection reporting | Three typed reasons | Gossip-specific rejection handling |
| Validators | Portable trace validators | Gossip-specific validators |

This PR does not claim gossip security, replication, equivocation detection, or eclipse resistance. Its contribution is a small authenticated registry implementation, typed registration failures, and a reusable differential scenario that can evaluate registry behavior through public interfaces.

## Tests

Coverage includes:

- protocol conformance with the runtime-checkable `Registry` interface;
- plugin entry-point resolution;
- all three typed rejection reasons;
- spoofed signer metadata;
- unknown signers;
- post-signing mutation;
- rejection atomicity;
- subscriber behavior;
- canonical capability ordering;
- metadata independence;
- `key_id` and `signed_at` pass-through;
- Hypothesis completeness and mutation-soundness properties;
- non-vacuous validator behavior;
- capability-squatting detection;
- sender/receiver trace deduplication;
- verified-versus-baseline end-to-end behavior;
- byte-identical traces for repeated runs with the same seed.

## Determinism

The scenario uses:

- no wall-clock decisions;
- no unseeded randomness;
- fixed identity seeds derived from agent IDs;
- sorted canonical card fields;
- one scheduled audit pulse.

Repeated runs with the same seed produce byte-identical traces.

## Local CI evidence

The exact five commands from `make ci-local` were run in order:

```text
uv sync                        PASS
uv run ruff check .            PASS — all checks passed
uv run ruff format --check .   PASS — 216 files already formatted
uv run pyright                 PASS — 0 errors, 0 warnings
uv run pytest -v               PASS — 1179 passed, 1 skipped, 1 deselected
```

`make` itself was unavailable in the Windows shell, so the commands defined by the target were executed directly in the same order.

## Deliberate limitations

- `deregister(agent)` has no authentication material in the current `Registry` protocol, so authenticated eviction would require a protocol change.
- Verification trusts the public keys provided by the configured identity layer.
- Signature freshness and replay protection across separate runs are out of scope.
- `key_id` and `signed_at` are preserved so identity implementations can enforce rotation or freshness policies.
- `VerifiedRegistry` is not replicated and does not inspect gossip traffic or detect signed equivocation.

## Reproduce

```bash
uv sync
uv run ruff check .
uv run ruff format --check .
uv run pyright
uv run pytest -v
```

Focused test suite:

```bash
uv run pytest \
  packages/nest-plugins-reference/tests/test_verified_registry.py \
  packages/nest-plugins-reference/tests/test_verified_registry_properties.py \
  packages/nest-core/tests/test_registry_integrity.py \
  -v
```

Run the protected scenario:

```bash
uv run nest run scenarios/registry_integrity.yaml
```

Validate its trace:

```bash
uv run python -c "
from pathlib import Path

for result in validate_trace(
    Path('traces/registry_integrity.jsonl'),
    'registry_integrity',
):
    print(
        'PASS' if result.passed else 'FAIL',
        result.name,
        '-',
        result.detail,
    )
"
```

For the baseline differential, change one line in the scenario:

```yaml
registry: in_memory
```

Then rerun the scenario and validators. The run completes, but the adversarial checks fail and the trace records both a poisoned owner card and capability squatting.


dhve:
Thanks @VladimirPutkov. This is one of the cleanest registry submissions in the queue. `canonical_card_bytes` is the right core design: order-independent capabilities, sorted keys, metadata deliberately excluded so the signature can ride there, which makes the three typed rejections in `VerifiedRegistry.register` fall out naturally. The tests show real thought about the attack surface: `test_spoofed_signer_field_rejected_bad_signature` (rewrite the signer field to the victim after signing with your own key) is the case naive implementations miss, and asserting the `in_memory` differential run end-to-end (attacks accepted, `honest-0` poisoned, validators fail, no crash) is the "fails on baseline" proof the charter asks for. I also checked your wiring against main: `plugins.py`, the `_agent_plugins` override channel, and the elif chain all match the `gossip`/`identity_rotation` precedents.

Not mergeable as-is though, mostly logistics rather than design. Four things to work through:

**1. The branch conflicts with main, and CI has never run on it.**

GitHub reports the PR as CONFLICTING. The causes are concrete: #53 and #149 merged after your branch point and hit the same three registration points you touched: the `VALIDATORS` dict in `nest_core/validators.py`, the expected-set in `tests/test_validators.py::test_all_scenario_types_registered`, and the elif chain in `nest_core/scenarios.py`; #148 and #138 landed on `scenarios.py` and `plugins.py` too. Rebase onto current main, re-add your entries alongside the new ones, and push. Zero checks are reported on `hackathon/vladimirputkov-registry-integrity`, so nobody can see your five stages green, and the judge panel does not score broken PRs.

**2. Update the novelty framing: map to the problem set, and account for #148.**

The charter says pick exactly one problem from `docs/hackathon/problems/`, and the only registry problem there (06) is gossip eventual consistency, already merged as #24. Your motivation cites both, and you hit every structural requirement, but the mapping to the brief is implicit; add one sentence up front: registry layer, self-scoped because 06 is taken. Bigger: the body's claim that registration authenticity is "untouched by any open or merged PR" is now stale. #148, a byzantine-resistant gossip registry with signed cards, just merged; its body cites #67 as the prior art it extends. Your differentiators survive (a registry-agnostic gate with typed rejection reasons, validators that judge any registry, the one-line differential YAML), but say so yourself; novelty is scored against merged peers.

**3. The discovery audit never measures the tamper attack's actual payoff.**

In `registry_integrity.py`, the auditor's `reg:lookup:` line only inspects the card whose `agent_id` equals the honest agent's. Under `in_memory`, sybil-2's accepted tamper card (its own id, carrying `service:honest-0`) also matches the capability query, but `validate_registry_authentic_discovery` never sees it; the failing run is caught there only because the impersonate and unsigned overwrites flip `authentic=0`. Add a `squatting=<0|1>` field (a different `agent_id` claiming the honest capability) and assert it in the validator. That makes all three attack kinds load-bearing at the discovery level, a cheap test-rigor win.

**4. Keep your validators in one block and drop the stray whitespace edits.**

Your four validators land in `validators.py` as two non-contiguous chunks, with `validate_registry_authentic_discovery` separated from its siblings by `validate_failure_detection_accuracy`, plus a stray blank line at the failure-detection separator and a duplicated `# ---` separator above the PARC section. Cosmetic, but it bloats the conflict surface in the repo's most conflict-prone file. Since you are rebasing anyway, consolidate into one contiguous "Registry integrity validators" section and drop the whitespace-only hunks.

**Suggested path forward:**

1. Rebase onto main; resolve `plugins.py`, `validators.py`, `test_validators.py`, `scenarios.py` against the entries added since your branch point (#53, #138, #148, #149).
2. While in there, consolidate the validator block and revert the two stray whitespace edits (finding 4).
3. Add the `squatting` field to the auditor line and the discovery validator, with a synthetic-trace test for it (finding 3).
4. Update the PR body: the problem-mapping sentence and the positioning against #148 (finding 2).
5. Run the five CI stages locally and confirm green once checks appear on the PR.

VladimirPutkov:
Thanks @dhve  I’ve addressed all four review items.

1. The branch is updated against current main. GitHub now reports PR #67 as mergeable, and the registry_integrity entries coexist with the recently merged scenario, validator, and plugin registrations.

2. I updated the positioning to make the scope explicit: this is a self-scoped registry-layer improvement because Problem 06 was already taken by #24. I also removed the stale novelty claim and positioned `verified` accurately relative to #148: `verified` is a registry-agnostic, non-replicated admission gate with typed rejection reasons and portable validators, while `byzantine_gossip` protects replicated gossip writes, detects equivocation, and adds eclipse resistance.

3. The discovery audit now emits `squatting=<0|1>` whenever a different `agent_id` advertises an honest agent’s capability. `validate_registry_authentic_discovery` rejects that condition. I added a synthetic test where the authentic owner remains present but capability squatting independently fails the validator, and strengthened the `in_memory` end-to-end test to assert the tamper card’s actual payoff.

4. The four registry-integrity validators are now one contiguous block, and the two stray whitespace/separator hunks were removed.

Local CI evidence:
- Ruff check: clean
- Ruff format check: clean
- Pyright: 0 errors
- Pytest: 1179 passed, 1 skipped, 1 deselected
- Focused registry_integrity suite: 13 passed

`make` is unavailable in my Windows shell, so I ran the exact five commands from the `ci-local` target directly and in the same order: `uv sync`, Ruff check, Ruff format check, Pyright, and pytest.

Separately from this one-layer Phase 1 PR, I am also submitting `AuditSkill` as a 2nd Phase audit skill for the hackathon.

Ready for another look — thank you for the precise review.

JamesCarnley:
Thank you, @VladimirPutkov. The core requirement here is already present in rebuilt Town: forged or mutated capability cards are marked unverified, excluded from default discovery, receive no traffic or value, and cannot prevent honest verified discovery and trade. TownIndex verifies signatures at publication, and the native capability_spoofing scenario plus its weak-auth control make the containment claim load-bearing. Typed rejection reasons can still improve future diagnostics, but no replacement is needed for the adopted behavior. I am closing this legacy PR as already covered in current main. No contributor code from this PR was merged or executed during this review.

PR #44, opened 2026-06-30
[Hackathon]NandaQuorum: implement NandaQuorum consensus protocol and scenario testing

saurabhvmagdum:
# feat(consensus): implement NandaQuorum consensus protocol and scenario testing

## Problem
In decentralized environments like Nanda Town, nodes must achieve consensus on state transitions and proposals in a deterministic, safe, and live manner despite potential node crashes or delayed network packets. Prior to this implementation, the codebase lacked:
1. A structured consensus engine capable of coordinating state agreements among multiple agents.
2. A formal protocol for tolerating crash-stop (non-Byzantine) faults up to the mathematical limit ($F = N - Q$ where $Q = \lfloor 2N/3 \rfloor + 1$).
3. Integration hooks in `nest-core` to run consensus simulations and scenarios.
4. Testing infrastructure to validate safety (Agreement, Validity, Integrity) and liveness (Termination) under simulated network degradation (packet drops) or Byzantine environments.

---

## What's in the diff
This diff introduces the entire **NandaQuorum Two-Phase Consensus Protocol** framework. Key additions include:
* **Two-Phase Commit Logic**: Follower nodes (`node.py`) and a leader orchestrator (`leader.py`) that transition through `PROPOSE`, `PREPARE_VOTE`, `QC` (Quorum Certificate), and `COMMIT_VOTE` phases.
* **Network & Metric Simulators**: A simulated peer-to-peer network (`network.py`) with configurable message drops/byzantine behaviors, and a telemetry module (`metrics.py`) to log stats like messages sent/dropped and consensus latency.
* **Scenario Integration**: Registration of `quorum_consensus` plugins and scenarios within the `nest-core` library.
* **Standard Scenarios**: Pre-configured YAML scenarios (`scenarios/`) covering baseline, stress, message-dropping, and Byzantine-presence scenarios.
* **Verification Suites**: A comprehensive unit test suite in `tests/test_consensus.py` (30 test cases) and an interactive command-line simulation script in `scripts/run_quorum_tests.py`.

---

## Summary
The NandaQuorum protocol has been successfully implemented and validated. The engine elects a leader in a round-robin fashion per height and round. Follower nodes vote on proposed values, yielding a Quorum Certificate (QC) once $2/3 + 1$ votes are gathered. Once the leader compiles a second quorum of commit votes, the value is finalized and committed to the state machine. 

To ensure stability, the codebase has been thoroughly verified:
[...]

PR #69, opened 2026-07-05
[Hackathon] brettleehari: fair-ordering a red-team of the coordination layer + the primitive that closes it

brettleehari:
[Hackathon] brettleehari: fair-ordering — a red-team of the coordination layer + the primitive that closes it

## Thesis
In an agent marketplace the **sequencer** sets the execution order of a batch — and *whoever authors that order can extract value from it* (front-running / MEV). Every coordination plugin in the tree assumes an **honest sequencer**: `contract_net` and `hotstuff` (merged), plus the open consensus/auction PRs (sealed-bid #5, PBFT #45, NandaQuorum #44, ResonanceBFT #58). **None addresses ordering fairness** — they answer *who agrees* and *who wins*, never *what if the orderer is the adversary?*

This PR doesn't add another "fair sequencer" (any single design breaks — I broke five). It ships a **systematic red-team of the ordering design space** and the one primitive that survives it, with an integrity oracle keyed on the engine's forge-proof `corr` — never on anything the sequencer wrote.

## The red-team result (six witnesses, one theorem)
| Scheme | Who authors the order | Result |
|---|---|---|
| sort commitments by hash | grinding **trader** (nonce grind) | pos 0 in tens of hashes, 100% — degenerates to a priority-gas-auction |
| beacon over **public** commits | **last-mover** trader | pos 0 in ~N hashes |
| beacon over **private** commits | **omniscient sequencer** (holds all commits) | pos 0 in ~N hashes |
| beacon + sequencer-commits-first | sequencer via **selective inclusion** | pos 0, 28/30 seeds |
| FIFO by **sequencer-polled** commits | sequencer via **emission order** | reverses arrival for free (on the real engine) |
| **FIFO by self-broadcast commits** | **the engine** (`corr` id) | **survives** — arrival order holds regardless of sequencer behavior |

**Theorem (empirically witnessed): whoever authors the order exploits it.** Every *sequencer-authored* scheme leaks. The escape is the last row — author the order with something the sequencer cannot see or forge.

> **Evidence scope (honest):** rows 1–4 are offline red-team scripts (pure-Python, stdlib-only, deterministic), reproducible on request and detailed in the appendix — not committed, per charter (no offline deps in the diff). Rows 5–6 are proven in-repo by the committed scenario + tests (`test_neutral_order_is_engine_authored_not_sequencer`) — so the survivor *and* the break of sequencer-polled FIFO are reproducible here.

## The survivor: engine-authored FIFO
Tier-1 stamps every broadcast with a monotonic, engine-assigned **`corr` id** the sequencer cannot forge (the trace has no `seq` field and all `ts` are `0.0`, so `corr` is the *only* neutral order signal). Traders **self-broadcast** in `on_start`, so arrival order is authored by `corr`, not the sequencer — confirmed on the real engine: a predatory sequencer emitting in reverse **can** reorder under polling but **cannot** under self-broadcast, which is why the scenario uses it. The integrity validator reconstructs arrival order from the submit `corr`s and asserts the executed order matches.

## What's in the PR
- **`coordination/fifo_fair.py`** — `FifoFairCoordination` (survivor: executes in arrival order) + `PredatoryCoordination` (matched attacker: reorders by descending price).
- **`scenarios_builtin/fair_ordering.py` + `scenarios/fair_ordering.yaml`** — 1 sequencer + 8 traders; traders self-broadcast, sequencer finalizes.
- **Two `corr`-keyed validators** — `fair_ordering_integrity` (executed order == engine arrival order) and `fair_ordering_no_injection` (every submit executed exactly once; no censorship, no phantoms).
- **Tests** — plugin units + scenario/discrimination/determinism/robustness.

## The discrimination (the teeth)
Same scenario, swap one line — this ranks three *plausible* sequencers, not just the absent default:

| `coordination:` | `fair_ordering_integrity` | `fair_ordering_no_injection` |
|---|---|---|
| `fifo_fair` | **PASS** | **PASS** |
| `predatory` | **FAIL** (reordered) | PASS (reorders, doesn't drop) |
| `contract_net` | **FAIL** (no `order:*`) | **FAIL** |

## Robustness (I red-teamed my own validator)
The verdict rests on two keys, so both are validated as forge-resistant, not trusted:
- **`corr` (neutral order):** missing / malformed / **duplicate** `corr` on a submit FAILS loudly — never silently bucketed — so a mangled `corr` can't collapse the order and slip a reorder past.
- **`pos` (execution order):** must be numeric, unique, and a clean `0..n-1` bijection; gaps / out-of-range / duplicates FAIL. `pos` is the sequencer's *claim*, only ever checked *against* the `corr`-derived truth.

Guard tests: `test_integrity_fails_on_malformed_corr`, `test_integrity_fails_on_duplicate_pos`, `test_integrity_fails_on_non_bijection_pos`.

## Scope, stated honestly
In zero-latency Tier-1, arrival order collapses to a fixed registration order, so this delivers **unmanipulability (anti-MEV) — no party can *alter* the order** — not egalitarian fairness. Real-latency first-come-first-served is the deployment reading; the fixed order is the Tier-1 model of it. Timestamp trust is engine-provided here; a real network needs trusted receipt timestamps — named, not assumed. The complete answer against a self-interested sequencer is an **encrypted order flow** (Shutter / Aequitas) — a privacy × coordination cross-layer step this PR points at but does not claim.

## How to verify
```bash
uv run nest run scenarios/fair_ordering.yaml -o traces/fair.jsonl
uv run python -c "
from pathlib import Path
for r in validate_trace(Path('traces/fair.jsonl'), 'fair_ordering'):
    print(('PASS' if r.passed else 'FAIL'), r.name, '—', r.detail)
"
# edit fair_ordering.yaml: coordination: fifo_fair -> predatory (or contract_net), re-run, watch integrity flip.
```
Determinism: same seed → byte-identical trace (`test_determinism_byte_identical`). `make ci-local` green: ruff check, ruff format --check, pyright (0 errors, strict), pytest (all pass).

**Persona:** `market-microstructure / adversary engineer` — leads with the threat model, ships the attacker beside the defender, and the validator's whole posture is "trust the engine's record, never the sequencer's narration."

## Appendix: reproduce the red-team (rows 1–4, offline)
Pure-Python, stdlib-only, deterministic (fixed seeds). Not committed per charter (no offline deps in the diff); available on request. Each reproduces one witness:
- **row 1 — sort-by-hash grind:** a trader brute-forces a nonce whose hash sorts first. Reaches position 0 in tens of `sha256` calls, 100% of runs; with competing grinders the biggest CPU budget wins (a priority-gas-auction).
- **rows 2–3 — beacon grind:** `beacon = sha256(sorted(all commit hashes))`, `position = sha256(H_i‖beacon)`. A *public*-commit last-mover, and a *private*-commit **omniscient sequencer**, each randomize their slot per attempt → ~N tries to land position 0.
- **row 4 — selective inclusion:** a commit-first sequencer drops ≤3 commits to reshuffle the beacon in its favour → position 0 in 28/30 seeds.
- **rows 5–6 (in-repo):** `test_neutral_order_is_engine_authored_not_sequencer` proves a predatory sequencer reorders arrival under *polling* but **cannot** under *self-broadcast* — the survivor, on the real engine.


Skyrider3:
Reviewed in depth — this is a strong submission and it is mergeable. CI is green on your head commit (lint/typecheck/test), the local five-command suite is green (962 passed), the diff is purely additive, seed-42 traces are byte-identical, and your discrimination table reproduces exactly as written (fifo_fair PASS/PASS, predatory FAILs integrity with "8 positions reordered", contract_net FAILs both, fail-closed rather than vacuous). I also traced the central claim end-to-end: corr really is engine-assigned by a monotonic counter at broadcast time (simulator.py), trader self-broadcasts pin arrival order, and a forged late submit can't displace a real one. Nice work — including the honest scoping of what Tier-1 can and can't model.

Before merge, one round of small fixes in the validators. The theme: your PR body says "trust the engine's record, never the sequencer's narration" — that's true for the arrival side (corr), but the execution side is currently 100% narration.

1) Bind identities to the engine record, not the msg string. _parse_order_events takes agent= from inside the message body and never reads the trace's own ev["agent"] (the engine-stamped broadcaster). Consequences when your validators run against other traces (which is exactly what the leaderboard does with adversarial validators): any agent can forge order:submit:agent=X / order:execute:agent=X lines attributing activity to others, so phantom orders and censorship laundering pass no_injection; and a sequencer that front-runs internally but narrates its executes in arrival order passes integrity — your PredatoryCoordination is caught because it honestly narrates its reordered batch. Cheap fix that closes most of this: require each submit's ev["agent"] to be a distinct trader matching its agent= token, and every execute's ev["agent"] to be the sequencer; FAIL on mismatch. The residual "the broadcast is the execution" is a legitimate Tier-1 model limit — state it as such.
2) no_injection: compare multisets, not sets. The dropped/injected diff uses set(), so a count-preserving drop-one/phantom-one pair slips through. collections.Counter fixes it in two lines.
3) Single-batch scope. All order:* events are pooled globally; a second legitimate batch emitting its own pos 0..n-1 would spuriously fail the bijection guard. Either group by round or document the single-batch scope in the validator docstrings.
4) Add the evasion tests. Your guard tests (malformed corr, duplicate/non-bijective pos) are real, but there's no test for a relabeled-narration front-run or a spoofed agent= token — the two cases in item 1. test_neutral_order_is_engine_authored_not_sequencer proves the submit side only; until the execute side has an adversarial test, the "catches front-running" claim rests on the cooperative attacker.

brettleehari:
Hi @Skyrider3 
Thanks for catching this. Looking into it and while working on it I uncovered a second issue as well and fixed it.

* **Identity binding:** Agreed. I also found that the parser allowed a `corr=` value in the message to overwrite the engine-generated correlation ID, which meant arrival order could be forged. The parser now always lets engine-generated fields (`corr`, `ts`, and `broadcaster`) win, and `_bind_order_identities()` enforces that submits come from the named trader, executes come from a single non-submitting broadcaster, and every executed trader previously submitted.

* **Injection detection:** The validation logic was already correct; only the diagnostics were misleading because they used `set()`. I've switched the reporting to `Counter`, so duplicate mismatches are now reported correctly.

* **Multiple rounds:** Documented as a current limitation rather than changing the trace format.

* **Tests:** Added regression tests covering the reported issues and the new parser bug.

Seed-42 traces remain unchanged, the discrimination results are unchanged, and the test suite now passes 968 tests (up from 962).

brettleehari:
Hi @Skyrider3 , @shilpa-kulkarni-14 
I am hoping that you would take a look @KavyaSriChennoju @Skyrider3 
It has passed all CI tests. I understand it there are pending comments or CI failure as reasons to not merge. 
Having completed it a week ago, I expect to get this 20% to get on with the competetion.

brettleehari:
Hi @JamesCarnley 
Can you please provide your comment on this PR ?

JamesCarnley:
Thank you, @brettleehari. The reusable future requirement is engine-attributed ordering: for a declared batch, actual arrival order must determine inclusion and order, duplicates must be counted as a multiset, and participant-injected markers, censorship, replay, or cross-batch leakage must not satisfy the evaluator. Current ContractNet has no mempool, proposer, transaction-batch, or FIFO promise. Imposing FIFO would introduce a new fairness/business rule rather than repair the current task-allocation contract, so I am closing the legacy PR and retaining its provenance, multiset, and multi-batch cases for a future ordering profile. No contributor code was merged or executed during this review.

PR #70, opened 2026-07-05
[Hackathon] fayner-delegatable-auth: macaroon capability tokens with cascading revocation (problem 04)

faynercosta:
# [Hackathon] fayner-delegatable-auth: macaroon capability tokens with cascading revocation

**Problem:** #04 — Delegatable capability tokens with cascading revocation (auth layer).
**Persona:** authorization-infrastructure engineer. The risk model, the test
emphasis (adversarial subset/revocation/audience invariants), and the
macaroon idioms below are the submission, not a label on a generic plugin.

> **Naming note (post-#138):** while this PR was in review, #138 merged an
> independent problem-04 solution under `auth/delegatable.py`,
> `DelegatableAuth`, `("auth", "delegatable")` and the `delegated_auth`
> scenario. This submission keeps its implementation unchanged and takes a
> distinct, honest identity so both plugins coexist: **`MacaroonAuth`**,
> registered as **`("auth", "macaroon")`** (the mechanism *is* a macaroon;
> the name says so), with the **`capability_delegation`** scenario — matching
> the problem brief's own slug, `04-auth-capability-delegation`.

## What sets this apart from the other problem-04 submissions

Most delegatable-token PRs (including strong ones) stop at "the HMAC chain
verifies." This one is built around the observation that **the chain proves
provenance, not permission** — any holder can extend the chain with whatever
link they like and compute a perfectly valid signature. Concretely:

1. **Verify-side semantic re-walk.** `verify` independently re-checks scope
   subsets, TTL nesting, and delegator→audience linkage across *every* link,
   so a forged-but-chain-valid broadening link is rejected
   (`test_scope_escalation_rejected_at_verify_when_chain_tampered` exercises
   exactly this, with a hand-forged HMAC-valid token).
2. **Both directions of the adversarial bar are demonstrated, not asserted.**
   Every attack probe runs against *both* plugins through small adapters
   (`macaroon_delegate`/`presenter_verify` vs `naive_jwt_delegate`/
   `blind_verify`): the probes pass against `macaroon` **and** fail against
   the reference `jwt`, in the same test run.
3. **Offline trace-replay validator.** The scenario's gatekeeper emits a
   one-line audit protocol; `validate_capability_delegation_trace` replays it
   offline and independently decides pass/fail — it FAILS on the `jwt` arm
   and PASSES on the `macaroon` arm.
4. **Four Hypothesis properties** (subset invariant at every depth, monotone
   revocation cascade, deterministic token construction, escalation always
   rejected) on top of the unit and end-to-end suites.
5. **Determinism in both arms, asserted.** Byte-identical traces are asserted
   for the `macaroon` *and* the `jwt` arm — the reference plugin is pinned to
   the logical clock at provisioning (`clock=0.0`), so no wall-clock leaks
   into any trace this scenario writes.

## Motivation

The default `jwt` plugin can mint and revoke flat tokens, but it cannot model
the single most common multi-agent authorization pattern: an orchestrator
holding a long-lived root capability, minting narrowly-scoped, short-lived
sub-capabilities for worker agents **without going back to the issuer**, and
having a single `revoke(parent)` cascade to every descendant. `JwtAuth._revoked`
is a set of exact token strings with no parent-child relationship, so revoking a
parent leaves its delegated children valid — the exact gap the problem calls
out, and the gap that makes "an LLM agent sub-renting a tool and cleanly
withdrawing it" impossible to express today.

## Design

`MacaroonAuth` is a macaroon (Birgisson et al., NDSS 2014): a token is a
chain of links, each an attenuating caveat, secured by an HMAC chain
`sig_i = HMAC(sig_{i-1}, link_i)` anchored at the verifier's root secret. Three
properties follow:

- **Offline attenuation.** Any holder can mint a narrower child (`attenuate`,
  or the verifier-checked `delegate`) because extending the chain only needs
  the parent's signature, not the root secret. This is real delegation, not
  central re-issuance (the named anti-pattern).
- **Cascading revocation by construction.** `revoke` records the hash of a
  chain *prefix*; every descendant embeds that prefix, so all of them fail
  `verify` from the next call on — no per-child revocation list.
- **Defense in depth at verify.** The HMAC chain alone cannot stop a malicious
  holder from *writing* broader scopes or a longer TTL into a link (they can
  compute a valid signature for anything they append). So `verify`
  independently re-walks the chain and rejects any link that escalates scope,
  outlives its parent, breaks the delegator→audience linkage, or is expired
  against the injected logical clock.

**Attenuation semantics** (post-review precision): a child's scopes must be a
subset-or-equal (`<=`) of its parent's — equality is deliberately legal, since
re-binding the same scopes to a new audience or shorter TTL is still
attenuation. Delegation TTLs are anchored at the plugin's logical *now*
(`ttl=600` means "600 units from now", capped by the parent's expiry): a `ttl`
the parent cannot cover raises `TtlViolationError` at mint, so mid-life
delegation can never produce a dead-on-arrival child. Offline `attenuate`
without a clock anchors at the parent's `iat`, keeping issuer-free attenuation
fully functional.

Typed errors (`ScopeEscalationError`, `RevokedAncestorError`,
`AudienceMismatchError`, `ExpiredTokenError`, `TtlViolationError`,
`InvalidTokenError`) subclass `ValueError` so existing `Auth` callers that
`except ValueError` keep working. `verify(token, presenter=None)` extends the
`Auth` surface without breaking it — presenter-blind callers keep the exact
protocol signature, and the scenario's gatekeeper falls back through
`TypeError` for plugins that don't accept a presenter.

**Determinism:** no wall-clock time anywhere — in *either* arm; the logical
clock is injected via `set_clock(ctx.time)` (and pinned via `clock=0.0` at
provisioning for plugins without `set_clock`, like the reference `jwt`).
Same seed → byte-identical trace, asserted per-arm in a parametrized test.

## Adversarial validators (mandatory)

`validators/capability_delegation_validators.py` ships three attack probes plus
a trace-level replay validator. Each probe runs against *both* plugins via small
adapter callables, so the "adversarial" property is demonstrated, not asserted:

| Attack | `macaroon` | reference `jwt` |
|---|---|---|
| Scope escalation (child claims scope parent lacks) | **denied** | admitted |
| Stale parent (ancestor revoked, child re-presented) | **denied** | admitted |
| Audience confusion (B's token presented by C) | **denied** | admitted |

The `capability_delegation` scenario (1 coordinator + 1 gatekeeper +
3 intermediaries × 4 workers) runs all three live. Result on the shipped trace:

```
auth=macaroon : ok:allow ×12, escalate:deny ×3, stale:deny ×4, confused:deny ×2  → validator PASS
auth=jwt      : ok:allow ×12, escalate:allow ×3, stale:allow ×4, confused:allow ×2 → validator FAIL
```

The validator FAILS against `jwt` and PASSES against `macaroon` — the
charter's bar.

## Tests

- `test_macaroon.py` — 15 unit tests: happy path, each typed rejection,
  two-level cascade, sibling isolation under revocation, tampered-signature and
  hand-forged-broadening rejection, offline-attenuate equivalence, mid-life
  delegation (now-anchored TTL + dead-on-arrival rejection), and the
  cross-plugin probe property.
- `test_macaroon_properties.py` — 4 Hypothesis properties: subset invariant
  holds at every chain depth, revocation cascades monotonically (depth ≥ k
  revoked, depth < k spared), token construction is deterministic, escalation
  is always rejected.
- `test_capability_delegation_scenario.py` — end-to-end: macaroon denies every
  attack, jwt admits them, and the trace is byte-deterministic in **both** arms
  (parametrized over `macaroon` and `jwt`).

23 tests, all green locally under the CI command sequence.

## Runnable verification

```bash
nest run capability_delegation
python -c "from nest_plugins_reference.validators.capability_delegation_validators import validate_capability_delegation_trace as v; \
           [print(('PASS' if r.passed else 'FAIL'), r.detail) for r in v('./traces/capability_delegation.jsonl')]"
```

## Tradeoffs / scope

- Strict tree (single parent per token), per the problem's allowance; DAG
  multi-parent delegation is out of scope.
- Revocation state lives in the verifier (a set of prefix hashes) — O(depth)
  membership checks per verify, no background GC. Fine for simulation; a
  production deployment would bound the set with token expiry.
- No hardware key attestation, no OAuth2 endpoints, no network introspection —
  all explicitly out of scope for the problem.


faynercosta:
## Phase-2 skill <-> Phase-1 PR link (per organizers' request)

**Skill:** `lex-automata` — escrow + instant, deterministic dispute resolution for agents that pay each other; every verdict is a signed, offline-verifiable receipt.

- Registry card (green / "link responded"): https://nandatown.projectnanda.org/skills — "LEX AUTOMATA" by Fayner Costa. Record: https://nandatown.projectnanda.org/api/skills/e96fd7cd-2e45-48c2-8e96-0a0bbe6fb67d
- SKILL.md (as submitted, = the record's source): https://raw.githubusercontent.com/faynercosta/lex-automata/main/SKILL.md
- Live service (self-documenting — the base URL serves that exact SKILL.md; capability metadata at `/agentfacts`): https://lex-automata-999015027200.us-central1.run.app
- Source repo: https://github.com/faynercosta/lex-automata
- Live arena, every animation backed by real calls to the deployed court: https://faynercosta.github.io/lex-automata/arena/

**Which layer each part adds value to:**

- **This PR -> the authorization/trust layer** (Problem #04): delegatable capability tokens with cascading revocation decide **who may act**, and how authority is narrowed and withdrawn across a delegation chain.
- **The skill -> the settlement/trust layer**: Lex Automata decides **who gets paid** when delegated work is disputed - escrow, deterministic replay of machine-checkable acceptance criteria (LLM jury only for semantic criteria), Ed25519-signed verdict receipts, and receipt-derived reputation with bans.

They are the two halves of trust in agent-to-agent commerce: authority to act (this PR) and accountability for results (the skill). Both artifacts are verifiable offline - the HMAC capability chain here, the signed verdict receipt there - so neither requires trusting a central party after the fact.

dhve:
Thanks @faynercosta. This is one of the cleanest problem 04 submissions I have read. The verify-side semantic re-walk is the part most competitors miss: you correctly treat the HMAC chain as proof of provenance only, and independently re-check scope subsets, TTL nesting, and the delegator-to-audience linkage, because any holder can extend the chain with whatever link they like. `test_scope_escalation_rejected_at_verify_when_chain_tampered` makes that distinction load-bearing instead of decorative. Keeping `verify(token, presenter=None)` as an optional kwarg (with the gatekeeper's `TypeError` fallback for presenter-blind plugins) is exactly how to extend the `Auth` surface without breaking the protocol, and the dual-adapter probe design (`naive_jwt_delegate`/`blind_verify` vs `delegatable_delegate`/`presenter_verify`) demonstrates the fails-on-jwt, passes-on-yours bar in both directions rather than asserting it.

This is close. Four things to tighten before merge:

**1. Wall-clock time leaks into the jwt arm of the scenario.**

`_provision_auth` in `scenarios_builtin/delegated_auth.py` instantiates `auth_cls(secret=b"delegated-auth-scenario-root")` with no `clock`. `JwtAuth._now()` falls back to `time.time()` when its clock is `None`, and `JwtAuth` has no `set_clock`, so the `hasattr` guards in your agents never fire for it. Wall-clock `iat`/`exp` values end up inside the tokens, tokens ride in the `present:` payloads, and the simulator records payloads in the trace `msg` field, so the trace written by `test_reference_jwt_admits_attacks` differs across runs. Your delegatable arm is genuinely deterministic and the validator verdicts are stable in both arms, but wall-clock anywhere in a simulated run is a charter red flag. Near one-line fix: try `auth_cls(secret=..., clock=0.0)` first, then fall through `TypeError` as you already do. Both `JwtAuth` and `DelegatableAuth` accept `clock`.

**2. TTL anchored to the parent's `iat` can mint dead-on-arrival children.**

In `attenuate`, `child_exp = float(parent_link["iat"]) + float(ttl)` and the child inherits the parent's `iat`. Delegate at logical time 3000 from a root with `iat=0`, `exp=3600`, and `ttl=600`: `delegate` succeeds (the parent verifies fine) and hands back a token that expired at t=600, unusable from birth, with no error at mint. It also means `AuthContext.issued_at` for every descendant reports the root's issue time. Your scenario never trips this because all delegation happens in the first few ticks, but a judge poking the API mid-run will. Suggested: have `delegate` pass `self._now()` into `attenuate` (optional `now` param defaulting to the parent's `iat`, so offline attenuation keeps working), set the child's `iat` from it, and either cap the expiry at the parent's or raise `TtlViolationError` when the requested ttl cannot be honored. At minimum document the anchoring in `delegate`'s docstring, since the `ttl=600` example reads as ttl-from-now.

**3. Say explicitly what differentiates this from the problem 04 crowd.**

Roughly twenty open PRs are implementing delegatable capability tokens right now, and novelty is scored against peers. Your PR body already describes the strongest parts well (the defense-in-depth verify, the dual-adapter probes, the trace replay validator, the four Hypothesis properties), but it never positions them against the field. Add a short "what is different about this one" paragraph naming them as differentiators: the verify-side re-walk that survives a forged-but-chain-valid link, and probes that run against both plugins instead of only self-testing, are the two most competitors skip. It costs five lines and directly addresses the most crowded dimension.

**4. Doc/code mismatch on "strict subset".**

The `delegatable.py` module docstring says a child's scopes "must be a strict subset" of its parent's, but the check is `set(scopes) <= parent_scopes` (subset-or-equal), and `test_revocation_cascades_monotonically` relies on equal-scope re-delegation being allowed. Subset-or-equal is the right macaroon semantics (re-binding the same scopes to a new audience is still attenuation), so keep the code. But the problem brief uses the same "strict subset" phrasing, so spend one docstring line saying explicitly that you enforce subset-or-equal and why.

Housekeeping: no CI checks have reported on this branch yet (likely awaiting first-contribution workflow approval), and GitHub is reporting the branch as conflicting with main, so it needs a rebase. With this many plugins landing, the shared registry tables (`nest_core/plugins.py` and the pyproject entry points) are the usual collision points. Registering in `nest_core/plugins.py` plus the entry point is exactly what problem 04 asks for, so no scope concerns there.

**Suggested path forward:**

1. Rebase onto current main to clear the conflicts.
2. Thread `clock=0.0` through `_provision_auth`.
3. Decide the ttl anchoring semantics, encode them in `delegate`, and add one test for mid-life delegation.
4. Clarify the subset wording and add the differentiation paragraph to the PR body.
5. Once workflows run, confirm all five CI stages are green; the judge panel does not score broken PRs.

Leaving this open. Ping me on the hackathon channel if you want to talk through the ttl anchoring tradeoff.

faynercosta:
Thank you for this kind and thourough review @dhve.  it's great to get someone that engages with the design at this level, and the four points were all real. Everything is addressed; here's the map, plus one structural change the rebase forced.

Sructural note first: #138 took all the namespace i had chosen. When I merged main, `ckeda-delegatable-auth` (#138) had landed an independent problem-04 solution occupying `auth/delegatable.py`, `DelegatableAuth`, `("auth", "delegatable")`, the `delegated_auth` scenario, and `delegation_validators.py`. Therefore "rebase to clear conflicts" meant every one of my files was an add/add collision. Rather than contest names, this submission now has a distinct, honest identity (f2900e0): **`MacaroonAuth`** registered as **`("auth", "macaroon")`** — the mechanism *is* a macaroon, so the name says so — with the scenario renamed to **`capability_delegation`** (which happens to match the problem brief's own slug, `04-auth-capability-delegation`). Implementation unchanged; both plugins coexist in the registry and both scenarios register cleanly. Happy to pick different names if the maintainers prefer.

**1. Wall-clock in the jwt arm — fixed** (dd5e387). `_provision_auth` now tries `auth_cls(secret=..., clock=0.0)` first and falls through the existing `TypeError` ladder, exactly as you suggested. And so this can't regress: the byte-determinism test is now parametrized over **both** arms (2d594fa), `test_trace_is_byte_deterministic[jwt]` would have caught the original leak, since wall-clock `iat`/`exp` land in the `present:` payloads that reach the trace.

**2. TTL anchoring — fixed, with the "raise" branch of your tradeoff** (dd5e387). `attenuate` takes an optional `now` (default: parent `iat`, so clock-free offline attenuation keeps working) and `delegate` passes `self._now()`. Children are stamped `iat=now`, `exp=now+ttl`, so `AuthContext.issued_at` now tells the truth for every descendant. For the un-honorable case I chose **raise `TtlViolationError` at mint** over capping: capping silently hands back something different from what was asked for, while raising keeps the existing mint-time error contract and makes the failure visible at the right place. Your exact scenario is now a test: delegating at t=3000 from a root with `exp=3600` succeeds with `ttl=600` (child expires exactly at the parent's boundary) and raises with `ttl=601` — no dead-on-arrival children, ever (`test_midlife_delegation_anchors_ttl_at_now`, `test_midlife_delegation_that_would_outlive_parent_raises`). If you feel capping is friendlier for the offline path, I'm open to it — but I'd rather a delegator learn at mint than a worker learn at present.

**3. Differentiation — added.** The PR body now opens with a "what sets this apart" section naming the two things you called out (the verify-side re-walk that survives a forged-but-chain-valid link, and probes that run against both plugins in both directions), plus the trace-replay validator, the Hypothesis properties, and the both-arms determinism assertion.

**4. "Strict subset" wording — fixed** (dd5e387). The module docstring now states the rule as subset-or-equal (`<=`) explicitly and says why equality is correct attenuation semantics: re-binding the same scopes to a new audience or a shorter TTL is still attenuation, and the monotone-revocation property test relies on equal-scope re-delegation being expressible. Code unchanged, as you recommended.

Branch is merged onto current main and mergeable again; full local run of the CI sequence (ruff check + format, pyright, pytest) is green. Thanks again! The review made the plugin measurably better, and the ttl point in particular was a real good catch.

faynercosta:
Re-synced onto current main (b28c9ae) — the new conflicts were one-line registry adjacencies with #155's `mesh_revocable` plugin in `nest_core/plugins.py` and the `nest.plugins.auth` entry points; resolved keeping both. Full local CI on the merged tree: 1305 tests pass, ruff + format clean, pyright 0 errors. Mergeable again.

One heads-up for the maintainers: main is currently red on the lint stage — `test_negotiation_determinism.py` (merged via #184) has an E501 overlong signature. This branch carries the mechanical `ruff format` rewrap (a785463, 3 lines, no behavior change) so the merge CI here is green regardless; feel free to drop that commit if you fix it on main first.

JamesCarnley:
Thank you, @faynercosta. This contribution reinforces a valuable future requirement: the verifier must recompute caveats and reject privilege amplification, wrong audience/resource, expiry, revocation, and wrong-root chains, with a real nondelegating control that fails. Current Town has no capability-token or delegation API, and choosing macaroons over another mechanism is an unratified policy decision. I am closing this as a duplicate in the legacy delegation family and folding its verifier-side negative cases into the consolidated future acceptance profile recorded from #73. No contributor code was merged or executed during this review.

PR #71, opened 2026-07-05
[Hackathon] chainaim-agentID-engineer: Ed25519 identity with KERI pre-rotation and hijack-resistant key rotation

chainaim-bala:
# [Hackathon] chainaim-agentID-engineer: Ed25519 identity with KERI pre-rotation and hijack-resistant key rotation

## Short summary

Problem #05 — identity / Ed25519 key rotation. Adds a KERI-style **pre-rotation** identity plugin at `("identity","ed25519_prerotation")`, building on the merged reactive-rotation plugin (PR #25) to defend a third attack class — rotation hijack — in addition to post-rotation forgery and backdating, with a recovery path. It is additive: `ed25519_rotating` and `did_key` are unchanged, and the merged `identity_rotation` validators run as-is on these traces.

**Persona — chainaim-agentID-engineer.** The design is driven by a live-key-compromise risk model rather than a happy-path feature: the plugin assumes an attacker can exfiltrate the current signing key, and every adversarial test is mapped to a production incident — post-rotation forgery to a stolen retired key (old backup, ex-employee), backdating to an antedated contract or tampered audit timestamp, rotation hijack to a leaked CI secret, and recovery to the incident-response path of rotating to a pre-committed cold key that never touched the compromised host.

## What it does

At inception the plugin derives signing key `K0` and pre-derives cold key `K1`, publishing `digest(K1.pub)` as a commitment. `rotate_key(new_seed)` reveals the pre-committed `K1` and commits `digest(K2)`. `verify_continuity` accepts a rotation only when the revealed key hashes to the prior commitment, so a party holding only the current private key cannot install a successor of its own choosing. `verify` evaluates a signature against a verifier-supplied as-of tick (not `sig.signed_at`), which is what rejects forgery and backdating. `forge_rotation` and `sign_with` are adversarial-only hooks used by the scenario; they mutate no state.

**Visual explanation:** *https://chainaim-keri-prerotation.vercel.app/*

## Run and check

```
uv sync
uv run pytest packages/nest-plugins-reference/tests/test_ed25519_prerotation.py -v            # 28
uv run pytest packages/nest-plugins-reference/tests/test_ed25519_prerotation_properties.py -v  #  8
uv run pytest packages/nest-core/tests/test_identity_prerotation.py -v                         # 23
uv run nest run scenarios/identity_prerotation.yaml --seed 42
```

Full keyword-scoped suite (new work plus the merged identity tests it coexists with):

```
uv run pytest -k "ed25519 or identity_rotation or prerotation" -v
```

## Registration / key surface

|Surface|Where|
|-|-|
|`("identity","ed25519_prerotation")`|`plugins.py` `_BUILTINS` **and** `pyproject.toml` `nest.plugins.identity` entry point|
|`sign` / `verify(..., as_of)` / `resolve` / `rotate_key(new_seed) -> KeyId`|plugin — protocol surface + spec-exact return|
|`register_peer` / `register_peer_inception`|legacy path + pre-rotation inception (key + commitment)|
|`verify_continuity` / `apply_rotation`|rotation = signature + chain-tip + commitment recompute|
|`digest_alg="sha256"`|constructor flag; unknown algorithms rejected at construction|

Every module carries the Nanda Town idioms: SPDX header, `from __future__ import annotations`, and an `Example::` block on each public class and function. The plugin uses `nest_core` types (`AgentId`, `AgentIdentity`, `Signature`) and implements the identity layer's `sign` / `verify` / `resolve` surface.

## Threat model — 3 attacks + recovery

|#|Attack|Defeated by|Caught by|
|-|-|-|-|
|1|Post-rotation forgery (stolen retired key signs)|closed as-of window|`verify`; validator (a)/(b)|
|2|Backdating (new key claims old-window tick)|as-of window, verifier-supplied tick|`verify` + `_select_record`; validator (a)/(b)|
|3|Rotation hijack (current key installs chosen successor)|revealed key must hash to prior commitment|`verify_continuity` / `_commitment_matches`; validator (c)/(d)|
|—|Recovery|rotate to pre-committed cold key, sign validly|validator (e)|

## How this PR relates to the merged rotation work

Problem #05 covered the identity layer, where the merged reactive-rotation plugin (PR #25) is the existing baseline. This PR extends that baseline rather than replacing it: it keeps the same protocol surface and as-of window semantics, and adds a pre-rotation commitment step that lets one further attack — rotation hijack — be rejected and recovered from. The two plugins are complementary; the comparison below is between reactive rotation as a design and pre-rotation as a design, not a criticism of the merged work.

|Axis|Reactive rotation (`ed25519_rotating`, PR #25)|Pre-rotation (`ed25519_prerotation`, this PR)|
|-|-|-|
|Attacks addressed|post-rotation forgery, backdating|the same two, plus rotation hijack, plus recovery|
|Rotation authority|old key signs the new key (continuity)|old key signs the new key **and** the revealed key must match a previously published commitment|
|Hijack case|a valid continuity signature over any successor is sufficient, so a current-key holder can install its own successor|the revealed successor must match the prior commitment, which a current-key holder cannot reproduce|
|Key material|one live key per rotation|live key `K0` plus a pre-derived cold key `K1`, committed before first use|
|Recovery|not modeled|victim rotates to the pre-committed cold key and signs validly (validator check (e))|
|Trace|`rotate:` / `signed:`|adds `commit:` and `rotate_attempt:...:hijack`; keeps `rotate:` / `signed:` byte-compatible|
|Validator|window checks|window checks re-expressed, plus commitment-chain, hijack, and recovery checks, all trace-only|
|Coexistence|—|additive: `ed25519_rotating` is untouched and its validators pass unchanged on these traces|

In short, reactive rotation authorizes a rotation by asking whether the current key signed it; pre-rotation additionally asks whether the revealed key matches a commitment fixed at the previous event. `TestDiscriminationMatrix` records that the same scenario passes under `ed25519_prerotation`, and fails without error under `ed25519_rotating` (which rotates but publishes no commitments) and `did_key` (which does not rotate).

## Validator

The `identity_prerotation` scenario dispatches two validators: the merged `validate_identity_rotation_occurred` (reused unchanged, confirms a rotation happened) and the new `validate_identity_prerotation`. The new validator reconstructs each key's window and the commitment chain **from the trace only** (never from plugin state) and enforces: (a) honest `ok` signatures are window-valid at both observed and claimed tick; (b) `forge`/`backdate` signatures are window-invalid; (c) every applied rotation matches its predecessor's commitment (strict); (d) every `hijack` attempt's digest mismatches the commitment and never becomes valid; (e) recovery occurs after every hijack. Its window semantics are re-expressed locally rather than imported from the merged validator's private helpers, so the two spec-mandated checks are visible in this diff.

## Trace grammar

Additive over the merged `identity_rotation` grammar — `rotate:` / `signed:` keep their shapes:

```
rotate:<agent>:<old_key_id>:<new_key_id>:<tick>
signed:<agent>:<key_id>:<claimed_tick>:<ok|forge|backdate>
commit:<agent>:<key_id>:<alg>:<hex>:<tick>                                    # new
rotate_attempt:<agent>:<old_key_id>:<alg>:<hex>:<alg>:<hex>:<tick>:hijack     # new (rejected)
```

`key_id` is the SHA-256 hexdigest of the public key, so a `sha256` commitment is recomputed directly against `new_key_id`; other algorithms are reported unverifiable.

## Meets Problem #05 success criteria

Real Ed25519 via `cryptography` (not simulation); `rotate_key(new_seed) -> KeyId` exact; old key verifies only as-of its window and post-rotation forgery fails; each `Signature` carries `key_id`; the scenario's validator set catches the two mandated attacks (forgery + backdating), passes on this plugin, and fails on `did_key` (which emits no rotation). The scenario ships ≥1 rotation per agent with 10% byzantine agents running both attacks. Anti-patterns avoided: real Ed25519 (no simulation RSA), rotation is not a no-op, verification is as-of rather than current-key-only, and private keys never appear in the trace.

## Files (identity layer only)

* `nest-plugins-reference/.../identity/ed25519_prerotation.py` — plugin
* `nest-core/.../validators.py` — `identity_prerotation` validator dispatch **and** the pre-rotation validator itself (`validate_identity_prerotation`, 5-check, trace-only), inlined alongside the other layer validators
* `nest-core/.../scenarios_builtin/identity_prerotation.py` — scenario driver
* `scenarios/identity_prerotation.yaml` — 11 agents (9 honest + 1 byzantine + 1 auditor), tier 1, seed 42
* `nest-core/.../plugins.py` — `_BUILTINS` registry entry
* `nest-core/.../scenarios.py` — `identity_prerotation` factory branch
* `nest-plugins-reference/pyproject.toml` — entry point + `cryptography>=42.0`
* `nest-core/tests/test_validators.py` — validator cases (registry + dispatch)
* tests: `nest-core/tests/test_identity_prerotation.py`; `nest-plugins-reference/tests/test_ed25519_prerotation.py`, `test_ed25519_prerotation_properties.py`

All pre-rotation code lives at its package root (no `chainaim/` subpackage): the plugin sits beside `did_key.py` / `ed25519_rotating.py`, the scenario driver beside the other `scenarios_builtin` drivers, the validator inline in `validators.py`, and the tests beside their sibling suites.

## Verify (fresh clone)

Cloning this branch fresh and running the keyword-scoped suite proves the PR is self-contained:

```
uv sync
uv run pytest -k "ed25519 or identity_rotation or prerotation" -v
```

Observed on a fresh clone + venv: **105 passed, 692 deselected, 0 failed**. Of these, **59 are new in this PR**: 28 plugin unit tests, 8 Hypothesis **property-based** tests (window correctness, forgery/backdating always rejected, commitment-chain integrity), and 23 validator/E2E tests including adversarial byzantine traces and a determinism (byte-identical) check. The other 46 are the merged PR #25 identity tests the keyword also selects, confirming the new work runs additively without disturbing them.

**Visual explanation:** *https://chainaim-keri-prerotation.vercel.app/*

## References

Design provenance: Samuel M. Smith, *Key Event Receipt Infrastructure (KERI)*, arXiv:1907.02143. No KERI implementation code is used or depended upon. Ed25519 per RFC 8032 (deterministic). Extends the merged reactive-rotation plugin (PR #25).


dhve:
Thanks @chainaim-bala. This is one of the most carefully argued identity submissions in the queue. `Ed25519PreRotatingIdentity` gets the commit-then-reveal discipline right (`rotate_key` reveals the key committed at the previous establishment event, and `new_seed` only influences the key after it, which `test_new_seed_cannot_choose_revealed_key` pins down exactly), the `TestDiscriminationMatrix` pass/fail matrix against `ed25519_rotating` and `did_key` is precisely the fails-on-baseline evidence we ask for, and returning a `KeyId` str subclass that stays spec-exact while still carrying the rotation evidence the merged scenario reads is an elegant solve for drop-in compatibility. The threat-model table and the explicit differentiation against merged #25 in the PR body is how a crowded-layer submission should be argued.

This is close. Three things to work through before it's scoreable:

**1. The branch conflicts with main and CI has never run.**

GitHub reports the PR as CONFLICTING. Your base predates recent merges that shifted `validators.py` substantially (your hunks anchor `validate_identity_rotation_signatures` near line 1062 and the `VALIDATORS` registry near 4598; on current main they sit near 981 and 4743), so at minimum that file needs a manual rebase. Your additions are self-contained blocks, so it should be mechanical: rebase onto main, re-run the keyword-scoped suite, push. No checks have been reported on this branch at all, and the judge panel does not score broken PRs, so a visible green lint/typecheck/test run is the gate for everything else. While you're in there: drop the stray blank-line deletion above the `VALIDATORS` registry (unrelated whitespace churn in the most conflict-prone file in the repo), and add a blank line between your new `[project.entry-points."nest.plugins.identity"]` table and the `failure_detector` table in `pyproject.toml`.

**2. `validate_identity_prerotation` never checks that the commitment came before the rotation.**

In pass 1 you parse the `commit:` tick (`digest_tick = _tick(parts[5])`) but use it only as a malformed-line guard, and `commitments` is a last-write-wins dict consulted only after the whole trace is ingested. So check (c) accepts a commitment that appears after the rotation it supposedly binds. Concretely: a cheating plugin that applies a hijack can emit `rotate:s0:K0:KX:2.0` followed by `commit:s0:K0:sha256:<KX-digest>:5.0` and the validator passes, because (c) finds a matching digest and (d)/(e) are vacuous when the attacker simply never emits a `rotate_attempt:...:hijack` line. Temporal ordering is the entire "pre" in pre-rotation, and the docstring promises checks "replayed purely from the trace", so this gap undercuts the headline property of the mandatory adversarial validator (correctness). The fix is small: use the commit tick (or trace position) and require the commitment to be observed strictly before the rotation tick in check (c), and make commitment recording first-write-wins per key id (or flag duplicates) so a late attacker commit cannot overwrite an honest one. Your honest traces already satisfy the stricter rule (inception commit at tick 0, first rotation at tick 2, recovery later), so nothing else should move. Then add two failing-direction cases to `test_identity_prerotation.py`: commit-after-rotate fails, and duplicate-commit-overwrite fails.

**3. `apply_rotation` is not idempotent on replay.**

`verify_continuity` deliberately accepts an already-applied rotation (the `already_applied` reasoning is nice), but `apply_rotation` then unconditionally appends a second `KeyRecord` for the same key. A peer that receives the same gossiped `RotationRecord` twice grows duplicate history, which `resolve()` faithfully exports in `metadata["keys"]`. Early-return `True` without appending when `rotation.new_key_id` is already in the chain, and add the replay test.

Two doc nits: all three test module docstrings still point at `tests/chainaim/...` paths in their `Example::` blocks (stale after the flattening you describe in the PR body), and the file list in the body names `cryptography>=42.0` as part of the pyproject change, but the diff only adds the entry point (the dependency is already on main), worth correcting so the no-new-deps story reads unambiguously.

**Suggested path forward:**

1. Rebase onto current main and resolve `validators.py` (your other hunks apply cleanly against today's main, so the conflict should be contained to that file).
2. Tighten check (c) with commit-before-rotate ordering plus first-wins commitments, and add the two failing-direction tests.
3. Make `apply_rotation` replay-safe.
4. Fix the stale docstring paths and the PR-body dependency line.
5. Run `uv sync`, `ruff check`, `ruff format --check`, `pyright`, and `pytest -v` locally, then confirm CI reports green on the PR.

On novelty: merged #25 is the baseline you build on and #117 is also in the identity lane, but pre-rotation plus hijack rejection plus recovery is a distinct mechanism, and your discrimination matrix makes the differentiator load-bearing in tests, which is exactly right. Leaving this open. Ping me on the hackathon channel if the validator ordering fix is unclear.

JamesCarnley:
Thank you, @chainaim-bala. A future rotation profile should prove that a committed next-key hash precedes rotation, replaying the same transition is idempotent, stale old-key signatures fail, and an uncommitted replacement cannot take control while the committed next key succeeds. Current Town's Lab identity is fixed per run and its portable identity path has no selected rotation lifecycle. Adding pre-rotation, persistence, recovery, and authority rules would be new identity policy, not a current regression. I am closing the legacy PR while retaining those ordering, replay, stale-key, and uncommitted-key cases for a future identity-rotation profile. No contributor code was merged or cryptographically executed during this review.

PR #59, opened 2026-07-03
[Hackathon] interview-integrity: Problem 08 (datafacts) — ogha_facts plugin: audience ACLs + pre-hash PII redaction + screening-evaluation scenario & validators

puja-ankitha-ivaturi-ogha:
## Problem 08 — datafacts (persona: interview-integrity engineer)

I run the AI services for a **verified-screening** business — the filter at the *front* of a client's hiring funnel. A human interviewer conducts a recorded screening interview; the audio is transcribed, and the interviewer files an evaluation — a `PASSED`/`FAILED` verdict, their gist notes, and supporting **Q&A pulled verbatim from the transcript** so the client sees exactly what was asked and answered — that we deliver to the company that posted the job. We only screen; candidates who pass go on to the client's own rounds (the hiring decision is theirs, not ours). Our value proposition is that the report is *genuine and grounded*: the right candidate, from the assigned interviewer, its evidence **provably drawn from a real interview** — not swapped, fabricated, leaked, or stale. This PR models that pipeline on-protocol as the chain **recording → transcript → report** and hardens every hop.

### What this adds over the merged `cid_facts`

`cid_facts` (merged in #31) already solves the three properties Problem 08 asks for — content-addressed URLs, a `parents` provenance DAG, and identity-signed freshness proofs. **Re-implementing those would be duplicate work**, so I don't. `ogha_facts` **subclasses** `CidFacts` and adds the two capabilities the `datafacts.md` wishlist explicitly lists as unclaimed (*"fine-grained ACLs"*, redaction), which an evaluation-delivery pipeline genuinely needs:

1. **Audience-scoped ACLs.** `CidFacts.request_access` is binary — owner-or-public gets `read`, everyone else is hard-denied with `PermissionError`. That's wrong for a verdict: it's private, but has *several* legitimate readers (the interviewer, the service, the posting company, the candidate), while every *other* company must still be able to audit that a record exists without reading it. `ogha_facts.request_access` returns `read` to the content-bound ACL audience and a **`metadata`-tier grant (never a hard denial)** to everyone else — fail-*closed* on content, fail-*open* on existence. The ACL is part of the hashed content, so *who may read a verdict cannot be changed without minting a new URL*.

[...]

PR #34, opened 2026-06-22
feat(datafacts): implement content-addressed plugin

24f3003188:
## Overview
This PR implements Problem 08: Content-Addressed Datasets with Provenance Chains and Freshness Proofs for the NANDA Town hackathon.

It introduces a new `content_addressed` DataFacts plugin that solves the vulnerabilities present in the reference `datafacts_v1` implementation by utilizing content-based hashing and cryptographic signatures.

## Key Features & Changes
- **Content Addressing**: Replaced vulnerable name-based URLs with `sha256` content hashes covering both metadata and payload bytes. This guarantees immutability and makes data substitution attacks impossible.
- **Cryptographic Freshness Proofs**: Replaced simplistic wall-clock (`time.time()`) freshness checks with cryptographically signed `FreshnessProof` objects using the identity layer, ensuring that freshness claims cannot be spoofed by third parties.
- **Provenance Chains**: Added support for tracking dataset ancestry via the `parents` metadata field. Implemented `verify_provenance_chain` to validate that a dataset's entire lineage forms an intact, cryptographically verifiable DAG.
- **Adversarial Validators**: Added a new suite of adversarial validators (`datafacts_validators.py`) that successfully catch substitution, stale freshness, and broken provenance attacks (failing against `datafacts_v1` and passing against the new `content_addressed` plugin).
- **Scenario & Tests**: Created the `provenance_supply_chain.yaml` scenario to demonstrate end-to-end functionality, alongside a comprehensive `pytest` suite ensuring 100% protocol conformance and determinism.

All local tests, type checks (pyright), and linters (ruff) pass successfully.

PR #72, opened 2026-07-06
[Hackathon] maurathat: sic_facts — serialization-invariant content addressing (closes the provenance-laundering gap in byte-level CIDs)

maurathat:
**TL;DR:** `cid_facts` prevents name substitution, but not identity forking under semantically equivalent serialization. `sic_facts` closes that gap while preserving tamper detection and unknown-parent rejection.

## Problem 08 follow-on: serialization-invariant datafacts addressing

PR #35, #34, and #59 closed the *name*-substitution attack by making datafact URLs content-addressed. This PR closes the remaining attack that byte-level content addressing still permits: **encoding substitution / provenance laundering**.

Today, `cid_facts` binds a dataset to a byte digest derived from the payload (`checksum`) plus `json.dumps` of the metadata dict. That prevents simple name substitution, but a semantically equivalent re-encoding still receives a fresh address:

- integer → float normalization (`1` re-exported as `1.0`)
- JSON key reordering
- NFC → NFD Unicode normalization
- re-gzipped exports (new bytes, same parsed structure)
- other encoding-level changes covered by the canonicalizer

None of these alter the dataset's structure, but each mints a new `df://sha256-*` URL. As a result, quarantine flags, annotations, ACL references, and provenance parents can be silently severed from the original dataset identity.

This PR adds `sic_facts`, which addresses datasets by a canonical **structural** digest instead of raw bytes. Semantically equivalent re-encodings covered by the canonicalizer collapse to the same URL; real tampering still changes the address; and provenance parents cited through a re-encoded alias resolve back to the original parent.

## New scenario

Adds `serialization_invariance` plus three adversarial validators:

- `sic_launder_neutralized`
- `sic_tamper_still_detected`
- `sic_phantom_parent_rejected`

With `layers.datafacts: sic_facts`, all validators pass. Flip one YAML line back to the currently merged `cid_facts` plugin and `sic_launder_neutralized` fails with forked addresses in the trace — a concrete gap in main, not a strawman.

## Receipts (reproduced on latest main @ 9a4e6fb, 2026-07-05 PT)

```
nest run scenarios/serialization_invariance.yaml
```

`layers.datafacts: sic_facts` (as committed):

```
PASS sic_launder_neutralized - re-encoded republish collapsed onto df://sic-9c03867928c831ad327ac8c11a0cf979454a7a52314a694d9273dc97a3881b3c
PASS sic_tamper_still_detected - tamper produced distinct address df://sic-4e7cb2b4f849f1ccbab2d24f4fa9eb6ebfb5e0acb7441f7e9baf8b59e6444f36
PASS sic_phantom_parent_rejected - unpublished parent rejected
```

Same scenario with one line flipped to `layers.datafacts: cid_facts`:

```
FAIL sic_launder_neutralized - re-encoded copy minted a fresh address: ['df://sha256-e6ebb4ffc8c8b5f4339c045148da0dc637fb1f51a4a6a4bb0115521b7ab96709', 'df://sha256-fe08b85307842ed986c9f2cd003c1f40c8110668ac7d83166b377cbd93d2eeb7'] (history severed)
PASS sic_tamper_still_detected - tamper produced distinct address df://sha256-33c6d2b6655a746696b7d5c5bb1acdbd79fb16d8a11f683e3f79c5f25bb6f70e
PASS sic_phantom_parent_rejected - unpublished parent rejected
```

## Test coverage

13 new tests, including an explicit contrast pair against `cid_facts`. All Problem 08 base criteria remain intact:

- content-addressed URLs
- provenance DAG with unknown-parent rejection
- identity-signed freshness proofs (logical ticks, never wall-clock)
- hash-keyed ACLs

## Local CI

```
make ci-local: green end-to-end
pytest: 749 passed / 0 failed
ruff check: clean
ruff format --check: clean
pyright: 0 errors
```

Deterministic throughout: logical ticks, seeded scenario, no wall-clock dependence, stdlib-only canonicalization.

## Background

The canonicalization approach follows my work on serialization-invariant content addressing for ML model artifacts (UOR-ADDR, UOR Foundation), where byte divergence across encoders for logically identical artifacts is the central failure mode.

---

This preserves all Problem-08 guarantees while adding serialization-invariant identity for canonicalized dataset structures.


maurathat:
Merged `upstream/main` (the sybil_bond validators landed at the same insertion point in `validators.py`). The resolution preserves both validator sets verbatim — `git diff upstream/main -- validators.py` is +114/−0, additions only, so this PR's effective diff is unchanged. `make ci-local` green at `249b65a`: 1018 passed, including `test_all_scenario_types_registered` with both registries present.

dhve:
Thanks @maurathat. This is one of the sharpest follow-ons in the datafacts lane: the injective, type-tagged canonical encoding in `_canon` (length prefixes, `True != 1 != "1"`, NFC key normalization, int-valued float collapse) is the right way to build a structural digest, and the flip-one-line contrast in `scenarios/serialization_invariance.yaml` plus the `test_laundering_attack_succeeds_against_cid_facts` pair demonstrates a real gap in the merged plugin rather than a strawman. The PR body is a model writeup.

**1. Datasets without `metadata["content"]` all collapse onto one address, which regresses the existing Problem 08 scenario.**

`content_address` binds only `owner`, `access_tier`, `schema_version`, `parents`, and the structural digest of `metadata["content"]`, using `None` when the key is absent. But the merged `provenance_supply_chain` scenario publishes content-less datasets throughout: `SourceAgent` does `DatasetMetadata(name=..., owner=self._id, description=self._description)`. Under `sic_facts`, the attacker's forged republish there (`description="tampered-by-attacker"`, same owner, no parents) computes the same `df://sic-` URL as the original, so `validate_provenance_substitution_resistant` flips to FAIL. So "All Problem 08 base criteria remain intact" holds for your scenario, but not as a drop-in for the existing one: a dataset whose identity lives in `checksum`/`description` bytes has no identity at all. Suggested fix: when `metadata["content"]` is absent, fall back to binding `checksum` (and `description`), or refuse the publish with a `CanonicalizationError` so the failure is loud. Either way, run `scenarios/provenance_supply_chain.yaml` with `datafacts: sic_facts` and show it stays green.

**2. Each agent handle gets its own clock; the factory docstring says otherwise.**

Your factory reuses `_build_datafacts_handles`, which shares the tick source via `getattr(handle, "clock", ...)`. `CidFacts` exposes a public `clock` property for exactly this; `SicFacts` only has `_clock`, so every per-agent handle silently constructs its own `SharedClock` while sharing `datasets`/`proofs`. Nothing in the committed scenario calls `verify_freshness`, so this is invisible today, but cross-agent freshness windows are meaningless with divergent ticks, and your factory docstring claims "one logical clock". Add the `clock` property and ideally a freshness probe in the scenario.

**3. Entry point missing from `packages/nest-plugins-reference/pyproject.toml`.**

You registered `sic_facts` in the `_BUILTINS` fallback in `nest_core/plugins.py`, which works, but api_fit expects the entry point under `[project.entry-points."nest.plugins.datafacts"]`, where `cid_facts` already lives. One line.

**4. No property-based tests, and canonicalization is the ideal hypothesis target.**

The 13 example tests are well chosen, but `test_cid_facts.py`, the baseline you are beating, already pulls in hypothesis. A strategy generating arbitrary JSON-ish structures plus a random re-encoder (key shuffle, int to float below 2**53, NFC/NFD) asserting digest equality, and a mutation strategy asserting inequality, would materially lift test rigor. It would also pin down the `_INT_SAFE_FLOAT` edge: `{"v": 2**53 + 2}` and its float re-encoding digest differently today, so the invariance guarantee quietly stops at 2**53. Possibly the right call, but say so in the docstring.

**5. Any agent can zero out anyone's freshness by republishing.**

`publish` overwrites `self._proofs[url]` unconditionally and `verify_freshness` requires signer == owner, so a republish claiming someone else's `owner` flips their dataset to not-fresh until they republish (`test_forged_freshness_by_outsider_fails` documents this, correctly fail-closed). `cid_facts` has the same clobber, but structural addressing makes any re-encoding land on the same URL, so the griefing is far easier to trigger. Cheap hardening: keep the stored proof unless the new signer matches the kept record's owner. Related question: since `owner` is part of the address, a launderer who re-encodes and also flips `owner` to themselves still mints a fresh URL. Deliberately out of scope because ownership changes are attributable? Worth a docstring line.

**Suggested path forward:**

1. Checksum fallback (or fail-loud) for content-less datasets, plus the `provenance_supply_chain` regression run under `sic_facts`.
2. Add the `clock` property and a freshness check in the scenario.
3. Add the pyproject entry point.
4. Add hypothesis round-trip properties for `structural_digest`.
5. CI has not run yet (no checks reported, likely awaiting first-contribution approval). Your ci-local receipts look right; confirm green once Actions runs.

Leaving this open. Ping me on the hackathon channel if the fallback semantics in point 1 need discussion.

maurathat:
Thank you — WHere would I comment in Discord?  What is your Discord handle?  

Thank you for your review --  point 1 is a definitely a bug. You are
right that it's the same class of error this PR is critical of, from the opposite
direction: byte addressing splits identical content, and `sic_facts` as written
merges *different* content whenever `metadata["content"]` is absent. A
content-less dataset gets no structural identity, so every one of them from the
same owner and tier collapses onto a single `df://sic-` URL, and
`validate_provenance_substitution_resistant` correctly flips to FAIL. I missed
that `provenance_supply_chain` publishes without `content`.

On the fallback semantics (your offer to discuss): I'd like to argue for
fail-loud over a checksum fallback, and I'd welcome pushback.  -- I dont know your Discord Handle to comment in the Nanda Discord...

A `checksum` fallback would restore identity for those datasets, but it
reintroduces byte-sensitivity through a side door — the same dataset re-exported
gets a new checksum, hence a new address, which is exactly the laundering channel
this plugin exists to close. A plugin whose invariance guarantee holds only for
some datasets is harder to reason about than one that refuses the datasets it
cannot address.

I propose: `publish()` raises `CanonicalizationError` when
`metadata["content"]` is absent, with a message saying a dataset with no
structural content has no structural identity, and pointing at `cid_facts` for
byte-addressed use. The failure is loud, at publish time, at the boundary of what
structural addressing can honestly claim.

The consequence is that `sic_facts` is *not* a drop-in for `provenance_supply_chain`
— it's a plugin for content-bearing datasets — and I'd document that plainly in
the module docstring. If you think it should be a drop-in,
the checksum fallback is straightforward and I can take that direction instead;
I just want the docstring to state that invariance does not hold on the
fallback path. Happy to talk it through on the channel.

On the rest, all fair, all queued:

2. **Clock.** Correct — `_build_datafacts_handles` reads `handle.clock` via
   `getattr`, `CidFacts` exposes it, `SicFacts` doesn't, so each handle silently
   builds its own `SharedClock` while sharing `datasets`/`proofs`. Invisible today
   only because nothing calls `verify_freshness`. Adding the property and a
   freshness probe to the scenario.

3. **Entry point.** One line under
   `[project.entry-points."nest.plugins.datafacts"]`. Adding it.

4. **Property tests.** Agreed, and thank you for the `2**53` catch. The
   invariance guarantee does stop at `_INT_SAFE_FLOAT` — `{"v": 2**53 + 2}` and
   its float re-encoding digest differently, because beyond that bound
   `float(int(x))` is lossy and collapsing them would make the encoding
   non-injective. That's the intended trade (injectivity over invariance at the
   boundary) but it was never stated. It'll be in the docstring, and hypothesis
   round-trip properties (arbitrary JSON-ish structure + random re-encoder for
   equality; mutation for inequality) will pin it.

5. **Freshness clobber.** Right, and structural addressing does make the griefing
   easier — every re-encoding lands on the same URL, so any peer can knock a
   dataset out of freshness by republishing under the owner's name. Hardening
   `publish()` to keep the stored proof unless the new signer matches the kept
   record's owner.

   On your related question: yes, deliberately out of scope, and it deserves a
   docstring line. `owner` is part of the address, so a launderer who re-encodes
   *and* reassigns ownership to themselves mints a fresh URL. That's by design —
   the address answers "is this the same content from the same owner," and an
   ownership change is an attributable, signed act rather than a silent
   re-serialization. Addressing content independent of owner would make the URL a
   pure content hash and drop the ACL binding, which is a different plugin. I'll
   say so explicitly.

CI still shows no check runs — I believe it's the first-contribution approval
gate. `make ci-local` is green at `249b65a` (1018 passed, ruff, format, pyright)
after merging today's `upstream/main`. Happy to have Actions confirm whenever
it's approved.

I'll push the five fixes and re-request review.

maurathat:
All five points addressed at `4db7e43` (`make ci-local` green: **1170 passed**, ruff, format, pyright; today's `upstream/main` merged cleanly first).

**1. Content-less datasets — checksum fallback (your suggested drop-in option).** `content_address` now binds `checksum` + `description` when `metadata["content"]` is absent, so content-less datasets keep distinct identities. Invariance is documented as NOT holding on that path (a re-export mints a new address, exactly as under `cid_facts`) — module docstring and `content_address` docstring both say so plainly. Content-bearing addresses are byte-identical to before, so the receipts in the PR body still reproduce. The regression run you asked for — `scenarios/provenance_supply_chain.yaml` flipped to `datafacts: sic_facts`:

```
PASS provenance_chain_integrity - chain resolved to depth 4
PASS provenance_substitution_resistant - attacker's differing content resolved to a distinct URL (df://sic-1f406a3b...)
PASS provenance_freshness_unforgeable - forged freshness claim for df://sic-0c7a2885... was correctly rejected
PASS provenance_chain_unforgeable - publish with phantom parent df://sha256-0000... was correctly rejected
```

It stays green.

**2. Shared clock.** `SicFacts.clock` property added (mirrors `cid_facts`), so `_build_datafacts_handles` genuinely shares one tick source. The scenario now runs a freshness probe (audit agent publishes a dataset it owns, verifies it over the shared clock) with a fourth registered validator, `validate_sic_freshness_signed` — scenario passes 4/4.

**3. Entry point.** `sic_facts` registered under `[project.entry-points."nest.plugins.datafacts"]` beside `cid_facts`; the `_BUILTINS` fallback stays, both resolution paths work.

**4. Property-based tests.** Hypothesis strategies for arbitrary JSON-ish structures plus a structure-preserving re-encoder (key shuffle, int→float ≤2^53, NFC→NFD): digest invariance under re-encoding, sensitivity under single-leaf mutation, and an injectivity spot-check (digest equality ⟺ canonical-encoding equality), 200 examples each. The `_INT_SAFE_FLOAT` boundary is now documented in `_canon_number` ("injectivity is preserved in preference to invariance at that boundary") and pinned by an explicit example test: `{"v": 2**53 + 2}` digests differently from its float re-encoding; `2**53` itself still collapses.

**5. Freshness clobber.** `publish` keeps the stored proof unless the new signer matches the kept record's owner (signer read off the signature itself — the Identity protocol exposes no "who am I"). `test_forged_freshness_by_outsider_fails` became `test_forged_freshness_by_outsider_cannot_clobber`: Mallory's republish no longer replaces the owner's proof and the owner's dataset stays fresh. The ownership-in-address rationale is now a module-docstring scope note, verbatim from the discussion above.

@dhve re-requesting review — and thank you again; point 1 was a real drop-in regression this PR would have shipped.

maurathat:
@dhve — following up on the five review points; all addressed at 4db7e43, make ci-local green (1170 passed), and provenance_supply_chain under sic_facts passes 4/4. Whenever you have a window

Ty

JamesCarnley:
Thank you, @maurathat. The reusable future contract is structural identity: semantically identical datasets with different serialization must resolve to one identity, a leaf mutation must change it, empty inputs need explicit non-colliding handling, and freshness/merge cannot depend on each writer's private clock. Current Town fingerprints each exact signed evidence payload; it does not claim a dataset DAG, structural canonicalization, aliases, or branch/merge semantics. Those rules would define a new data-facts profile rather than repair current single-fact evidence, so I am closing the legacy PR and retaining its permutation, mutation, empty-input, and clock cases for future work. No contributor code was merged or executed during this review.

maurathat:
Thanks — that's a fair line to draw. Exact signed-payload fingerprinting is the right identity for evidence; structural identity is a dataset-level property, and I agree it belongs in its own profile rather than a patch to single-fact evidence. Glad the permutation, mutation, empty-input, and clock cases are useful. I have a draft data-facts profile that encodes those four rules as conformance clauses; happy to open it as an issue or discussion if that's a useful place to continue.

PR #73, opened 2026-07-06
[Hackathon] capsec-engineer: offline-attenuating capability tokens with epoch-fenced cascading revocation

zukhriddingit:
# NandaHack Step 1: capsec-engineer offline attenuation for auth

## Motivation

The reference `jwt` auth plugin signs exact tokens and revokes exact token strings. That is useful scaffolding, but it is not a capability system: a holder cannot attenuate authority offline, and revoking a parent does not invalidate children because there is no parent-child cryptographic relation to inspect.

This PR adds `auth: capability_tokens`, a Biscuit/macaroon-style auth plugin for Nanda Town's capability-delegation problem. The security objective is deliberately narrow and testable: authority only shrinks as it moves down a strict delegation tree, and stale revocation knowledge fails closed.

## Design

`CapabilityTokens.issue(subject, scopes)` creates a root link signed with `HMAC(root_secret, canonical_root_caveats)`. `delegate(parent_token, audience, scopes_subset, ttl)` parses the parent token and mints a child link by signing canonical child caveats with the parent link signature as the HMAC key. The root secret is not used for child minting, so a holder can attenuate offline without issuer re-issuance.

Verification is a full replay from the root secret. For each link, the verifier checks the expected HMAC, the embedded parent chain hash, scope monotonicity, TTL monotonicity, expiry, audience binding, and revocation. Every link's chain hash is derived from the previous chain hash, canonical caveats, and signature. Revocation stores the chain hash of the revoked link, so revoking an ancestor invalidates descendants by construction: a descendant verify necessarily replays and encounters the revoked ancestor hash.

Revocations carry a monotonic epoch in `RevocationStore`. A verifier tracks a local visible epoch and is configured with `stale_after`. If `store.epoch - visible_epoch > stale_after`, verification raises `RevocationViewStaleError` rather than accepting on stale knowledge. That is the explicit partition semantics: availability loses to fail-closed authorization.

## Differentiation

What sets this apart from a plain HMAC-chained macaroon is that two additional guards are load-bearing in the adversarial scenario: the `stale_after` / `RevocationViewStaleError` epoch fence treats a stale revocation view as an attack surface and fails closed, while `authorize()` binds presenter verification and a required resource scope in one confused-deputy guard. The scenario makes the epoch fence observable with a real partition-group split and `partition_heal_at_tick: 80`, rather than proving it only in an isolated unit test.

## Threat Model

| Attack | Mechanism that stops it | Test/validator proving it |
|---|---|---|
| Scope escalation by requesting `admin:all` from a `read` parent | `delegate` rejects scopes not subset of parent; verifier also rejects widened links as invalid chains | `test_attack_scope_escalation_denied_before_child_exists`; `delegated_auth_scope_escalation_blocked` |
| TTL extension beyond parent lifetime | `delegate` rejects child expiry later than parent expiry; verifier rejects outliving child links | `test_attack_ttl_extension_denied_at_delegation_boundary`; property attenuation monotonicity |
| Stale parent / revoked ancestor with child replay | Verify checks every replayed ancestor chain hash against the revocation view | `test_attack_revoked_ancestor_cascades_to_child`; `delegated_auth_cascading_revocation`; revocation completeness property |
| Audience confusion / stolen token presentation | `verify_for_audience` and per-agent `verify` bind final token audience to presenter | `test_attack_audience_confusion_rejected`; `delegated_auth_audience_confusion_blocked` |
| Confused deputy using its own token for a third-party unscoped action | `authorize(token, presenter, required_scope)` rejects actions outside the token's verified scopes; scenario records `payments:write` rejection for a `payments:read` deputy token | `test_attack_confused_deputy_unscoped_resource_rejected`; `delegated_auth_confused_deputy_blocked` |
| Partitioned verifier with stale revocation view | Monotonic epoch fence raises `RevocationViewStaleError` when the local view is older than `stale_after` | `test_attack_partitioned_revocation_view_fails_closed`; `delegated_auth_epoch_fence_fail_closed` |
| Byte-level chain tampering | Canonical JSON plus HMAC replay detects any selected token mutation | `test_property_chain_tamper_detection_flipping_any_selected_byte_fails_verify` |

## Scenario

`scenarios/capability_tokens_delegated_auth.yaml` builds one coordinator, three intermediaries, and twelve leaves. The coordinator issues the root token and delegates to intermediaries; intermediaries attenuate offline to their leaves. The run records pipe-delimited `auth:<kind>|key=value` broadcasts for validator evidence, performs a mid-run revocation of `intermediary-1`, and exercises the stale-epoch partition fence.

The unique `capability_tokens_delegated_auth` scenario name preserves the `delegated_auth` implementation already merged on `main`; both auth submissions and their registries coexist after the rebase. These validators are intentionally adversarial and fail on missing evidence, so `auth: jwt` fails the suite instead of getting a vacuous pass.

## Tradeoffs

Strict tree, not DAG: each token has one parent. That keeps replay, revocation, and evidence simple enough for deterministic scenario validation. DAG delegation would need multi-parent caveat semantics and a more complex revocation proof surface.

Fail-closed epoch fence, not best-effort availability: a partitioned verifier may reject a valid token if its revocation view is too old. That is intentional for a capability-security posture; stale authorization is treated as an attack surface, not a convenience problem.

Stdlib HMAC, not production crypto packaging: the implementation uses only `hmac`, `hashlib`, `json`, and `base64`, matching the hackathon constraint. It is a deterministic simulation plugin, not a production token format.

## Verification

```bash
uv sync
uv run pytest -q packages/nest-plugins-reference/tests/test_capability_tokens.py
uv run python -m nest_core.cli run scenarios/capability_tokens_delegated_auth.yaml --seed 42
uv run python - <<'PY'
from pathlib import Path
print([
    f"{r.name}:{r.passed}"
    for r in validate_trace(
        Path("traces/capability_tokens_delegated_auth.jsonl"),
        "capability_tokens_delegated_auth",
    )
])
PY
```

Full local gate used after rebasing onto `main`:

```bash
```

Result: `1167 passed, 1 skipped, 1 deselected` (`ruff` clean, format clean, Pyright `0 errors`).

### DIFF
--- a/docs/layers/auth.md
+++ b/docs/layers/auth.md
@@ -42,6 +42,35 @@ Source: [`nest_plugins_reference/auth/delegatable.py`](../../packages/nest-plugi
 Validators: [`nest_plugins_reference/validators/delegation_validators.py`](../../packages/nest-plugins-reference/nest_plugins_reference/validators/delegation_validators.py).
 Scenario: [`scenarios/delegated_auth.yaml`](../../scenarios/delegated_auth.yaml).
 
+## Capability delegation plugin: `capability_tokens`
+
+`capability_tokens` is an HMAC-chained, macaroon-style capability-token
+plugin for offline attenuation. A root token is signed by the issuer. Each
+holder can delegate a child token without an issuer round-trip by signing the
+child caveats with the parent link signature as the HMAC key. Verification
+replays the entire chain from the root secret, so scope widening, TTL
+extension, signature tampering, and parent-hash substitution fail closed.
+
+The plugin satisfies the base `Auth` protocol and adds explicit audience and
+resource guards:
+
+```python
+auth = CapabilityTokens(secret=b"root", clock=0.0)
+root = await auth.issue(AgentId("coordinator"), ["read", "write"])
+child = await auth.delegate(root, AgentId("worker"), ["read"], ttl=60)
+ctx = await auth.verify_for_audience(child, AgentId("worker"))
+ctx = await auth.authorize(child, AgentId("worker"), "read")
+```
+
+Revocation is by chain hash. Revoking any ancestor causes every descendant
+to fail verification with `RevokedAncestorError` once the verifier has a
+fresh revocation view. Revocation views also carry a monotonic epoch: a
+verifier configured with `stale_after=0` and cut off from the latest epoch
+raises `RevocationViewStaleError` instead of accepting with stale knowledge.
+
+Source: [`nest_plugins_reference/auth/capability_tokens.py`](../../packages/nest-plugins-reference/nest_plugins_reference/auth/capability_tokens.py).
+Scenario: [`scenarios/capability_tokens_delegated_auth.yaml`](../../scenarios/capability_tokens_delegated_auth.yaml).
+
 ## Writing your own
 
 See [`writing-a-plugin.md`](../writing-a-plugin.md). Register under
--- a/docs/superpowers/specs/2026-07-10-pr-73-review-fixes-design.md
+++ b/docs/superpowers/specs/2026-07-10-pr-73-review-fixes-design.md
@@ -1,0 +1,53 @@
+# PR #73 Review Fixes Design
+
+## Goal
+
+Make the capability-token submission mergeable and satisfy every actionable maintainer request without changing its security model or the public `Auth.verify(token)` protocol.
+
+## Rebase and Conflict Resolution
+
+Rebase `hackathon/capsec-engineer-offline-attenuation` onto the latest upstream `main`. Resolve shared-registry conflicts additively in `plugins.py`, `scenarios.py`, and `validators.py`: retain all registrations already present on `main` and restore the capability-token plugin, delegated-auth scenario, and delegated-auth validator registrations exactly once. If upstream has merged another `delegated_auth` scenario, preserve it unchanged and register this submission's scenario under the unique `capability_tokens_delegated_auth` name.
+
+The existing untracked `PR_BODY.md` and `report.html` files are user-owned and must remain untouched.
+
+## Trace Evidence Format
+
+Replace the ambiguous `auth:<kind>:key=value:key=value` encoding with:
+
+```text
+auth:<kind>|key=value|key=value
+```
+
+The colon remains only between the `auth` namespace and event kind. A pipe separates fields, so scope values such as `admin:all`, `alpha:read`, and `payments:read` survive parsing unchanged. The emitter and parser change together; no capability-token cryptography or runtime authorization behavior changes.
+
+## Validator Coverage
+
+Strengthen `validate_delegated_auth_scope_escalation_blocked` so a passing result requires trace evidence with all three facts on the same row:
+
+- `requested == "admin:all"`
+- `rejected == "1"`
+- `error == "ScopeEscalationError"`
+
+This makes the separator fix load-bearing: truncating the scope value causes the validator to fail. Existing end-to-end scenario tests exercise the assertion with the capability-token plugin and continue proving that the baseline JWT plugin fails the adversarial validators.
+
+## PR Differentiation
+
+Add a concise PR-body paragraph distinguishing the submission from a plain HMAC-chained macaroon. It will name:
+
+- the fail-closed `stale_after` / `RevocationViewStaleError` epoch fence;
+- the `authorize()` resource guard for confused-deputy prevention; and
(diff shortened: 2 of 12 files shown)


dhve:
Thanks @zukhriddingit, this is one of the more complete Problem 04 submissions I have read. The core delegation move in `delegate()` is exactly right: signing the child caveats with the parent link signature as the HMAC key (`_hmac_hex(_sig_key(parent.sig), "child", child_body)`) instead of the root secret is what gives you real offline attenuation, and the full replay in `_replay()` re-derives every chain hash so a widened or outliving link is rejected structurally rather than by a flag. Two things I want to call out specifically: `test_default_jwt_fails_delegated_auth_validators` proves the validators actually bite (fails on the baseline, passes on your plugin, which is the whole point of the mandatory adversarial validator), and you extended the layer with `verify_for_audience` and `authorize` rather than changing the `Auth.verify(token)` signature, so you did not break the protocol for every other auth plugin. The `RevocationViewStaleError` epoch fence and the property tests (`test_property_revocation_completeness_kills_every_descendant`, tamper detection) are strong.

This is close, but it cannot merge in its current state. Four things to work through:

**1. The branch conflicts with `main` and needs a rebase.**

`mergeable` is `CONFLICTING` (`mergeStateStatus: DIRTY`). With roughly twenty open PRs all editing the same auth wiring, the collision is almost certainly in the shared registries you touch: `_BUILTINS` in `packages/nest-core/nest_core/plugins.py`, the `_try_load_builtin` branch in `scenarios.py`, and the validator registry near the bottom of `validators.py`. Rebase onto latest `main` and re-resolve those three spots. The panel does not score a PR that does not merge cleanly.

**2. No CI checks are reported on the branch.**

`gh pr checks` returns "no checks reported on the branch", so I cannot confirm the gate is green from here. Your PR body says `753 passed, 1 skipped` locally, which is great, but please make sure CI actually runs and is green after the rebase, since the judge panel does not score a broken PR. Run `uv sync && uv run ruff check . && uv run ruff format --check . && uv run pyright && uv run pytest -v` on the rebased branch and confirm all stages pass.

**3. The trace field separator collides with scope values.**

In `delegated_auth.py`, `_emit()` joins `k=v` pairs with `:`, and in `validators.py`, `_parse_auth_events` splits the whole message on `:`. But your scope values contain colons (`admin:all`, `alpha:read`, `payments:read`), so a field carrying a scope is silently truncated: `_emit("attack_scope_escalation", {"requested": "admin:all", ...})` parses back as `requested=admin`, and `all` is dropped. It happens to be harmless today only because no validator reads a scope-valued field and you renamed the confused-deputy field to `resource=payments.write` with a dot to dodge this. That is a latent trap: the next validator that needs to read a scope value will break with no error. Pick a separator that cannot appear in a value (for example `|` between pairs) or encode the values, so the evidence format is robust rather than accidentally safe.

**4. Make the differentiator explicit for the crowded field.**

Problem 04 is the most contested problem in this hackathon, so novelty is judged hard against peers. Your genuine differentiators are the fail-closed epoch fence (`stale_after`, `RevocationViewStaleError`) that treats a stale revocation view as an attack surface, and the `authorize()` resource guard for the confused-deputy case. Both are real and both are tested, which is the hard part. Add a short paragraph to the PR body that names these as what sets this apart from a plain HMAC-chained macaroon, and keep the epoch fence load-bearing in the scenario (the partition group split with `partition_heal_at_tick: 80` is a nice touch; call it out).

**Suggested path forward:**

1. Rebase onto `main`, resolving `plugins.py`, `scenarios.py`, and `validators.py` registrations.
2. Change the `_emit`/`_parse_auth_events` encoding to a separator that cannot appear in a scope value, and add one validator assertion that reads a scope-valued field so the format is exercised.
3. Add the novelty paragraph to the PR body.
4. Run the full five-stage gate on the rebased branch and confirm every stage is green before requesting re-review.

Leaving this open. Ping me on the hackathon channel if the rebase collisions in the shared registries are messy.

JamesCarnley:
Thank you, @zukhriddingit. This is useful requirements evidence for a future delegated-capability profile: authority may be attenuated but never amplified; scope, resource, audience, expiry, and the complete chain must bind at verification; advancing an issuer epoch must invalidate all descendants; and tampered/replayed chains must fail closed. Current Town intentionally uses flat controller-signed Run Grants and has no transferable delegation or revocation epoch. Selecting this mechanism now would introduce new authorization, revocation, time, and serialization policy, so I am closing the legacy PR while retaining it as the consolidated future acceptance profile for this family. No contributor code was merged or executed during this review.

PR #74, opened 2026-07-06
[Hackathon] andrea-cola: medical data exchange scenario for hybrid_x25519 privacy

andrea-cola:
## Summary

- Adds a new Tier-1 scenario, `medical_data_exchange.yaml`, exercising the `hybrid_x25519` plugin's selective-disclosure path in a **medical-data workflow** (1 hospital + 3 insurers + 1 auditor). This covers a use case called out in the plugin's own docstring but not exercised by any existing scenario — `sealed_bid_with_privacy.yaml` only tests `encrypt`/`decrypt`.
- Adds `test_medical_exchange.py` with **14 tests** including a **fifth adversarial check** not present in the existing test suite: **cross-agent proof replay**. A valid proof anchored to patient-A's credential root is replayed against a verifier expecting patient-B's root; the Merkle-root binding defeats it.
- Confirms `noop` fails where `hybrid_x25519` succeeds: noop leaks all plaintext and accepts replayed proofs.

### What it covers

| Test class | # | What it proves |
|---|---|---|
| `TestMedicalSelectiveDisclosure` | 4 | Hospital proves different field subsets to each insurer; hidden fields never leak |
| `TestCrossAgentProofReplay` | 2 | Proof for root-A fails verification under root-B; tampered root fails |
| `TestNoopFailsWhereHybridSucceeds` | 5 | noop leaks plaintext, accepts replayed proofs; hybrid blocks both |
| `TestEncryptedMedicalExchange` | 3 | Full encrypt+prove round-trip, eavesdropper blocked, revoked insurer excluded |

### Design decisions

- **Scenario-only, no plugin changes.** The `hybrid_x25519` plugin is already solid. This PR adds coverage for its selective-disclosure code path in a domain (medical data) that motivates the plugin's existence but had no scenario.
- **Cross-agent replay** is distinct from the existing 4 attacks (eavesdropper, replay, field-injection, stale-revocation). It targets the proof layer rather than the encryption layer: a valid credential proof from one session must not transfer to a session anchored to a different credential root.
- **Deterministic** — salt seeds are explicit, all tests are reproducible.

## Test plan

All CI checks pass locally:

```bash
uv sync
uv run ruff check .          # 0 errors
uv run ruff format --check . # 0 reformats
uv run pyright               # 0 errors
uv run pytest -v             # 790 passed (including 14 new)
```

## Persona

`privacy-engineer` — the test design reflects credential-lifecycle thinking: credential issuance (commitment), selective proving (audience-scoped disclosure), and cross-session replay resistance (root binding). The scenario models a real regulated workflow (medical data exchange under selective disclosure) rather than a generic test harness.

### DIFF
--- a/packages/nest-plugins-reference/tests/test_medical_exchange.py
+++ b/packages/nest-plugins-reference/tests/test_medical_exchange.py
@@ -16,7 +16,27 @@ Coverage layers:
    Insurer-B's verifier with a different root; must fail.
 3. **Adversarial discrimination** — ``noop`` leaks all fields and accepts
    replayed proofs; ``hybrid_x25519`` blocks both.
-4. **Scenario integration** — Boots the YAML and confirms determinism.
+4. **Encrypted medical exchange** — full encrypt/prove/verify round-trip,
+   eavesdropper exclusion, and revocation in a medical context.
+5. **Deeper privacy invariants** — six additional attack surfaces not covered
+   by the existing test suite:
+
+   a. Salt hiding — the same low-entropy field value committed with two
+      different salts produces different leaf hashes, so the root does not
+      leak the value to a passive observer who only knows the schema.
+   b. Reveal-set integrity — a proof built for ``reveal=X`` does not verify
+      against a statement demanding ``reveal=X,Y`` (extra-field mismatch).
+   c. Merkle path substitution — the authentication path from leaf *i* cannot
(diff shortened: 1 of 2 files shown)


shilpa-kulkarni-14:
@andrea-cola CI green but adds no new capability; tests an already-merged plugin, and its net-new scenario cannot run. Please confirm if this is the case.

PR #76, opened 2026-07-07
[Hackathon] geick: typed CRDT memory for concurrent agent writes

geeick:
## Summary

This submission adds a `typed_crdt` memory plugin for problem 02, concurrent memory under unreliable delivery.

The default `blackboard` memory plugin uses last-writer-wins behavior, so concurrent writes to the same key can overwrite each other depending on message delivery order. This plugin makes memory writes self-describing with a `type` field and merges them using deterministic CRDT-style behavior.

Supported memory types:

- `set`: unions tagged facts/items
- `counter`: keeps the maximum count per writer and derives a total
- `vote`: preserves one ballot per writer and derives a deterministic majority result

If two writes use different types for the same key, the plugin rejects the merge instead of silently corrupting memory.

## Files changed

- Added `TypedCrdtMemory`
- Registered the plugin as `memory: typed_crdt`
- Added validator showing blackboard is order-dependent while typed CRDT memory converges
- Added pytest coverage
- Added README documentation

## Validation

I ran the required local CI sequence:

```powershell
uv sync
uv run ruff check .
uv run ruff format --check .
uv run pyright
uv run pytest -v

Results:

uv sync passed
ruff check passed
ruff format --check passed
pyright passed with 0 errors
Full pytest suite passed

I also ran the typed CRDT validator:

python3 validators\validate_typed_crdt_memory.py

The validator shows:

blackboard fails convergence as expected
typed_crdt passes set convergence
typed_crdt passes counter convergence
typed_crdt passes vote convergence

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -39,6 +39,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("negotiation", "pareto"): f"{_REF}.negotiation.pareto:ParetoNegotiation",
     ("memory", "blackboard"): f"{_REF}.memory.blackboard:Blackboard",
     ("memory", "lww_register"): f"{_REF}.memory.lww_register:LwwRegisterMemory",
+    ("memory", "typed_crdt"): f"{_REF}.memory.typed_crdt:TypedCrdtMemory",
     ("privacy", "noop"): f"{_REF}.privacy.noop:NoopPrivacy",
     ("privacy", "hybrid_x25519"): f"{_REF}.privacy.hybrid_x25519:HybridX25519Privacy",
     ("datafacts", "datafacts_v1"): f"{_REF}.datafacts.datafacts_v1:DataFactsV1",
--- a/packages/nest-plugins-reference/nest_plugins_reference/memory/README_typed_crdt.md
+++ b/packages/nest-plugins-reference/nest_plugins_reference/memory/README_typed_crdt.md
@@ -1,0 +1,190 @@
+# Typed CRDT Memory Plugin
+
+## Problem
+
+The default `blackboard` memory plugin uses last-writer-wins behavior.
+
+That means if several agents write to the same key, the final value depends on message delivery order. Earlier writes can be silently overwritten.
+
+Example:
+
+```text
+agent_1 writes fact_1
+agent_2 writes fact_2
+...
+agent_8 writes fact_8
+```
+
+With `blackboard`, the final memory only contains whichever write arrived last.
+
+This is unsafe for multi-agent coordination because NANDA Town scenarios can include concurrent writers, dropped messages, and reordered delivery.
+
+## Solution
+
+`typed_crdt` is a memory plugin that merges values according to the declared memory type.
+
+Instead of treating every key as a normal value, each write includes a `type` field. The first write establishes the type for that key. Later writes to the same key must use the same type.
+
+Supported types:
+
+| Type | Merge behavior | Use case |
+|---|---|---|
+| `set` | Union tagged items | Shared facts, catalogue entries, presence |
+| `counter` | Keep max count per writer and derive total | Vote totals, event counts, reports |
+| `vote` | Preserve one ballot per writer and derive majority result | Agent decisions, reputation labels, approvals |
+
+If two agents write different memory types to the same key, the plugin rejects the write instead of guessing.
+
+## Example: set memory
+
+First write:
+
+```json
+{"type":"set","writer":"agent_1","value":"fact_1"}
+```
+
+Second write:
+
+```json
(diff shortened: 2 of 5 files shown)


dhve:
Thanks @geeick. This is a thoughtful take on problem 02: making writes self-describing with a `type` field and giving each type its own merge (tagged-set union, max-per-writer counter, ballot map with a derived majority) is a genuinely different shape from the merged `lww_register` plugin, and the determinism hygiene is right (sorted keys in `_encode`, no wall clock or unseeded RNG anywhere). The "blackboard fails, typed_crdt passes" pattern shows up in both `test_blackboard_is_order_dependent_for_concurrent_set_writes` and the validator script, which is exactly the discriminating structure the problem asks for.

**1. The `vote` type is not actually convergent when a writer revotes.**

`_merge_vote` does `merged_ballots.update(new_ballots)`, so the incoming write wins per writer. If `agent_1` writes `approve` and then `reject`, delivering those two writes in opposite orders to two `TypedCrdtMemory` instances yields different winners, so forward and reverse disagree on your own convergence check. Your tests never hit this because every writer votes exactly once. The same latent issue exists in `_merge_set` when callers pass an explicit `tag`: two writers using the same tag with different values collide, and merge order decides. (`counter` is safe; max is commutative.) Fix: version each ballot with a writer-local counter and merge by max with a deterministic tie-break, the same `(lamport, node)` trick `lww_register.py` uses. This is the correctness core of the submission.

**2. The mandatory scenario YAML is missing, and the validator lives outside the repo's validator machinery.**

The charter requires a scenario under `scenarios/` (or `examples/<theme>/`) demonstrating your plugin under the adversarial validator; this PR ships none, and `validators/validate_typed_crdt_memory.py` creates a new top-level directory that does not exist on main. The repo already has homes for both pieces: reference validators live in `packages/nest-plugins-reference/nest_plugins_reference/validators/`, `nest_core.validators.validate_crdt_convergence` already implements the fails-on-blackboard convergence check, and `scenarios/memory_concurrent_writers.yaml` is the shape to follow. Note `typed_crdt` will not drop into that scenario as-is: `CrdtWriterAgent` gossips via `export(key)` / `merge(key, state)`, which `TypedCrdtMemory` lacks (`export` can return the stored state, `merge` can delegate to `write`), and the factory constructs `memory_cls(node_id)` and writes plain bytes like `value-from-writer-0`, which `_decode` rejects. Plan on a scenario variant whose writers emit typed JSON.

**3. First-writer type poisoning turns honest writes into exceptions.**

The first write to a key establishes its type, and any later mismatch (or non-JSON bytes) raises `ValueError` out of `write`. The `Memory` protocol in `nest_core/layers/memory.py` does not document `write` as raising, and `blackboard` never does. So a byzantine agent that writes `{"type": "set", ...}` to `decision` first makes every honest vote write raise from then on: a denial of service on the key rather than a rejected merge. The README frames rejection as a feature for honest schema conflicts, which is fair, but your adversarial coverage should include the hostile version.

**4. Housekeeping: branch name, entry point, CI.**

The charter's branch pattern is `hackathon/<handle>-<short-theme>`; `hackathon/memory-bouncer` has no handle, and non-matching branches are auto-closed, so rename to something like `hackathon/geeick-typed-crdt-memory`. Also register the plugin under `[project.entry-points."nest.plugins.memory"]` in `packages/nest-plugins-reference/pyproject.toml` alongside your `_BUILTINS` entry, the way `lww_register` is wired. And no CI checks have reported on this branch yet; your local runs are noted, but the judge panel only scores PRs that are green on GitHub, so confirm the workflow runs after your next push.

**Suggested path forward:**

1. Version ballots (and tagged set items) so merges commute under revotes; add the reorder tests.
2. Add `export`/`merge`, move the validation logic into `nest_plugins_reference/validators/`, and ship a scenario YAML with 8+ writers under `failures.message_drop: 0.1`.
3. Add byzantine-input tests: garbage bytes, hostile first-writer type claim.
4. Fix the branch name, add the pyproject entry point, confirm CI is green.
5. In the PR body, state what differentiates this from merged #22 (`lww_register`): typed, structure-aware merges vs an opaque LWW register. The CRDT memory field is crowded and novelty is scored against merged peers.

Leaving this open. Ping me on the hackathon channel if the ballot-versioning reshaping is unclear.

JamesCarnley:
Thank you, @geeick. A future typed replicated-memory profile must converge when identical operations arrive in opposite orders, including revotes and same-sequence tags from independent replicas, and it must prevent an unauthorized first writer from choosing a key's type. Current Town memory is a per-agent exact-key store with no replication, merge, vector clock, vote CRDT, or cross-agent convergence claim. Those semantics would be a new profile, while this frozen implementation still has order-dependent revotes, tag collision, type-poisoning, and unwired validation concerns. I am closing the legacy PR and retaining those concrete cases for future CRDT work. No contributor code was merged or executed during this review.

PR #77, opened 2026-07-07
Hackathon/a2r commerce checkout frontier

Skyrider3:
## Motivation

Agentic checkout is multi-attribute. A shopping agent negotiating a purchase is trading price against delivery deadline ("I'll pay a premium for expedited"), under a constraint no utility weight can express: a hard wallet budget. Price-only bargaining cannot represent the premium, and unconstrained multi-attribute bargaining cannot represent the wallet. This submission evaluates the **buyer side**  in agent-to-retail commerce the shopping agent is the deployed software; the vendor side is usually a quoting engine.

## What ships

- **Plugin** `("negotiation", "checkout_frontier")` — `packages/nest-plugins-reference/nest_plugins_reference/negotiation/checkout_frontier.py`, registered in `_BUILTINS` and as a `nest.plugins.negotiation` entry point.
- **Scenario** `checkout_market` — `scenarios/checkout_market.yaml` + `packages/nest-core/nest_core/scenarios_builtin/checkout_market.py`: 10 shopper-vendor pairs, scripted margin-only vendors quoting an early expedited "rush" lane, the configured plugin in the buyer seat, hard budgets and cost floors drawn per pair from the scenario seed.
- **Adversarial validators** — `validate_checkout_pareto_efficient` (no closed deal is Pareto-dominated by a *cap-feasible* bundle exchanged in the same session) and `validate_checkout_budget_and_floor` (no quote or deal crosses the quoting agent's own declared money cap). Both fail vacuous traces.
- **Tests** — 17 unit tests; 6 property tests (Hypothesis, 200 examples on the flagship dominance property, plus the pinned-counterexample and disjoint-caps cases); 16 validator/integration tests including the end-to-end discrimination gate on seeds 42/7/1337 and a byte-identical-trace determinism test.

## The mechanism, and the guarantee it buys

`CheckoutFrontier` layers one rule on top of classical monotonic concession (Keeney–Raiffa additive MAUT, Zeuthen aspiration decay, per-attribute patience, direction-reading trade-off): **best-standing-quote acceptance**. Every bundle the counterparty ever offered stays live; once the best cap-eligible one clears the aspiration floor, the agent accepts it if it is the offer on the table, or re-proposes it verbatim (the counterparty running the same rule accepts its own returning quote — one extra round, guaranteed convergence).

That changes the guarantee class. The merged `pareto` plugin (#30) documents its own boundary in `test_fsj_tradeoff_does_not_guarantee_pareto_optimality`: a Pareto-improving bundle offered early can be rejected while aspiration is high and be *gone* by the time aspiration decays — self-play can close a dominated deal, so its self-play guarantee was deliberately downgraded to individual rationality. Standing quotes remove the ephemerality: nothing generous ever expires. The flagship Hypothesis property — **no self-play agreement is Pareto-dominated by any cap-feasible bundle either side exchanged** — runs green at 200 examples, and a dedicated test replays PR #30's pinned counterexample configuration and settles it clean.

Supporting details that make the property hold and replay byte-identically: earliest-offer tie-breaks (under the counterparty's own concession, the earliest of equally-good-for-us quotes is best for *them*), acceptance only when the best standing quote *is* the current offer, fresh counters capped by the previous fresh counter's own utility, and all weights on a six-decimal lattice so no utility gap on the integer grid can fall inside the validators' `1e-9` epsilon.

Hard caps are constraints, not utility terms: a buyer never accepts, echoes, or proposes above `budget`; a vendor never goes below `floor`; with mutually exclusive caps the plugin declines to counter and the checkout breaks down honestly (`close` → `None`) instead of closing an unaffordable deal.

## Relationship to merged PR #30 (novelty delta)

This is the same problem statement (07) solved for a different regime, not a re-issue; nothing in #30 is modified and the two plugins interoperate over the same `Terms` encoding.

| | `pareto` (#30, merged) | `checkout_frontier` (this PR) |
|---|---|---|
| Evaluated seat | scripted buyer vs configured **seller** | scripted vendor vs configured **buyer** |
| Self-play guarantee | individual rationality (dominated deals possible, pinned) | + no agreement dominated by any cap-feasible exchanged bundle |
| Money constraints | none | hard budget / floor, honored in-mechanism and audited from the trace |
| Dominance evidence | all exchanged bundles | cap-feasible exchanged bundles (an unaffordable bundle is not evidence of a better deal) |
| `close()` | never exercised by its driver | exercised on both deal and no-deal paths |
| Registration | `_BUILTINS` only | `_BUILTINS` + pyproject entry point |

## Adversarial demonstration

The scenario's vendors are margin-only (`w_deadline = 0`): fulfillment slack costs an in-stock vendor nothing, so the expedited lane is free to give away and its schedule offers rush delivery in rounds 2–3 while price descends toward clearance. `checkout_frontier` buyers (deadline weight drawn in [0.85, 0.95], budget in [110, 120]) accept the rush logroll by round 3 on every seed — round 2's price can exceed a low budget draw, so the cap visibly bites in-trace before the cheaper rush re-offer lands. `alternating_offers` never reads `conditions["deadline_days"]` and only accepts at its round limit, closing late on a cheaper slow-lane bundle that the rush quote dominates for **both** sides (strictly better for a margin-only vendor, strictly better for every drawn deadline weight). The dominance is arithmetic over the drawn ranges, not a lucky seed; the gate still runs three seeds:

- `negotiation: checkout_frontier` → both validators PASS (seeds 42, 7, 1337)
- `negotiation: alternating_offers` → `checkout_pareto_efficient` FAILS with the dominating rush bundle named in the detail (seeds 42, 7, 1337)

## Verification

```bash
uv sync
uv run ruff check . && uv run ruff format --check .
uv run pyright
uv run pytest -v
uv run nest run scenarios/checkout_market.yaml
```

```python
from pathlib import Path
for r in validate_trace(Path("traces/checkout_market.jsonl"), "checkout_market"):
    print(r.name, r.passed, r.detail)
```

## Tradeoffs and scope

- Fresh-counter selection enumerates the feasible grid per respond call — fine at catalog scale here (80×20), quadratic in span.
- The shipped scenario always deals by design; the breakdown path is exercised by unit, property (disjoint caps), and validator tests.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -37,6 +37,9 @@ _BUILTINS: dict[tuple[str, str], str] = {
         f"{_REF}.negotiation.alternating_offers:AlternatingOffers"
     ),
     ("negotiation", "pareto"): f"{_REF}.negotiation.pareto:ParetoNegotiation",
+    ("negotiation", "checkout_frontier"): (
+        f"{_REF}.negotiation.checkout_frontier:CheckoutFrontier"
+    ),
     ("memory", "blackboard"): f"{_REF}.memory.blackboard:Blackboard",
     ("memory", "lww_register"): f"{_REF}.memory.lww_register:LwwRegisterMemory",
     ("privacy", "noop"): f"{_REF}.privacy.noop:NoopPrivacy",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -107,6 +107,10 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("multi_attribute_market", multi_attribute_market_factory)
+    elif name == "checkout_market":
+        from nest_core.scenarios_builtin.checkout_market import checkout_market_factory
+
+        register_scenario("checkout_market", checkout_market_factory)
     elif name == "provenance_supply_chain":
         from nest_core.scenarios_builtin.provenance_supply_chain import (
             provenance_supply_chain_factory,
--- a/packages/nest-core/nest_core/scenarios_builtin/checkout_market.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/checkout_market.py
@@ -1,0 +1,416 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Checkout market scenario: buyer-side evaluation under budget caps.
+
+Ten agentic checkouts, each between a *scripted* margin-only vendor and the
+*configured* negotiation plugin in the buyer seat. This inverts the merged
+``multi_attribute_market`` evaluation (scripted buyer, configured seller):
+here the shopping agent is the artifact under test, because in agent-to-retail
+commerce the buyer is the deployed software and the vendor side is a quoting
+engine.
+
+The vendor is deliberately **margin-only** (``w_deadline = 0``): fulfillment
+slack costs an in-stock vendor nothing, so expedited shipping is free for it
+to give away. Its quote schedule concedes price monotonically toward a
+clearance target, offering the **expedited lane twice, early** (the rush
+rounds) and a slow lane otherwise. A deadline-aware, budget-disciplined buyer
+grabs the rush logroll the moment it fits both its aspiration and its wallet;
+a deadline-blind, timeout-driven buyer (the reference ``alternating_offers``,
+which never reads ``conditions['deadline_days']`` and only accepts at its
+round limit) holds out and closes late on a cheaper slow-lane bundle that the
+early rush quote Pareto-dominates for *both* sides.
+
+Why the dominance is seed-independent: the buyer's deadline weight is drawn
+from ``[0.85, 0.95]``, so the rush bundle beats any slow-lane bundle for the
+buyer by at least ``0.85 * (slow - rush) / d_span - w_p * price_gap / p_span``
+which is positive for every draw here, and the vendor is margin-only, so the
+rush round's higher price beats every later round's lower price outright.
+The rush price also sits at or below every drawn budget by construction. The
+buyer's aspiration ``patience ** t`` is 0.9 in round 2 and 0.81 in round 3;
+the rush bundle's buyer utility is at least 0.8725 for every drawn weight, so
+round 3 guarantees acceptance even when round 2's price momentarily exceeds a
+low budget draw (the cap visibly bites, then the cheaper rush re-offer lands).
+
+Every exchanged bundle, both parties' private utility parameters, and their
+hard money caps are written to the trace so the offline validators can
+reconstruct utilities, compute cap-feasible dominance, and audit budget and
+floor discipline. Frame grammar (floats to 6 dp for byte-determinism; the
+cap is the buyer's budget or the vendor's floor)::
+
+    checkoutcfg:<agent>:<side>:<w_price>:<w_deadline>:<plo>:<phi>:<dlo>:<dhi>:<reservation>:<cap>
+    quote:<sid>:<agent>:<side>:<round>:<price>:<deadline>
+    deal:<sid>:<price>:<deadline>:<accepting_agent>
+    no_deal:<sid>:<rounds>
+
+The vendor drives the whole exchange synchronously inside ``on_start`` (no
+cross-agent event interleaving to go non-deterministic), emitting authentic
+``ctx.send`` frames. All randomness is drawn in the factory, before any agent
+runs, from a generator seeded only by ``(config.seed, pair)`` with a fixed
+draw order (buyer weight, budget, floor, then the slow-lane deadlines); that
+order is load-bearing for byte-identical replays.
+
+Example::
+
+    from nest_core.runner import ScenarioRunner
+    runner = ScenarioRunner(ScenarioConfig.from_yaml("scenarios/checkout_market.yaml"))
+    await runner.run()
+"""
+
+from __future__ import annotations
+
+import inspect
+import random
+from typing import Any
+
+from nest_core.scenario import ScenarioConfig
+from nest_core.sim.agent import AgentContext, StateMachineAgent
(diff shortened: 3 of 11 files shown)


dhve:
Thanks @Skyrider3. This is one of the strongest negotiation submissions I've read. Best-standing-quote acceptance in `checkout_frontier.py` is a genuinely elegant answer to the ephemerality boundary that the merged `pareto` plugin pins in `test_fsj_tradeoff_does_not_guarantee_pareto_optimality`, and you made the upgrade load-bearing instead of just claiming it: I cross-checked the pinned counterexample config in `test_flagship_holds_on_the_pinned_pareto_counterexample` against `test_pareto_negotiation.py` on main and the values match exactly. The both-direction e2e gate (checkout_frontier passes all validators, alternating_offers fails `checkout_pareto_efficient` with the dominating bundle named in the detail, seeds 42/7/1337), the vacuous-trace guards in both validators, the cap-feasibility filter on the dominance frontier, and the byte-identical-trace test are all exactly what the rubric wants. CI is green and the branch is clean against main.

On the novelty question you preempted: I read the delta table against merged #30 carefully, and I buy it. Buyer-seat evaluation, hard money caps as constraints rather than utility terms, and a guarantee class #30 explicitly documents as out of its reach is a different regime, not a re-issue, and you touched nothing in #30 (zero deletions in the whole diff). The registration edits in `nest_core/plugins.py`, `scenarios.py`, and `validators.py` mirror the `multi_attribute_market` pattern, so no scope concern there either.

This is close. Three things to work through before merge:

**1. A buyer can accept a deal priced above its own budget: cap eligibility is checked on the clamped bundle, but acceptance transacts the raw terms.**

In `respond`, the current offer is evaluated as `current = self._clamp(*self._extract(opponent))` and eligibility as `self._eligible(current[0])`; `offer()` also stores standing quotes post-clamp. So an off-grid quote priced above `price_range[1]` clamps down to `phi`, becomes cap-eligible whenever `budget >= phi`, and once that clamped bundle is the best standing quote clearing aspiration, the plugin returns `accepted=True` on the raw offer. `close()` then embeds the raw `session.current_terms` in the Agreement. I reproduced it against your file: buyer with `budget=120`, `price_range=(40, 120)`, weights `{"price": 0.1, "deadline": 0.9}`, vendor quotes `(500, 1)` twice; round 2 accepts and the agreement price is 500. The vendor side mirrors it: `floor=40`, raw quote priced 1, accepted at 1. This directly contradicts the docstring invariant ("a buyer never accepts, echoes, or proposes a bundle priced above budget") and `test_buyer_never_accepts_or_counters_above_budget`, which only feeds in-range prices. Note the nice irony: your own `validate_checkout_budget_and_floor` would flag the resulting deal frame, so the validator catches a violation your plugin can actually commit. Fix: gate acceptance (and standing-quote recording) on the raw extracted price, e.g. check `_eligible` before clamping, or treat out-of-range prices as non-offers and decline. While you're there, decide whether the `opponent.price is None` auto-accept fallback should survive when a cap is set; accepting a priceless bundle sits oddly with the "a checkout must not capture payment without both signatures" stance in `close()`, reference-plugin parity or not.

**2. The Hypothesis strategies never leave the feasible grid, so the adversarial input surface of `respond` is untested.**

`_cfg_and_offers` draws offer prices from `st.integers(cfg.plo, cfg.phi)` and deadlines likewise, and the flagship dominance property drives self-play, where both agents quote on-grid by construction. So nothing in the suite ever hands `respond` an off-grid quote, which is exactly where finding 1 lives. Widen the offer strategy to include off-grid prices and deadlines (and a missing-price `Terms()` case), add a raw-offer-feed property asserting no acceptance or counter ever crosses the agent's own cap, and pin the 500-credit repro as a regression test. Cheap change, and it moves the byzantine-input needle the test_rigor dimension looks for.

**3. Branch name doesn't match the charter pattern.**

The charter asks for `hackathon/<handle>-<short-theme>`; yours is `hackathon/a2r-commerce-checkout-frontier` with no handle, and mismatched names are subject to auto-close under the charter rules. Renaming the branch to something like `hackathon/skyrider3-checkout-frontier` through GitHub's branches page keeps the PR attached. Same nit for the title: "Hackathon/a2r commerce checkout frontier" is the raw slug; a descriptive title like "negotiation: budget-capped checkout frontier, buyer-side (problem 07)" reads much better.

**Suggested path forward:**

1. Move the cap check to the raw quoted price on both the accept path and standing-quote recording, for both sides, and settle the missing-price fallback.
2. Widen the property strategies to off-grid and malformed offers, assert cap discipline over those feeds, and add the pinned regression case.
3. Rename the branch via the GitHub UI and retitle the PR.
4. Re-run the full local gate (`uv run ruff check .`, `uv run ruff format --check .`, `uv run pyright`, `uv run pytest -v`) and confirm green before requesting re-review.

Leaving this open. Ping me on the hackathon channel if any of this is unclear; the fix for finding 1 is small and the rest of this is in very good shape.

JamesCarnley:
Thank you @Skyrider3. We re-reviewed the complete frozen patch and discussion against rebuilt current `main`.

The hard buyer-cap requirement remains useful and exposed two current problems: the stock buyer could proceed when a seller accepted above its cap, and the evaluator could count acceptance, order, and settlement records without proving they belonged to the same transaction. We preserved that selected requirement in a fresh implementation in #256, now merged as `53178b9`. It rejects above-cap purchases before order/escrow creation and replayably binds each negotiation ID to one purchase order, order ID, and exact escrow settlement. The current-base suite passed 655 tests, all three GitHub CI jobs passed, and independent review found no blockers.

This does not mean the original source code was merged or the full original guarantee was adopted. The `checkout_frontier` plugin, multi-attribute/Pareto negotiation, standing quotes, `checkout_market` scenario, property suite, and quote-level budget/floor validator are not part of current core. They remain useful acceptance ideas if Town later selects an explicit advanced-negotiation profile.

Closing this legacy implementation now that its selected current requirement is represented on `main`. The source branch is being preserved.

PR #80, opened 2026-07-07
fix: reputation scenario fails its own validator + dashboard SSRF

athiye:
# 1. Fix the `reputation` scenario failing its own validator

Running `nest run reputation` with the default seed of 42 produced a trace that failed the `reputation_scoring` validator. The committed demo trace had the same issue.

The problem was that a malicious agent could cheat another malicious agent. Since malicious agents do not report cheating, the validator failed because it expects every cheat to have a corresponding bad report.

**Fix:** Malicious agents now only initiate trades with honest agents. This ensures that every cheat is received by an honest agent and reported, so the validator passes across all seeds without weakening the validation logic.

The demo trace was regenerated, and a regression test was added that runs every deterministic built-in scenario through its own validators across several seeds. Previously, only the marketplace scenario was covered.

## 2. Fix SSRF in the dashboard SkillMD reachability check

`checkReachable` was fetching user-submitted URLs from the server while only validating the URL scheme. This allowed URLs such as `http://localhost`, private IP ranges, or the cloud metadata endpoint at `169.254.169.254` to be requested from inside the server.

**Fix:** The hostname is now resolved before the request, and loopback, private, link-local, and reserved IP addresses are rejected. Redirects are also followed manually so that every redirect target is validated before being fetched.

Saving is never blocked. Unsafe or unreachable links are simply stored with `reachable=false`.

## Verification

* `ruff` passed
* `pyright` passed
* `pytest` passed with 348 tests
* Dashboard `tsc` passed
* Dashboard `eslint` passed


shilpa-kulkarni-14:
Hard gates fail, suggest resolving the conflicts for further review consideration.

JamesCarnley:
Thank you @athiye. This PR identified the dashboard catalog submission as an SSRF boundary: an untrusted URL could cause a server-side request, including through redirects. The retained requirement has now been independently implemented, tested, reviewed, and merged in #253 (merge 1811275f1a0e6533c381f6ea49ba162e3f96693e), with explicit credit to this source. The current fix removes submission-time URL fetching entirely and records new links as unprobed, which also avoids DNS-rebinding and redirect races instead of relying on an incomplete address denylist. We are closing this archive/legacy PR now that the current gap has landed. The legacy reputation scenario and generated trace were not ported, and no contributor code from this PR was executed or merged.

PR #83, opened 2026-07-07
[Hackathon] nandha-datafacts-verifier: Delegatable capability tokens …

anilchowdary07:
Problem #4: Delegatable Auth (True Macaroon-style Capability Tokens)

This PR introduces an enterprise-grade delegatable capability auth plugin, moving beyond naive delegation to implement a mathematically robust, adversarial-resistant distributed token system.

Architecture Context & Prior Art (#138)
While I submitted this architecture early in the hackathon, I see that PR #138 was recently merged to provide a baseline solution for Problem 04. This PR (#83) remains highly differentiated and pushes the architecture further for high-security environments by introducing Epoch Fencing and strict Resource Binding. Rather than relying on simple TTL expiration like the merged baseline, this implementation introduces a hard cryptographic boundary across epochs, preventing cross-epoch delegation propagation entirely. It also scopes capabilities directly to specific resources via the authorize() resource guard. This introduces a far more robust security posture against persistent threats, supported by the validate_delegation_epoch_fence adversarial validator.

Security Posture vs. Eventual Consistency Alternatives
While other submissions may explore eventual consistency models (such as G-Set CRDTs) for partition tolerance, this architecture explicitly rejects that approach for high-security environments. CRDT-based revocations inherently fail-open during network partitions—meaning a disconnected node will blindly accept revoked tokens until it merges gossip state. By implementing Monotonic Epoch Fencing, this PR intentionally trades eventual consistency for strict security boundaries, ensuring verifiers fail-closed (RevocationViewStaleError) when partitioned. This defensive posture is critical for true capability tokens.

Key Features
Offline Attenuation (delegate API): Implements delegate(parent_token, audience, scopes_subset, ttl) with strict bounds checking. Child tokens are cryptographically signed using their parent's signature as the HMAC key, allowing intermediaries to securely attenuate permissions entirely offline.
O(1) Cascading Transitive Revocation: Eliminates slow recursive database lookups. By anchoring tokens to a cryptographic _chain_hash accumulator, revoking any parent instantly and mathematically invalidates all descendants in O(1) time during verification.
Monotonic Epoch Fencing (Partition Tolerance): Added a global epoch clock to the RevocationStore. If an agent is cut off by a network partition and its view becomes too stale, verify() enforces fail-closed semantics by raising a RevocationViewStaleError.
Adversarial Hardening (Resource Guards): Implemented an authorize() resource guard to actively block Confused Deputy vulnerabilities.
Rigorous Adversarial Validation: Overhauled test_delegatable_auth.py, test_validators.py, and the delegated_auth.py scenario to explicitly simulate and trap 5 attack vectors: Scope Escalation, Stale Parent, Audience Confusion, Confused Deputy, and Epoch Fence Bypasses.
Passes CI locally: 100% test pass rate. Verified against make ci-local (uv sync, ruff format/check, pyright, pytest) with zero linting or type errors.

### DIFF
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -77,10 +77,32 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("identity_rotation", identity_rotation_factory)
+    elif name == "attested_peering":
+        from nest_core.scenarios_builtin.attested_peering import (
+            attested_peering_factory,
+        )
+
+        register_scenario("attested_peering", attested_peering_factory)
     elif name == "gossip_registry":
         from nest_core.scenarios_builtin.gossip_registry import gossip_registry_factory
 
         register_scenario("gossip_registry", gossip_registry_factory)
+    elif name == "gossip_byzantine_forgery":
+        from nest_core.scenarios_builtin.gossip_byzantine import (
+            gossip_byzantine_forgery_factory,
+        )
+
+        register_scenario("gossip_byzantine_forgery", gossip_byzantine_forgery_factory)
+    elif name == "gossip_signed_equivocation":
+        from nest_core.scenarios_builtin.gossip_byzantine import (
+            gossip_signed_equivocation_factory,
+        )
+
+        register_scenario("gossip_signed_equivocation", gossip_signed_equivocation_factory)
+    elif name == "gossip_eclipse":
+        from nest_core.scenarios_builtin.gossip_byzantine import gossip_eclipse_factory
+
+        register_scenario("gossip_eclipse", gossip_eclipse_factory)
     elif name == "memory_concurrent_writers":
         from nest_core.scenarios_builtin.memory_concurrent_writers import (
             memory_concurrent_writers_factory,
@@ -91,6 +113,10 @@ def _try_load_builtin(name: str) -> None:
         from nest_core.scenarios_builtin.comms_versioning import comms_versioning_factory
 
         register_scenario("comms_versioning", comms_versioning_factory)
+    elif name == "comms_downgrade":
+        from nest_core.scenarios_builtin.comms_downgrade import comms_downgrade_factory
+
+        register_scenario("comms_downgrade", comms_downgrade_factory)
     elif name == "receipt_reputation":
         from nest_core.scenarios_builtin.receipt_reputation import (
             receipt_reputation_factory,
@@ -101,6 +127,10 @@ def _try_load_builtin(name: str) -> None:
         from nest_core.scenarios_builtin.empic_payments import empic_payments_factory
 
         register_scenario("empic_payments", empic_payments_factory)
+    elif name == "delegated_auth":
+        from nest_core.scenarios_builtin.delegated_auth import delegated_auth_factory
+
+        register_scenario("delegated_auth", delegated_auth_factory)
     elif name == "multi_attribute_market":
         from nest_core.scenarios_builtin.multi_attribute_market import (
             multi_attribute_market_factory,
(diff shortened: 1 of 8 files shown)


dhve:
Thanks @anilchowdary07. The core of `delegatable.py` is well built: `_derive_child_key` with the `cap/v1/delegate/{nonce}` domain separation, the `_chain_hash` accumulator that makes revocation transitive by construction, and `authorize()` as a distinct resource guard are all the right shape for Problem 04. The test suite has real teeth too, `test_revoking_mid_invalidates_leaf_not_root` (siblings survive, descendants die) and the depth-tampering and TTL-boundary cases in `TestBoundaryAndInvariants` are exactly the kind of adversarial coverage the rubric rewards.

**1. The branch has conflicts with main, and CI never ran.**

GitHub reports this PR as CONFLICTING against `main`, and no checks are reported on the branch. `nest_core/validators.py` is the file every submission touches, so conflicts there are expected; please rebase onto current `main`. The judge panel does not score broken PRs, so green checks on the actual branch matter more than the local `make ci-local` run described in the PR body.

**2. Duplicate validator definitions in `validators.py`.**

`validate_delegation_confused_deputy` and `validate_delegation_epoch_fence` are each defined twice (once in the new "Delegation Chain validators" block, again after the PARC section). Python silently keeps the second definition, and I would expect `ruff check` to flag the first pair as F811, which is hard to square with the "zero linting errors" claim. Also, the first block is inserted between the dashes and the title of the "Failure-detection validators" banner, splitting that header. Dedupe to one block and restore the banner.

**3. Determinism: `secrets.token_hex(16)` and wall-clock time in the simulation path (correctness).**

`issue()` and `delegate()` mint nonces with `secrets.token_hex(16)`, and `_now()` falls back to `time.time()`. Your unit tests pin `clock=`, but `delegation_chain_factory` constructs the plugin as `auth_cls(secret=b"delegation-chain-secret")` with no clock, and the tokens (with random nonces and wall-clock `iat`/`exp`) travel through `ctx.send`, so they land in the trace. `scenarios/delegated_auth.yaml` declares `tier: 1` with `seed: 42`, and Tier 1 requires byte-identical traces per seed. Suggested fix: derive nonces deterministically (e.g. HMAC of parent signature, depth, and audience, or a `random.Random` seeded from the scenario seed) and drive the clock from sim time rather than `time.time()`.

**4. One problem per participant: which PR is your primary?**

The charter asks for exactly one problem and a single PR, and you also have #86 open. Please say which one is your submission, or explain how they relate. Related: the branch is `hackathon/nandha-datafacts-verifier`, but this PR is auth-layer capability delegation, not a datafacts verifier, and the charter's pattern is `hackathon/<handle>-<short-theme>` (mismatched branch names are subject to auto-close). If this is the primary, rename to something like `hackathon/anilchowdary07-delegatable-auth` and update the "Author: nandha-datafacts-verifier" byline in `DELEGATABLE_AUTH.md` to match.

**5. Make the epoch fence, your differentiator, first-class (novelty, test rigor).**

Roughly twenty open PRs implement Problem 04, so the epoch fence and resource binding are what set this apart. Right now the scenario exercises the fence by reaching into privates (`auth._stale_after = 2` inside `CoordinatorAgent.on_message`), and `stale_after` defaults to `300.0` (reads as seconds) while `verify()` compares it against an epoch count delta. Configure it through the factory, pick one unit and document it, and property-test the fence. Two smaller items in the same vein: compare the final signature with `hmac.compare_digest` as `jwt_auth.py` does rather than `!=`, and confirm the spec's fails-on-baseline requirement by actually running the scenario with `auth: jwt`. Since `JwtAuth` has no `delegate`, the coordinator's `on_start` will raise `AttributeError`; check that this surfaces as failing validators rather than a crashed run.

**Suggested path forward:**

1. Rebase onto `main`, resolve the `validators.py` conflicts, and dedupe the two repeated validator functions.
2. Replace `secrets.token_hex` and the `time.time()` fallback in the sim path with seed-derived equivalents; confirm two runs at `seed: 42` produce identical traces.
3. Reply here stating whether #83 or #86 is your primary submission, and align the branch name and doc byline with this problem.
4. Verify the delegation validators fail against `auth: jwt`, then re-run `make ci-local` and confirm checks go green on the branch.

Leaving this open. Ping me on the hackathon channel if the determinism reshaping is unclear.

anilchowdary07:
Hi @dhve, thank you for the great feedback! I have pushed all the requested fixes: the duplicate validators are removed, the determinism reshaping (using HMAC nonces and the simulation clock) is complete, and make ci-local is now fully green and passing without conflicts. I also updated the main PR description at the top to explicitly frame our Epoch Fencing architecture against the recently merged PR #138, as I know novelty framing is important for the judges.

To answer your question: please consider this PR (#83 Delegatable Capability Tokens) as my primary submission for the hackathon scoring.

I also built the Tor-style Onion Routing in PR #86 because I thought it would be a really cool and helpful addition to the NandaTown ecosystem for privacy comms! It is fully rebased and passing CI as well, so feel free to review and merge it whenever you have time, but please use PR #83 for my actual hackathon evaluation.

JamesCarnley:
Thank you, @anilchowdary07. The valuable discriminator is epoch fencing: after authority advances, every old descendant must fail even when its signature, scope, and TTL otherwise appear valid, with wrong resource/audience and invalid TTL checked independently. Current Town has no delegation chain or authority epoch. This overlaps the legacy delegation family and still carries review concerns around encoding, time, scenario wiring, and vacuous evidence, so I am closing it as a duplicate and retaining the stale-descendant case under the stronger future profile represented by #73. No contributor code was merged or executed during this review.

PR #88, opened 2026-07-07
[Hackathon] registry: DNS-style caching resolver with TTL, negative caching, and self-evicting agents

SwasthikaDev:
## Summary

Adds `resolver`, a **DNS-style caching resolver** registry plugin: per-record TTL, negative-result caching, and self-eviction of stale/dead agents. The reference `in_memory` registry never expires records, so a crashed agent stays discoverable forever; the resolver models real service discovery where records go stale. Ships with an adversarial validator suite, a deterministic discovery scenario, docs, and tests.

## What and where — exact file manifest

**New plugin** — `packages/nest-plugins-reference/nest_plugins_reference/registry/caching_resolver.py` (186 lines)
- `CachingResolverRegistry` (**line 47**): registry with per-record TTL, negative caching (remembers "not found" for a bounded window), and stale eviction.
- `set_clock(now)` (**line 70**): pins the virtual clock the simulator advances. The clock defaults to `0.0` and `_now()` returns it directly — there is no `time.time()` fallback and no `time` import, so TTL expiry is driven only by `set_clock`. Records never expire until the clock is advanced, which keeps traces byte-identical under a fixed seed.
- `live_agents()` (**line 159**): the set of currently non-expired agents.

**Registration (2 lines)**
- `packages/nest-core/nest_core/plugins.py` — `("registry", "resolver")` builtin.
- `packages/nest-plugins-reference/pyproject.toml` — `nest.plugins.registry` entry point.

**Adversarial validators** — `packages/nest-plugins-reference/nest_plugins_reference/validators/resolver_validators.py`
- `check_stale_records_expire`, `check_heartbeat_keeps_alive`, `check_negative_cache_is_consistent`, and `run_all_resolver_checks`. Each **fails** against the reference `in_memory` registry (records never expire) and **passes** against the resolver. Exported from `validators/__init__.py`.

**Scenario**
- `packages/nest-core/nest_core/scenarios_builtin/discovery_resolver.py` + registration branch in `scenarios.py`.
- `scenarios/discovery_resolver.yaml`.

**Tests**
- `packages/nest-plugins-reference/tests/test_caching_resolver.py` — TTL expiry, negative caching, eviction, determinism.
- `packages/nest-plugins-reference/tests/test_discovery_resolver_scenario.py` — end-to-end scenario, deterministic across seeds.
- `packages/nest-plugins-reference/tests/test_resolver_benchmark.py` — the adversarial in_memory-fails / resolver-passes contrast.

## Maintainer review addressed

Removed the wall-clock fallback per review: `clock` now defaults to `0.0` (was `Optional[float] = None`), the `time` import is dropped, and `_now()` returns the pinned clock with no `time.time()` path. The plugin stays Tier-1 deterministic under any scenario selecting `registry: resolver`.

`uv run ruff check .` clean · `uv run ruff format --check .` clean · `uv run pyright` strict **0 errors** · `uv run pytest` **999 passed**, 1 skipped.


SwasthikaDev:
Update: added a property-surfacing benchmark and edge-case tests.

`tests/test_resolver_benchmark.py` churns 2,000 agents through both registries, four in five of which "crash" (register once, never heartbeat), then advances time and asks who is still resolvable:

- `in_memory` returns all 2,000 (including the 1,600 crashed ones) and holds them all in memory.
- `resolver` returns only the 400 still alive and prunes the rest, so memory stays bounded to the live set.

The gap (1,600) is the number of dead agents the default registry would advertise as reachable. Also added edge cases: re-register after expiry revives, deregister of an unknown agent is a no-op, and the expiry boundary is exclusive. ruff, ruff format, pyright (strict), and pytest all still pass.

Skyrider3:
Thanks,  reviewed and verified locally, and this is a strong submission: all five CI commands green at your head (810 passed, matching your PR body), green again on a local merge with current main (967 passed), nest run scenarios/discovery_resolver.yaml --seed 42 byte-identical across runs, and the adversarial contrast is real, your benchmark pins in_memory serving 1,600 dead agents out of a 2,000-agent fleet while the resolver serves exactly the 400 live ones, and the validators fail closed against in_memory ("registry has no clock"). The instance-level validator style matches the merged gossip registry's convention.

Two changes before merge:

1. Rebase/merge main. One conflict, in packages/nest-plugins-reference/pyproject.toml, against the trust entry-points that landed with #89. It's a pure both-added union  keep your nest.plugins.registry table and main's nest.plugins.trust table. I ran exactly that resolution locally and the full suite is green.

2. Remove the wall-clock fallback. CachingResolverRegistry.__init__ defaults clock=None, and _now() then returns time.time() (caching_resolver.py:83). Your scenario pins clock=0.0, so the shipped demo is deterministic — but the plugin is entry-point-registered precisely so other scenarios can select registry: resolver from YAML, and any that do will get wall-clock TTL expiry: latent non-determinism, which Tier 1 prohibits. Your PR body's "no wall-clock reads" is false as written, and the clock=None path is the one path none of your 13 unit tests exercise. Suggested one-liner: default clock: float = 0.0 and drop the time import — records then simply never expire until something advances the clock, which is in_memory-equivalent, deterministic, and preserves every existing test. Please also correct the PR body claim.

Non-blocking, take it or leave it: if the factory instantiated plugins["registry"] with a kwargs-fallback (the way the budget-enforcement driver does) instead of hardcoding the import, your scenario would also discriminate under a YAML swap to registry: in_memory ,  a nice extra proof for the leaderboard. Precedent doesn't require it (the gossip driver hardcodes too).

Outside this PR: you now have two open problem submissions (#88 registry, #94 memory). The charter is one problem per participant , since you consolidated onto #88 once already, please pick which of the two is your entry and close the other.

Push those and I'll merge.

SwasthikaDev:
Thanks for the review. Done:

- Removed the wall-clock fallback. `clock` now defaults to `0.0`, the `time` import is gone, and `_now()` returns the pinned clock directly. Records never expire until `set_clock` advances the clock, so expiry is fully deterministic under a fixed seed.
- Corrected the PR body's clock claim to match the code.
- Merged current `main` and re-ran full CI on the merge: ruff check/format clean, pyright strict 0 errors, pytest 999 passed / 1 skipped.

This (#88) is my entry; I'm closing #94 to consolidate. Ready to merge when you are.

SwasthikaDev:
For the one-problem-per-participant charter: this registry resolver is my **primary submission**. My other open/merged PRs (#95 metrics fix, merged #96 validator fix) are small infra corrections alongside it, not competing entries.

JamesCarnley:
Thank you, @SwasthikaDev. The useful future resolver contract is clear: injected logical time, exact expiry, heartbeat renewal, registration invalidating negative cache, and re-registration after expiry, with provider failure represented as unknown rather than absence. Current Town's in-run index verifies signed cards and ranks matches, but it makes no TTL, heartbeat, negative-cache, persistent-liveness, or external-registry claim. Adding those rules now would choose discovery and liveness policy and overlap the separate registry workstream, not repair the local index. I am closing the legacy PR while retaining those expiry and revival cases for a future resolver profile. No contributor code or provider call was executed during this review.

PR #89, opened 2026-07-07
[Hackathon] stellarminds-ai: AAE permit gate — pre-action authorization with signed denial receipts

Sharathvc23:
**Layer:** trust · **Persona:** stellarminds-ai

## Motivation

Every trust plugin in the town — the reference `score_average`, EigenTrust variants, receipt-corroborated reputation, all of it — scores what an agent **did**. None of them decide what an agent **may do**. That gap has a name: a well-reputed agent attempting an unauthorized action sails through every post-hoc reputation layer, because reputation is measured *after* the fact and the unauthorized action is exactly the one no prior good behavior should license.

Capability-token work (Problem 04) governs who may *hold* an authority. This plugin governs each *exercise* of it: it evaluates an action before it runs, and — crucially — a denial is a first-class, signed, verifiable receipt rather than a silent absence.

## Design

`aae_permit_gate` issues a **pre-action envelope** for every evaluation. Eight fields: `agent_id`, `action{verb,resource,params}`, `policy_id`, `outcome` (`authorized`|`denied`|`conditional`), `prev_hash` (per-agent causal chain), `issued_at`, `sig` (Ed25519 over the canonical envelope minus `sig`), `pubkey`.

- **Evaluate before acting.** The engine does not intercept actions, so gating is a scenario contract: an agent calls `evaluate(...)` and proceeds only on an authorized envelope. Plugins without an `evaluate` method are consulted as scores, so scenarios stay drop-in (`hasattr` capability-gate — the house pattern).
- **Deny by default.** Unmatched actions are denied; the posture is explicit and documented.
- **Signed denials.** A denied action does not execute, and the denial envelope *is* the receipt — anyone can verify the town refused it, under which policy, when.
- **Per-agent chain.** Each envelope's `prev_hash` points at the agent's previous one; `order_chain` recovers the unique order and rejects gaps, forks, deleted interior links, and foreign-agent splices.
- Self-contained: stdlib + `cryptography` only. Deterministic: every timestamp is supplied by the caller's virtual clock; same seed → byte-identical envelopes.

Registered both ways — `_BUILTINS[("trust","aae_permit_gate")]` and the `nest.plugins.trust` entry point in `nest-plugins-reference/pyproject.toml`.

## Adversarial validator (mandatory)

`rogue_trusted_agent_blocked` catches a high-reputation agent reaching outside its authorization. In the `rogue_trusted_agent` scenario one agent earns a strong reputation through eight in-policy actions, then attempts `spend town/treasury`:

```
exec:veteran:read:town/events                       <- reputation accrued
rogue_attempt:veteran:spend:town/treasury
permit:veteran:spend:town/treasury:denied:84773277  <- signed denial
blocked:veteran:spend:town/treasury                 <- never executes
```

**FAILS against `trust: score_average`** — reputation buys the treasury, the rogue action executes:

```
rogue_attempt:veteran:spend:town/treasury
exec:veteran:spend:town/treasury                    <- rogue action executes
```

**PASSES against `trust: aae_permit_gate`** — refused before it runs. No crash on either trace. A second validator, `rogue_trusted_agent_reputation`, corroborates that the veteran genuinely accrued reputation first and that no in-policy action was ever spuriously denied — so "reputation is irrelevant to the verdict" is grounded in the trace, not asserted.

## How to verify

```
make ci-local                                              # all 5 green
uv run nest run scenarios/rogue_trusted_agent.yaml -o /tmp/b.jsonl
grep -E 'permit:veteran|blocked:veteran' /tmp/b.jsonl      # the signed denial + block
# swap trust: aae_permit_gate -> score_average and rerun:
grep -E 'exec:veteran:spend' /tmp/b.jsonl                  # the rogue action executes
uv run nest plugins list                                   # aae_permit_gate under trust
```

Local: **798 passed** (all 5 `ci-local` steps green). Coverage is example-based + a Hypothesis property battery (round-trip, single-field tamper across all eight fields + key/sig mutations, chain re-orderability with gap/fork/foreign-splice detection, policy first-match determinism with "no denied pair ever permits", score invariants) + named adversarial cases (replay out of sequence, foreign-chain splice, forged-key mid-chain). Byte-identical reruns at fixed seed; the block/execute invariant holds across a 5-seed integration bank.

## Interop with #32 (Agent Action Capsules)

@StevenMih's Agent Action Capsules give the town an anchored ledger of what agents did; this plugin adds the other bookend — signed decisions about what agents may do. The optional adapter (off by default, public `capsule_emit.emit()` API only, no-op when the package is absent) seals each permit into the capsule ledger under a `permit.granted` / `permit.denied` namespace, with the decision object as output — deliberately never the underlying verb, so a permit can never be misread as an executed action. A denial that is anchored is evidence that the gate worked. The two submissions strengthen each other and neither depends on the other to merge. (One friendly API note: an external-reference field on `emit()` would let a permit carry its envelope hash as first-class correlation metadata, so a later execution capsule and its permit read as one story.)

## Persona

StellarMinds.ai, extending its civic role. The Town Notary attests claims; receipt-corroborated reputation (#26) proves what agents *did*; Ed25519 key rotation (#25) keeps identities verifiable over time; and the permit gate governs what agents *may* do. Is, did, is-still, may — the four tenses of an accountable citizen, each a signed artifact.

## Tradeoffs

- **Hash-committed, not key-pinned.** The per-agent chain is bound by `envelope_hash` (which covers the signature and key), not by a registry mapping each agent to a fixed key. Interior tamper, gaps, forks, and foreign splices are all caught structurally; binding an agent to a *stable* key across its chain is the identity layer's job (`ed25519_rotating`), which this gate composes with rather than duplicates. Stated plainly so a reviewer knows exactly where the boundary sits.
- **`conditional` is minimal** — it records "authorized subject to the stated condition params" and is not itself permission; enforcing conditions is left to the caller.
- **`report` is storage-only** for now (evidence a future policy may reference), and **`stake` is a documented parity no-op** — this gate is authorization-based, not stake-based.
- The scenario contract (agents consult the gate) reflects a real property of the engine: there is no action interceptor, so no trust or auth plugin can *force* a denial. The validator proves the contract held over the deterministic trace — the same shape used by the merged receipt work.

### DIFF
--- a/docs/layers/trust.md
+++ b/docs/layers/trust.md
@@ -84,10 +84,14 @@ with a typed reason).
 
 Every other trust plugin scores what agents **did**. `aae_permit_gate`
 decides what an agent **may do**, before it does it — and signs the
-decision either way. A well-reputed agent attempting an unauthorized
-action sails through every post-hoc reputation layer; this one stops it,
-and the stop is not a silent absence but a signed, verifiable denial
-receipt.
+verdict either way. A well-reputed agent attempting an unauthorized
+action sails through every post-hoc reputation layer on its record
+alone; this gate answers the request before it runs, and a refusal it
+returns is a signed, first-class receipt — provable from the envelope
+chain by anyone, not a silent absence. The refusal does not consult
+reputation, so no amount of prior good standing outranks it; the
+`rogue_trusted_agent` scenario below turns that returned verdict into a
+blocked action and proves it.
 
 Source:
 [`trust/aae_permit_gate.py`](../../packages/nest-plugins-reference/nest_plugins_reference/trust/aae_permit_gate.py)
--- a/docs/layers/trust.md
+++ b/docs/layers/trust.md
@@ -80,6 +80,108 @@ Scenario: `parc_migration` (two trust domains in one run; forged,
 replayed, stale-key, inflated, and wash-ring credentials each denied
 with a typed reason).
 
+## `aae_permit_gate` — pre-action authorization with signed denial receipts
+
+Every other trust plugin scores what agents **did**. `aae_permit_gate`
+decides what an agent **may do**, before it does it — and signs the
+decision either way. A well-reputed agent attempting an unauthorized
+action sails through every post-hoc reputation layer; this one stops it,
+and the stop is not a silent absence but a signed, verifiable denial
+receipt.
+
+Source:
+[`trust/aae_permit_gate.py`](../../packages/nest-plugins-reference/nest_plugins_reference/trust/aae_permit_gate.py)
+· envelope format:
+[`trust/aae_envelope.py`](../../packages/nest-plugins-reference/nest_plugins_reference/trust/aae_envelope.py).
+
+**The permit envelope.** Before an action runs, the gate evaluates it
+against a declarative policy table and issues a signed JSON object — a
+*pre-action envelope* — recording the verdict. Eight fields, no more:
+
+| Field | Meaning |
+|---|---|
+| `agent_id` | The acting agent's town identity (whatever string the identity layer yields). |
+| `action` | `{verb, resource, params}` — what the agent intends to do. |
+| `policy_id` | Which policy rule the action was evaluated against. |
+| `outcome` | `"authorized"` \| `"denied"` \| `"conditional"` — denials are first-class. |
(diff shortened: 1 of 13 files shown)


shilpa-kulkarni-14:
LGTM

PR #91, opened 2026-07-07
[Hackathon] theodornengoy (security-engineer): strict holder-derived capability delegation

TheodorNEngoy:
## Problem

Solves problem 04, auth capability delegation, as a distinct implementation
that coexists with the baseline `delegatable` plugin merged in #138.

The reference `jwt` plugin cannot model a coordinator attenuating its
capability through intermediaries to leaf workers while retaining cascading
revocation. The new plugin preserves the base `Auth` protocol and adds an
explicit audience-aware verification API for that workflow.

## Differentiation from #138

This revision uses separate module, plugin, scenario, validator, and test names
so both implementations remain runnable and scoreable:

- `StrictDelegatableAuth`, registered as `auth: delegatable_strict`;
- `scenarios/strict_delegated_auth.yaml` and task type
  `strict_delegated_auth`;
- `check_strict_delegated_auth_attack_suite`, which exercises only public auth
  APIs and records the exception type for every rejected attack.

The implementation also makes three different choices load-bearing:

- child scopes must be a non-empty **strict** subset; equal scopes are rejected;
- a child-signing key is derived from the parent token's signature and hash,
  and a test rejects issuer-secret child re-signing;
- the validator crafts the direct parent-signature key-confusion forgery shape
  that the merged suite does not construct.

## What changed

- Added `delegate(parent_token, audience, scopes_subset, ttl)` without changing
  the base `Auth.verify(token)` protocol signature.
- Added `verify_for(token, presenter)` for explicit audience enforcement.
- Added transitive ancestor verification, cascading revocation, and strict
  scope and expiry attenuation.
- Added domain-separated HMAC signing for roots, holder-derived child keys, and
  child payloads.
- Added the requested deterministic topology: 1 coordinator, 3 intermediaries,
  and 12 leaves.
- Added adversarial coverage for scope escalation, stale-parent use, audience
  confusion, and key-confusion token forgery.

## Security model and tradeoffs

This is intentionally an in-process Nanda Town rig, not a production token
introspection service. The instance retains an in-memory ancestor-token table
so the simulator can deterministically verify the full chain and prove
cascading revocation.

`verify(token)` follows the base protocol and therefore treats a valid child as
a bearer token. Callers that must bind the presenter to the child audience use
`verify_for(token, presenter)`. `AuthContext.subject` identifies that
holder/audience, while the token's `sub` claim retains the original delegator.

Invalid child TTL requests now raise `DelegationError`; `ExpiredAncestorError`
is reserved for a chain whose ancestor actually expired or is outlived.

## Verification

Rebased onto current `main` after #138 and ran:

```bash
uv run ruff check .
uv run ruff format --check .
uv run pyright
uv run pytest -q
uv run nest run scenarios/strict_delegated_auth.yaml
```

Results:

```text
ruff check: all checks passed
ruff format --check: 215 files already formatted
pyright: 0 errors, 0 warnings, 0 informations
pytest: 1168 passed, 1 skipped, 1 deselected, 1 warning
scenario: Running scenario: strict_delegated_auth
scenario: agents: 16 seed: 4242 ticks: 3000
scenario determinism: two fresh runs produced byte-identical 122-line traces
scenario SHA-256: fd9877021de5396c31a2de13f1f52231acc77f0254410571c29f633828443fce
merge simulation against current main: clean
```

The scenario trace contains all three required attack outcomes:

- `attack:scope_escalation:blocked`
- `attack:stale_parent:blocked`
- `attack:audience_confusion:blocked`

The focused strict-delegation suite has 18 tests, including key-confusion
forgery rejection, issuer-secret child-signature rejection, non-ASCII malformed
signature handling, typed rejection evidence, and a Hypothesis strict-subset
property.

### DIFF
--- a/docs/layers/auth.md
+++ b/docs/layers/auth.md
@@ -42,6 +42,29 @@ Source: [`nest_plugins_reference/auth/delegatable.py`](../../packages/nest-plugi
 Validators: [`nest_plugins_reference/validators/delegation_validators.py`](../../packages/nest-plugins-reference/nest_plugins_reference/validators/delegation_validators.py).
 Scenario: [`scenarios/delegated_auth.yaml`](../../scenarios/delegated_auth.yaml).
 
+## Strict holder-derived variant: `delegatable_strict`
+
+`delegatable_strict` coexists with the baseline `delegatable` plugin while
+making three additional choices load-bearing: every child must remove at
+least one parent scope, child-signing keys are derived from the parent token's
+signature and hash rather than the issuer secret, and the public-API validator
+crafts a direct parent-signature key-confusion forgery.
+
+The base protocol's `verify(token)` call authenticates a valid child as a
+bearer token. Call `verify_for(token, presenter)` when the caller must also
+enforce that the presenter matches the child's audience. `AuthContext.subject`
+therefore identifies the holder/audience; the token's `sub` claim retains the
+original delegator.
+
+The validator records the exception type for every rejected attack so a typed
+security rejection remains distinguishable from an unrelated plugin crash.
+The deterministic `strict_delegated_auth` scenario preserves the requested
+1 coordinator, 3 intermediary, and 12 leaf topology.
+
+Source: [`nest_plugins_reference/auth/strict_delegatable.py`](../../packages/nest-plugins-reference/nest_plugins_reference/auth/strict_delegatable.py).
+Validator: [`nest_plugins_reference/validators/strict_delegation_validators.py`](../../packages/nest-plugins-reference/nest_plugins_reference/validators/strict_delegation_validators.py).
+Scenario: [`scenarios/strict_delegated_auth.yaml`](../../scenarios/strict_delegated_auth.yaml).
+
 ## Writing your own
 
 See [`writing-a-plugin.md`](../writing-a-plugin.md). Register under
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -27,6 +27,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("registry", "gossip"): f"{_REF}.registry.gossip:GossipRegistry",
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
     ("auth", "delegatable"): f"{_REF}.auth.delegatable:DelegatableAuth",
+    ("auth", "delegatable_strict"): (f"{_REF}.auth.strict_delegatable:StrictDelegatableAuth"),
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
     ("trust", "agent_receipts"): f"{_REF}.trust.agent_receipts:AgentReceiptsTrust",
     ("trust", "parc"): f"{_REF}.trust.parc:ParcTrust",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -131,6 +131,12 @@ def _try_load_builtin(name: str) -> None:
         from nest_core.scenarios_builtin.delegated_auth import delegated_auth_factory
 
         register_scenario("delegated_auth", delegated_auth_factory)
+    elif name == "strict_delegated_auth":
+        from nest_core.scenarios_builtin.strict_delegated_auth import (
+            strict_delegated_auth_factory,
+        )
+
+        register_scenario("strict_delegated_auth", strict_delegated_auth_factory)
     elif name == "multi_attribute_market":
         from nest_core.scenarios_builtin.multi_attribute_market import (
             multi_attribute_market_factory,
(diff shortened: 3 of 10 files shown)


dhve:
Thanks @TheodorNEngoy. Of the roughly twenty Problem 04 submissions in the queue, this is one of the strongest reads. Two choices stand out: you kept `Auth.verify(token)` untouched and added `verify_for(token, presenter)` alongside it instead of breaking the protocol signature, and child keys are derived from the parent token's own signature through `_CHILD_KEY_DOMAIN` rather than re-signing with the issuer secret. `test_child_signature_is_derived_from_parent_token_not_issuer_secret` plus the key-confusion forgery test make that property load-bearing, and the fixed-clock injection in `delegated_auth_factory` keeps the sim path deterministic.

Hard news first, because the ground moved after you opened this. Not ready as-is; four things to work through:

**1. Main just merged a competing Problem 04 implementation at the same paths, so this needs a rename, not just a rebase.**

GitHub reports the branch as conflicting, and a merge simulation against current main shows why: #138 landed with the same `DelegatableAuth` class, the same `("auth", "delegatable")` registry key, and the same file paths. `auth/delegatable.py`, `scenarios_builtin/delegated_auth.py`, `tests/test_delegatable_auth.py`, and `scenarios/delegated_auth.yaml` all hit add/add conflicts, plus content conflicts in `docs/layers/auth.md`, `nest_core/scenarios.py`, `validators/__init__.py`, and the plugins-reference `pyproject.toml`. To coexist you need a distinct plugin key, module, scenario name, and file names. Novelty is scored against merged peers, so state plainly in the PR body what this adds beyond it. From my read the real differentiators are the key-confusion forgery probe and the issuer-secret child-signature rejection (the merged suite tests tampered payloads but never crafts that forged-key token shape), strict-subset-only scopes (the merged plugin allows equal scopes), and a validator that exercises an auth plugin's public API rather than audit events. Make those load-bearing in the body. Also, no CI checks have reported on this branch yet; the judge panel does not score broken PRs, so confirm the pipeline goes green here after the rework.

**2. The validator counts any exception as a blocked attack, and it probes a private method.**

In `check_delegated_auth_attack_suite`, every attack arm is `except Exception: pass`. A third-party plugin whose `verify` crashes with an unrelated `AttributeError` would read as "attack blocked". Keep the broad catch, but record the exception type in `evidence` so a crash is distinguishable from a typed rejection. Separately, the `getattr(auth, "_parse", None)` probe tests private surface: an implementation whose `_parse` decodes without verifying would be flagged even though its public `verify` rejects the forged token. The `await verify(forged)` check right below it is the meaningful one; drop the `_parse` probe. Also prefix `materialize_report`; it is too generic a name to export from the shared `validators` namespace.

**3. Audience binding is opt-in; the docs should say so.**

Plain `verify(child)` succeeds for any presenter; the audience check only fires through `verify_for`. That is a defensible tradeoff given the base protocol, but one sentence in `docs/layers/auth.md` and the PR body's security model should state that any caller still using bare `verify()` treats the child as a bearer token. Related nit: `verify()` returns `AuthContext(subject=claims.audience, ...)` while the token's `sub` claim carries the delegator; a short comment where the `AuthContext` is built would save the next reader who compares against `jwt_auth` semantics.

**4. `ExpiredAncestorError` for an over-long child TTL is a misleading type.**

In `delegate()`, `ttl > remaining` raises `ExpiredAncestorError`, but nothing is expired at that point; the request is invalid. Plain `DelegationError` (or a dedicated type) keeps the otherwise clean typed-exception story consistent.

**Suggested path forward:**

1. Pick a distinct plugin key, module, scenario name, and file names, rebase onto latest main, and resolve the remaining content conflicts.
2. Add a differentiation paragraph to the PR body aimed at the merged #138.
3. Apply the validator changes from point 2 and the doc and exception-type cleanups from points 3 and 4.
4. Re-run the full local pipeline, re-verify byte-identical traces across two fresh scenario runs, and confirm CI reports green on the PR.

Leaving this open. The collision is bad timing, not bad work. Ping me on the hackathon channel if you want to talk through the renames.

TheodorNEngoy:
Thanks for the detailed review. I rebased onto current main and moved the
submission to the distinct `delegatable_strict` plugin/scenario family so it
coexists with #138. I also removed the private `_parse` validator probe,
recorded rejection exception types in validator evidence, documented bare
`verify()` versus audience-bound `verify_for()`, clarified the
`AuthContext.subject` semantics, and corrected the over-long child TTL error
type. The full local pipeline is green (1168 passed, pyright clean), two fresh
scenario traces are byte-identical, and a merge simulation against current
main is clean.

JamesCarnley:
Thank you, @TheodorNEngoy. This is strong requirements evidence for a future holder-derived delegation profile: a parent holder must create a valid child without the issuer secret, issuer-secret and key-confusion forgeries must fail, and bearer verification must remain explicitly distinct from holder-bound presentation. Current Town uses flat controller-to-session Run Grants and does not support child delegation. Adopting this mechanism would choose new authorization and revocation policy rather than fix the selected one-hop contract, so I am closing the legacy PR while retaining its holder issuance, key-confusion, presenter, time, and bounded-chain cases for future work. No contributor code was merged or executed during this review.

PR #92, opened 2026-07-07
feat(auth): Macaroon-based delegatable tokens with cascading revocation 

AdityaDarke97:
## Overview
This PR solves **Problem #04** by adding the `macaroons` auth plugin. 

Before, if an orchestrator agent wanted to give a sub-agent narrow, temporary access to a tool, it had to request a brand new token from the root issuer. Now, agents can delegate capabilities themselves. This allows for parent-child token relationships and cascading revocation.

## How it works (HMAC Chaining)
Based on the 2014 Macaroons paper, this plugin uses HMAC signature chaining rather than tracking every individual token.

* **Delegation:** The `delegate(parent_token, audience, scopes_subset, ttl)` method lets an agent mint a child token. The child's signature is generated using the parent's signature as the key (`new_sig = HMAC(parent_sig, child_claims)`).
* **Cascading Revocation:** If a parent token is revoked, its signature becomes invalid. Because all child tokens rely on the parent's signature to verify, this breaks the chain for all descendants. The entire sub-tree is revoked instantly without needing to track every single child token.
* **Constraints:** The plugin ensures a child token's scopes are a strict subset of the parent's, and the child's expiration time cannot exceed the parent's.

## Adversarial Validation
To make sure this is secure, I added a validator in `test_scenario.py` that tests three specific attacks:
1. **Scope Escalation:** Trying to mint a child token with more permissions than the parent.
2. **Stale Parent:** Trying to use a child token after its parent was revoked.
3. **Audience Confusion:** Presenting a token to an agent other than its intended audience.

The tests show that the default `jwt` plugin is vulnerable to these delegation-specific attacks, while the new `macaroons` plugin blocks all of them. *(Note: I used `inspect.signature` in the test setup so it handles the different argument footprints of both plugins safely).*

## What's Included
- `MacaroonAuth` plugin in `nest_plugins_reference/auth/macaroons.py`.
- Registered `("auth", "macaroons")` in `nest_core/plugins.py`.
- `scenarios/delegated_auth.yaml` featuring a 16-agent delegation tree (coordinator, 3 intermediaries, 12 leaves) to test it end-to-end.
- Attack validation tests in `test_scenario.py`.

## Testing
This is a standard in-memory, deterministic plugin with no external dependencies. It respects the existing `Auth` protocol.

Ran the local suite:
```text
uv run ruff check .
uv run ruff format --check .
uv run pytest -v

Result: 778 passed, 1 skipped, 1 deselected


dhve:
Thanks @AdityaDarke97. This is the right shape for Problem 04: macaroon-style HMAC chaining in `delegate()` (`new_sig = HMAC(parent_sig, child_claims)`), so a parent can mint children offline, and revocation cascades by construction since `verify()` walks every layer and checks each `jti` against `_revoked`. The typed exceptions (`RevokedAncestorError`, `ScopeEscalationError`, `AudienceConfusionError`) are exactly what the problem asked for.

**1. `verify()` never enforces scope attenuation, so scope escalation is only blocked at the polite API.**

`delegate()` checks the subset rule, but the parent's signature is embedded in the token itself. Anyone holding a parent token can compute `HMAC(parent_sig, forged_claims)` and append a layer with `"scopes": ["admin"]`. `verify()` checks the signature chain, per-layer expiry, and revocation, but never checks that each layer narrows its ancestor's scopes. That breaks the core macaroon invariant: caveats can only attenuate. Fix: walk the chain in `verify()` carrying the allowed scope set, reject any layer that widens it, and add an adversarial case that forges a layer directly instead of calling `delegate()`.

**2. `MacaroonAuth` does not conform to the `Auth` protocol.**

The PR body says it respects the existing protocol, but `nest_core/layers/auth.py` defines `async def issue(subject, scopes) -> Token`, `async def verify(token) -> AuthContext`, `async def revoke(token) -> None`. Yours are all synchronous, `issue` takes a required `ttl`, and `verify` takes a required `audience` and returns `dict[str, Any]`. Since `("auth", "macaroons")` is in `_BUILTINS`, anything that resolves the layer and does `await auth.issue(...)` breaks. Make the three base methods async with the protocol's signatures (extra params as optional kwargs), return `Token` and `AuthContext`, and keep `delegate()` as the additive surface.

**3. The JWT half of the validator never actually runs the JWT plugin.**

`JwtAuth.issue/verify/revoke` are coroutines, and `validate_delegation_security` calls them without awaiting. So `parent_token` is a coroutine object, `_verify` returns another un-awaited coroutine and never raises, and the revoke never executes. All three attacks get recorded as exploited without any JWT verification logic running, so `test_jwt_fails_delegation_security` passes vacuously instead of proving the baseline is genuinely vulnerable. Make the validator async and drive both plugins through the same awaited path; that also removes the `inspect.signature` branching.

**4. `scenarios/delegated_auth.yaml` cannot run.**

`task.type: delegation_tree` has no registered factory, so `get_scenario_factory("delegation_tree")` raises `KeyError`, and no test loads the YAML. Follow `test_empic_payments_partition_yaml` in the same file: `ScenarioConfig.from_yaml(...)`, `ScenarioRunner(config).run()`, validate the trace. Register a small `delegation_tree` factory exercising issue, delegate, verify, and revoke across the 16-agent tree.

**5. Determinism and lint red flags.**

`time.time()` and `uuid.uuid4()` in the token path are charter red flags for Tier 1 (same seed, byte-identical trace). `JwtAuth` shows the pattern with its injectable `clock`; derive `jti` deterministically too. Also, none of the four files ends with a newline, and `ruff format --check` fails the three Python ones for it (the only change to `plugins.py` besides your registration line is deleting the final newline). Add `from __future__ import annotations` and `Example::` docstring blocks to `macaroons.py` while you are in there.

Logistics: the branch is `feat/macaroon-auth-delegation`, but the charter asks for `hackathon/<handle>-<short-theme>` and mismatched names are subject to auto-close; no CI checks have reported yet. You also have #146 open on a different problem, and the charter is one problem per participant, single PR, so please pick your primary. Roughly twenty open PRs target Problem 04: state your differentiator in the PR body (offline HMAC chaining is a real one once item 1 is fixed) and make it load-bearing in tests.

**Suggested path forward:**

1. Enforce scope attenuation in `verify()` and add the forged-layer test.
2. Reshape to the async `Auth` protocol, keeping `delegate()` additive.
3. Rewrite the validator async so the JWT baseline actually executes.
4. Register a `delegation_tree` factory and add a runner-based YAML test.
5. Fix determinism and formatting, then run `ruff check`, `ruff format --check`, `pyright`, `pytest -v` before re-requesting review.

Leaving this open. Ping me on the hackathon channel if the protocol reshaping is unclear.

JamesCarnley:
Thank you, @AdityaDarke97. The important security invariant is verifier-side attenuation: a holder-computed, MAC-valid widened child must fail, and the same real interface must exercise both the subject and baseline. At this head, scope is not re-enforced during verification, the synchronous methods do not conform to the legacy async Auth protocol, the baseline is not awaited, the YAML has no runnable factory, and wall time/UUIDs break determinism. Current Town's flat grant API is entirely different and already tests its selected one-hop contract. I am closing this obsolete implementation while retaining widening, expiry, revocation, presentation, and real-baseline cases for a future delegation profile. No contributor code was merged or executed during this review.

PR #93, opened 2026-07-08
hackathon/superheroin888: delivery-gated streaming payments with mid-stream cancellation (problem 03)

superheroin888:
## What this is

**Problem 03 — streaming pay-per-second payments with mid-stream cancellation** (`payments` layer). One problem, one layer.

The tree already contained a `streaming` plugin, three validators, and two scenario YAMLs — but they were scaffolding, not a solution:

- `nest run scenarios/streaming_payments.yaml` **crashed**: no `streaming_payments` scenario factory existed.
- The plugin hardcoded `opened_at_tick=0` ("will be set by context"), stamped `closed_at_tick=0` on close, had zero partition awareness, and named the wrong payer on a payee-side close.
- All three validators checked event shapes (`payment_debited`, `event_type: stream_opened`) that the trace layer **never emits** — they passed vacuously on any trace, including an empty one.

## What ships

**Delivery-gated billing** as the core design: the payer records each delivered work unit (`record_delivery`) and only then drains one tick (`tick_stream`). A unit the payee never delivered — dropped message, network partition, or a closed stream — is never billed. This is the per-request metering shape of x402-style HTTP payments on the simulator's logical clock.

- **Plugin** (`("payments", "streaming")`, drop-in for the stock `Payments` protocol): `open_stream`/`close_stream` per the brief, close idempotent and final at any tick (unused remainder never spent), `pay()` literally implemented as a one-tick stream so one-shot and streaming share every invariant, refund only on settled refs. `verify_payment` returns `STREAMING` for a half-drained stream — drained ticks are final and irrevocable, so the settlement is in progress, not `PENDING` (documented in the docstring, as the brief asks). Deterministic: ticks come from the caller; no wall clock, no RNG.
- **Scenario factory** (was missing): 5 buyer/seller pairs, rolling streams, per-unit ack → per-unit debit, mid-stream cancellation on every third stream, cancel deadlines so the run makes progress under drop. Both existing YAMLs now run and pass.
- **Validators rewritten against the real trace format and made non-vacuous** (they FAIL on a trace with no streaming lifecycle — a plugin that quietly pre-pays cannot pass):
  1. *conservation* — settled total == sum of per-unit debits, never past `max_total`;
  2. *no-drain-after-close* — trace-order check, plus no unit billed twice;
  3. *no-overbill-on-partition* — every debit must cite a previously **delivered** ack (`receive` event), which subsumes both random drop and full partition.

## Adversarial discrimination

- Under `payments: streaming`: all three validators PASS (5% drop run: 97 streams, 297 debits, every settled receipt equals its debit sum).
- Under `payments: prepaid_credits` on the identical scenario: all three FAIL ("no streaming lifecycle observed").
- Under a full buyer/seller partition: 96 streams open, **zero** credits move, validators pass — not billing for undelivered work is the correct behavior.
- 8 synthetic attack traces (drain-after-close, double-bill, over-bill without ack, time-travel billing, receipt mismatch, past-cap billing, missing lifecycle) each caught by the right validator.

## Tests

26 plugin unit tests · 6 hypothesis properties (conservation under random op sequences — adapted from the escrow suite as the brief suggests — cap enforcement, close finality) · 6 end-to-end scenario tests (discrimination, byte-identical replay at seed 42, multi-seed, partition ⇒ zero flow) · 6 validator smoke tests.

## Reproduce

From a clone of this PR branch:

```bash
gh pr checkout 93   # or: git fetch origin pull/93/head && git checkout FETCH_HEAD
make ci-local       # the repo's Definition of Done: uv sync, ruff check,
                    # ruff format --check, pyright, pytest -v
```

Expected: `ci-local: all 5 checks passed. Safe to push.` (804 passed, 1 skipped.)

Then the behavioral evidence, straight from the scenario YAMLs:

```bash
# 1. 5% drop run — all three validators PASS
uv run nest run scenarios/streaming_payments.yaml -o ./traces/sp.jsonl
[print(('PASS' if r.passed else 'FAIL'), r.name, '-', r.detail) for r in validate_trace(Path('traces/sp.jsonl'),'streaming_payments')]"
# expected: 97 streams, 297 debits, every settled receipt equals its debit sum

# 2. full buyer/seller partition — streams open, ZERO credits move, validators PASS
uv run nest run scenarios/streaming_payments_partition.yaml -o ./traces/spp.jsonl
[print(('PASS' if r.passed else 'FAIL'), r.name, '-', r.detail) for r in validate_trace(Path('traces/spp.jsonl'),'streaming_payments')]"
# expected: '96 streams, 0 total flow' — partitioned payers bill nothing

# 3. discrimination — flip payments: streaming -> prepaid_credits in
#    scenarios/streaming_payments.yaml, re-run step 1: all three validators FAIL
#    ('no streaming lifecycle observed')

# 4. determinism — re-run step 1 twice; traces are byte-identical (seed 42)
```

CI state on this branch (on top of main @ 5527bd7): ruff ✓ · format ✓ · pyright (strict) ✓ · **804 tests passed** · `make ci-local` exits green.

### DIFF
--- a/docs/layers/payments.md
+++ b/docs/layers/payments.md
@@ -23,7 +23,21 @@ Source: [`nest_plugins_reference/payments/prepaid_credits.py`](../../packages/ne
 
 ## Additional reference plugins
 
-`streaming` — bilateral per-tick streams with mid-stream cancellation.
+`streaming` — bilateral per-tick streams with mid-stream cancellation
+and **delivery-gated billing**: the payer records each delivered work
+unit (`record_delivery`) and only then drains one tick (`tick_stream`),
+so dropped or partitioned payees are never billed for work they could
+not deliver — the per-request metering shape of x402-style HTTP
+payments. `open_stream`/`close_stream` bound the flow by `max_total`;
+closing settles exactly what was billed and the remainder is never
+spent. One-shot `pay()` is a stream that drains everything in one tick.
+The `streaming_payments` scenario (and its `_partition` variant)
+exercises it under 5% message drop and a full buyer/seller partition;
+three adversarial validators check conservation, drain-after-close, and
+over-bill-on-partition, and all three fail loudly under a plugin that
+quietly pre-pays instead of streaming.
+
+Source: [`nest_plugins_reference/payments/streaming.py`](../../packages/nest-plugins-reference/nest_plugins_reference/payments/streaming.py).
 
 `empic_escrow` — EMPIC-shaped escrow for service providers and
 consumers. Pull mode locks one request payment until accepted data is
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -123,3 +123,9 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("escrow_marketplace", escrow_marketplace_factory)
+    elif name == "streaming_payments":
+        from nest_core.scenarios_builtin.streaming_payments import (
+            streaming_payments_factory,
+        )
+
+        register_scenario("streaming_payments", streaming_payments_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/streaming_payments.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/streaming_payments.py
@@ -1,0 +1,362 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Streaming payments scenario -- five buyer/seller pairs with rolling streams.
+
+Each buyer opens a sequence of payment streams to its seller and bills one
+work unit at a time, **only after the seller's ack for that unit arrives**.
+Every billing-relevant transition is broadcast as a structured
+``stream:<kind>:k=v`` payload that the three ``streaming_payments``
+validators (see :mod:`nest_core.validators`) read back from the trace:
+
+* ``stream:opened`` -- ref, payer, payee, rate, max, tick;
+* ``stream:debit``  -- one per billed unit, citing the acked unit number;
+* ``stream:closed`` -- ref, settled total, tick, closing party, reason.
+
+The seller's per-unit ack travels as a *direct* message
+(``stream:ack:ref=..:unit=..``), so the trace shows exactly which acks were
+delivered (``receive``) and which the failure injector killed (``dropped``).
+A debit without a delivered ack is an over-bill — that is what the
+partition validator hunts.
+
+If the payments plugin lacks the streaming surface (e.g. ``prepaid_credits``),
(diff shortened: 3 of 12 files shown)


Skyrider3:
**backward-incompatible semantics change to a merged registered plugin**
(streaming.py) tick_stream return changed bool → int with inverted meaning (was "still open", now "amount drained"), and open_stream no longer drains the first tick. Grep confirms zero external consumers today (only the registry references the class; no other YAML uses payments: streaming), so the blast radius is nil, but the break is undisclosed in the PR body and should be stated explicitly.

 **"conservation" validator overclaims (one-sided self-consistency)**
(validators.py ~1315-1360). billed[ref] comes from the amount= field of stream:debit broadcasts and is compared to the total= of stream:closed — both authored by the same payer-side driver. No payee-credit or balance event exists in the trace, so "money is neither created nor destroyed" is not what's checked; a plugin whose trace narration is self-consistent while its actual ledger moves different amounts is invisible. The genuinely adversarial tooth is the cap check (driver-set max=, independent of the plugin). Re-scope the docstring or emit/consume a balance-delta event.

**ack identity not bound to the engine record**
delivered_acks keys on the ack recipient (engine-attributed — good) but ignores ev["from"] (the supposed deliverer), and never cross-checks opened.payer/payee strings against engine agent fields. Not exploitable by a plugin-only adversary; closes cheaply by requiring ack from == opened[ref].payee and debit broadcaster == opened[ref].payer.

 **unguarded int() on author-controlled fields crashes validation**
int(fields.get("amount","0")) etc. have no try/except and validate_events doesn't isolate validator exceptions — one amount=abc line turns a should-be-FAIL into an unhandled crash of the whole validation run. Parse defensively; treat unparseable numerics as violations.

**verify_payment reports CONFIRMED for a refunded ref**
refund deletes the payment record but leaves the closed StreamHandle in _streams; verify_payment falls through to the stream branch → CONFIRMED (the pre-rewrite plugin returned FAILED). One-line fix: self._streams.pop(ref, None) in refund.

Skyrider3:
please sync your fork and resubmit it!

JamesCarnley:
Thank you, @superheroin888. The valuable future contract is precise: bill only uniquely delivered work units, never twice or after close/cancel, never beyond a cap, bind debit, acknowledgement, and receipt to the actual payer/payee/reference, and let refund change later verification state. Current Town has one-shot transfer and escrow only; adapted legacy streaming keys are explicitly marked not tested. The frozen plugin also has unresolved binding, parsing, compatibility, and refunded-state issues. I am closing this legacy PR as a future profile rather than importing unselected streaming settlement semantics, while retaining its zero-flow partition, cancellation-race, cap, and no-double-bill cases. No contributor code was merged or executed during this review.

PR #95, opened 2026-07-08
[Hackathon] metrics: fix delivery_rate/success_rate exceeding 1.0 on broadcast-heavy scenarios

SwasthikaDev:
## Summary

Fixes `delivery_rate` (and its `success_rate` alias) returning a value **greater than 1.0** on any scenario that uses broadcasts. A delivery "rate" above 1.0 is impossible and misreports every broadcast-heavy run.

## Where — exact location

- **File:** `packages/nest-core/nest_core/metrics.py`
- **Function:** `_delivery_rate` at **line 57**, surfaced as the `delivery_rate` metric and via the alias `_success_rate = _delivery_rate` at **line 92** (the `success_rate` metric name).

## Root cause

The simulator logs a one-to-one message as a `send` record (sender) plus a `receive` record (recipient). A **broadcast** is logged differently in `packages/nest-core/nest_core/sim/simulator.py`: one `broadcast` record on the sender and one `receive` record per recipient — **no `send` records at all**.

The old `_delivery_rate` used `send` events as the denominator and `receive` events as the numerator:

```python
# BEFORE  (metrics.py:57)
sends = 0
receives = 0
for ev in events:
    kind = ev.get("kind", "")
    if kind == "send":
        sends += 1
    elif kind == "receive":
        receives += 1
if sends == 0:
    return 0.0
return receives / sends
```

Every broadcast adds N to `receives` and 0 to `sends`, so the ratio climbs above 1.0.

## Reproduction — a shipped scenario

`scenarios_builtin/reputation.py`'s observer calls `ctx.broadcast(...)`:

```
$ uv run nest run reputation
# trace event kinds: {'send': 280, 'receive': 340, 'broadcast': 3, ...}
# delivery_rate = 340 / 280 = 1.2143   <-- impossible
```

## The fix

Count **delivery attempts** for the denominator. Every attempt the simulator makes ends as exactly one record — a `receive` (delivered) or a `dropped` (lost) — whether it came from a `send` or a `broadcast`:

```python
# AFTER  (metrics.py:57)
receives = 0
dropped = 0
for ev in events:
    kind = ev.get("kind", "")
    if kind == "receive":
        receives += 1
    elif kind == "dropped":
        dropped += 1
attempts = receives + dropped
if attempts == 0:
    return 0.0
return receives / attempts
```

**Why this is safe for existing traces:** for unicast, every `send` yields exactly one `receive` or one `dropped`, so `sends == receives + dropped` and the value is *identical*. It only changes the broadcast case, and it can never exceed 1.0. After the fix, `reputation` reports `1.0`.

## Tests

- **Unchanged, still pass:** `test_delivery_rate` (`test_metrics.py:35`) and `test_success_rate_backward_compat` (`test_metrics.py:41`) — a unicast trace with 3 sends / 2 receives / 1 dropped, still `2/3`.
- **Added:** `test_delivery_rate_with_broadcast_stays_within_one` at **`packages/nest-core/tests/test_metrics.py:48`** — a trace with two broadcasts (one fully delivered, one with a dropped fan-out). Asserts the rate is `<= 1.0` and equals `3/4` (3 delivered of 4 attempts). This test **fails** on the old `receives / sends` formula.

## Files changed (2 files, +19 / -7 and +26)

| File | +/- | What |
|---|---|---|
| `packages/nest-core/nest_core/metrics.py` | +19 / -7 | Rewrote `_delivery_rate` denominator (line 57) |
| `packages/nest-core/tests/test_metrics.py` | +26 | Broadcast regression test (line 48) |

`uv run ruff check .` clean · `uv run ruff format --check .` clean (172 files) · `uv run pyright` strict **0 errors** · `uv run pytest` **792 passed**, 1 skipped. Scoped to the one metric plus its test; no other metric or behavior is touched.


dhve:
Thanks @SwasthikaDev. Real bug, clean fix, and one of the best-written PR bodies this round: root cause, shipped-scenario repro, safety argument. I verified it against main: `broadcast()` in `sim/simulator.py` logs one `broadcast` record and fans out via `InMemoryTransport.broadcast()` into per-recipient `deliver` events, each resolving to one `receive` or `dropped`, so `receives / (receives + dropped)` is the right formula.

This is close. Three things to tighten before merge:

**1. CI never ran on this branch.**

No checks have run on this branch, while #88 and #96 both show lint, test, and typecheck runs. Local verification in the body helps, but the judge panel only scores PRs with green CI on GitHub. Push a rebase on current main (or an empty commit) to trigger the workflows and clear the UNKNOWN mergeability state GitHub currently reports.

**2. `test_success_rate_bounded` still documents the bug you just fixed.**

The hypothesis property at `packages/nest-core/tests/test_properties.py:288` asserts only `rate >= 0.0`, with a comment saying the rate "can exceed 1.0 if more receives than sends". After your fix, `rate <= 1.0` holds for arbitrary traces, so tighten the assertion to `0.0 <= rate <= 1.0` and update the stale comment. One line, and the property suite proves exactly the invariant this PR restores.

**3. One semantic nuance worth a docstring sentence.**

The new denominator counts only resolved attempts: unicast messages still queued when `max_ticks` truncates a run no longer lower the rate, and `ctx.schedule()` timer fires (logged as a `receive` with no `send`) now count as delivery attempts. Both defensible (the old formula handled timers worse), but a sentence in the `_delivery_rate` docstring would save the next reader the archaeology.

Housekeeping: the charter is one problem, one PR per participant. Infra fixes like this and your merged #96 are welcome alongside that; just state in #88 that the registry resolver is your primary submission.

Suggested path forward:

1. Rebase on main, confirm lint, typecheck, and test go green on GitHub.
2. Tighten `test_success_rate_bounded` and fix its comment.
3. Optional: the docstring sentence from point 3.

SwasthikaDev:
Thanks for the detailed review. All three addressed:

1. **CI now runs.** Merged current `main` into the branch and pushed, so lint/typecheck/tests trigger on GitHub and the mergeability state clears.
2. **`test_success_rate_bounded` tightened.** It now asserts `0.0 <= rate <= 1.0` (was `rate >= 0.0`), and the stale `receives / sends` comment is replaced with the `receives / (receives + dropped)` reasoning, so the property suite proves the `<= 1.0` invariant this PR restores.
3. **Docstring nuance added.** `_delivery_rate` now spells out the two resolved-attempt edge cases: messages still queued when `max_ticks` truncates a run, and `ctx.schedule()` timer fires (a `receive` with no `send`).

Full local run on the merged tree is green: ruff check/format clean, pyright strict 0 errors, pytest 1151 passed / 1 skipped. For the charter: my **registry resolver (#88) is my primary submission**; this and the merged #96 are infra fixes alongside it.

JamesCarnley:
Thank you, @SwasthikaDev. This patch correctly preserves an important metric invariant: a delivery or success rate must remain within 0..1 and define its denominator across fan-out, drops, queued work, and timer callbacks. Rebuilt Town no longer has the legacy delivery_rate or success_rate computation; upstream metric names are imported only as metadata and are not executed. There is therefore no current function to repair. I am closing this obsolete legacy patch and retaining its broadcast/drop/property cases for any future metrics implementation. No contributor code was merged or executed during this review.

PR #99, opened 2026-07-08
Hackathon/amit auth delegation

Amitk003:
DelegatableAuth Plugin Problem 04

Adds HMAC-chain delegation and prefix-trie cascading revocation to Nanda Town's auth layer. The reference jwt_auth plugin issues flat tokens with no delegation, no scope monotonicity, and no audience enforcement.

Key features:
- delegate() creates constrained child tokens with an HMAC chain proving ancestry
- Prefix-trie revocation: revoking "alice" cascades to all "alice.*" delegates
- Scope monotonicity: delegates can't grant scopes they don't own
- Audience enforcement: verify requires resource_id to match aud
- 3 adversarial validators proving jwt_auth fails all three attacks while DelegatableAuth blocks them
DOD: 820 tests pass, ruff/pyright clean, scenario YAML included.


dhve:
Thanks @Amitk003. This is one of the more carefully built entries in the Problem 04 pile. Embedding ancestor payloads in the child token so `verify()` can recompute the whole chain statelessly is the right instinct, `TestValidatorsFailAgainstJwt` does the fails-on-baseline / passes-on-yours pattern the charter asks for, and adding `presenter` as an optional keyword-only arg keeps `verify()` compatible with the `Auth` protocol in `nest_core/layers/auth.py`. Good API discipline.

**1. The token `path` is not authenticated, which breaks cascading revocation.**

`_chain_nonce()` signs the ancestor payloads plus the stripped leaf payload, but the `path` list is serialized separately (`<path_hex>|<payload_b64>|<nonce>`) and never enters the HMAC. So the holder of a child token whose ancestor was revoked can rewrite `path_hex` to a fresh prefix, keep payload and nonce unchanged, and `verify()` passes: the HMAC still checks out and the revocation scan runs against the attacker-chosen path. The gap cuts the other way too: `revoke()` deserializes without checking the nonce, so anyone can craft a token whose path is `[root_handle]` (visible inside every descendant token) and kill the whole tree. This is the headline feature of the problem, and `test_tampered_token_rejected` only tampers the nonce, never the path. Fix: fold `path` into the signed material (e.g. include it in the payload before `_canonical()`), verify the nonce inside `revoke()`, and add a regression test that rewrites `path_hex` on a revoked-ancestor child and asserts `RevokedAncestorError` still fires.

**2. The scenario is not runnable: no `delegated_auth` task factory exists.**

`scenarios/delegated_auth.yaml` sets `task.type: delegated_auth`, but `get_scenario_factory()` in `nest_core/scenarios.py` only knows the types wired into `_try_load_builtin()`, and there is no `scenarios_builtin/delegated_auth.py`. The run command in your README raises `KeyError: No scenario factory registered for 'delegated_auth'`. So no scenario run ever calls `issue`/`delegate`/`verify`, and the charter's scenario-demonstration requirement is unmet. `scenarios_builtin/identity_rotation.py` is the closest template: build the delegation tree, revoke one intermediary at `revoke_at_round`, emit trace lines the validators can check.

**3. Random handles break Tier 1 determinism.**

`issue()` and `delegate()` call `secrets.token_hex(16)`, the unseeded RNG the charter's determinism rule bans, and your YAML declares `tier: 1` with `seed: 42`: same seed must give a byte-identical trace, impossible once random handles land in tokens. Your own `test_deterministic_issue` concedes it ("Tokens differ because handles are random"), so `TestDeterminism` currently asserts the opposite of its name. Derive handles deterministically, e.g. `HMAC(secret, f"{subject}:{counter}")` with a per-instance counter, or accept an injected seeded `random.Random`.

**4. `_chain_nonce()` swaps HMAC key and message.**

`hmac.new(chain_bytes, secret, hashlib.sha256)` uses the attacker-visible chain bytes as the key and the secret as the message. Your own module docstring says `nonce = HMAC(secret, concat(parent_payloads..., child_payload))`, and `JwtAuth._sign` puts the secret first. It only works because both sides call the same function, so no test catches it, but it inverts HMAC's security model. One-line fix: `hmac.new(secret, chain_bytes, hashlib.sha256)`.

**5. Housekeeping: merge conflict, two open PRs, no CI run.**

The branch currently conflicts with main and needs a rebase. The charter is one problem per participant, single PR, and you have #99 and #119 open; pick your primary or explain how they relate. No checks have run on this branch yet, and the judge panel does not score broken PRs.

**Suggested path forward:**

1. Authenticate `path` in the HMAC, verify the nonce in `revoke()`, add the path-rewrite regression test.
2. Fix the key/message order in `_chain_nonce()`.
3. Replace `secrets.token_hex` with deterministic handle derivation.
4. Add a `delegated_auth` factory (pattern: `identity_rotation.py`) and wire it in `_try_load_builtin()`.
5. State in the PR body what differentiates this from the roughly twenty other open Problem 04 entries. Stateless chain verification via embedded ancestors is a real one; say so and make it load-bearing in a test. (Nit: the README says "prefix-trie" but `_revoked_paths` is a plain set of prefixes.)
6. Rebase onto main, run `make ci-local`, and confirm all 5 stages green before requesting re-review.

Leaving this open. Ping me on the hackathon channel if the path-authentication fix is unclear.

JamesCarnley:
Thank you, @Amitk003. The important invariant is that every field consulted for revocation, including ancestry and path, must be signed, and revoke must verify authority before mutating state. The frozen head corrected several earlier issues, but the legacy scenario remains unwired and the old plugin API has no current counterpart. Rebuilt Town's flat grants already sign their complete payload and test the selected one-hop contract; they do not implement prefix-path delegation. I am closing this obsolete implementation while retaining path-rewrite and verify-before-revoke cases for a future delegation profile. No contributor code was merged or executed during this review.

PR #100, opened 2026-07-08
Add delegatable capability tokens with cascading revocation (auth layer)

rudranaresh0201:
The default jwt plugin can't delegate — there's no way for an agent
holding a token to hand a scoped, temporary copy to another agent
without going back to the issuer, and revoking a token has zero effect
on anything derived from it, because nothing tracks that relationship.

This adds a macaroon-style plugin instead: each delegated token is
signed using its parent's own signature as the key, so the whole chain
gets re-verified back to the root every time. That's what makes
revocation cascade for free — kill the root and every token minted
under it stops verifying, without maintaining a list of descendants
anywhere.

Includes tests for the three ways this can be attacked (claiming more
scope than you were given, a token outliving its revoked/expired
parent, a token being used by the wrong agent) with the same attacks
run against the plain jwt plugin to show it has no defense against any
of them. Also ships a 16-agent scenario that runs a real delegation
tree through the simulator and shows revoking one branch only kills
that branch, while revoking the root kills everything.

Solves docs/hackathon/problems/04-auth-capability-delegation.md.

---

**Update:** rebased on `main` and addressed review feedback. Renamed
`delegatable` → `chained_capability` (plugin key, scenario name, and
module path) since PR #138 merged an implementation under the original
names while this PR was open — this is an independent, functionally
different design, not a resubmission of the same idea. Three concrete
differences from the merged reference plugin, since problem 04 has
~20 entries and it's worth being specific about what's actually
different here rather than just restating the mechanism:

1. **Audience binding can't be skipped.** The merged reference plugin
   splits `verify()` (no audience check) from `verify_presented()`
   (checks it) — a caller can still pick the wrong one and the
   wrong-agent-presents-the-token attack goes unguarded. Here,
   presenting a *delegated* (non-root) token through `verify()` without
   `expected_audience` is itself a rejection (`AudienceMismatchError`),
   not an optional parameter a caller can forget. Root tokens are
   unaffected — there's no delegation hop to confuse there.
2. **The logical clock never reads real time, under any code path.**
   The merged reference plugin's `_now()` falls back to `time.time()`
   if `clock` was never explicitly set, so it can go non-deterministic
   silently. This plugin's clock starts at `0.0` and has no fallback —
   a forgotten `set_clock()` call fails loud (wrong tick) instead of
   quietly breaking trace reproducibility.
3. **The adversarial checks are wired into the real judging path, not
   just a private test file.** The scenario's `CoordinatorAgent` now
   runs a live probe of all three named attacks (scope escalation,
   stale ancestor, audience confusion) and emits a `delegation_audit`
   event per attack. `validate_chained_capability_*` functions are
   registered under `VALIDATORS["chained_capability_delegation"]` in
   `nest_core/validators.py`, so the leaderboard's own validator runner
   sees these checks directly from a trace. (Notably, the merged
   reference PR's `delegation_validators.py` functions are exported
   from its `validators/__init__.py` but never actually registered in
   `nest_core.validators.VALIDATORS` — confirmed by grepping `main` —
   so its own adversarial checks aren't reachable by the validator
   runner either.)

All five local CI stages green (`uv sync`, `ruff check`,
`ruff format --check`, `pyright`, `pytest -v`: 1174 passed, 1 skipped).

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -32,6 +32,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
     ("auth", "delegatable"): f"{_REF}.auth.delegatable:DelegatableAuth",
     ("auth", "mesh_revocable"): f"{_REF}.auth.mesh_revocable:MeshRevocableAuth",
+    ("auth", "chained_capability"): f"{_REF}.auth.chained_capability:ChainedCapabilityAuth",
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
     ("trust", "agent_receipts"): f"{_REF}.trust.agent_receipts:AgentReceiptsTrust",
     ("trust", "parc"): f"{_REF}.trust.parc:ParcTrust",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -103,6 +103,12 @@ def _try_load_builtin(name: str) -> None:
         from nest_core.scenarios_builtin.gossip_byzantine import gossip_eclipse_factory
 
         register_scenario("gossip_eclipse", gossip_eclipse_factory)
+    elif name == "chained_capability_delegation":
+        from nest_core.scenarios_builtin.chained_capability_delegation import (
+            chained_capability_delegation_factory,
+        )
+
+        register_scenario("chained_capability_delegation", chained_capability_delegation_factory)
     elif name == "memory_concurrent_writers":
         from nest_core.scenarios_builtin.memory_concurrent_writers import (
             memory_concurrent_writers_factory,
--- a/packages/nest-core/nest_core/scenarios_builtin/chained_capability_delegation.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/chained_capability_delegation.py
@@ -1,0 +1,336 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Chained-capability-delegation scenario -- exercises the
+``chained_capability`` auth plugin.
+
+Topology: 1 ``coordinator`` delegates a scoped, time-boxed token to each of
+3 ``intermediary`` agents, and each intermediary further delegates a
+narrower token to 4 ``leaf`` agents (16 agents total, tree depth 2).
+
+Timeline:
+
+1. ``t=0`` -- coordinator issues a root token (``read, write, admin``),
+   delegates a ``read, write`` token to each intermediary, and schedules
+   two future self-ticks: revoke one intermediary's branch at
+   ``t=REVOKE_BRANCH_TICK``, revoke the root at ``t=REVOKE_ROOT_TICK``.
+2. Each intermediary verifies its handoff, delegates a ``read``-only
+   token to each of its 4 leaves, and arms a periodic re-check tick.
+3. Leaves verify their handoff and arm the same periodic re-check.
+4. At ``REVOKE_BRANCH_TICK`` the coordinator revokes *only*
+   ``intermediary-1``'s token -- its 4 leaves go dark on their next
+   re-check, but ``intermediary-0``/``intermediary-2`` and their leaves
+   are unaffected (revocation is scoped to the subtree, not global).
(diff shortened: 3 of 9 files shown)


dhve:
Thanks @rudranaresh0201. This is a careful, well-built take on problem 04. The macaroon construction in `delegatable.py` is exactly the right shape: `_verify_chain` re-keys each link's HMAC with the previous link's signature and re-derives the whole chain root-to-leaf, so revoking any ancestor's `jti` kills every descendant by construction with no per-child revocation list. The determinism work is deliberate and documented: `_next_jti`'s monotonic counter plus the `set_clock` logical clock (with the never-rewinds guard) keeps wall-clock time and RNG out of the sign/verify path entirely, and you backed it with an explicit same-seed replay test. The baseline-pairing tests (`test_naive_jwt_has_no_revocation_propagation`, `test_naive_jwt_has_no_audience_binding`) are precisely the "fails on the reference plugin, passes on mine" shape the brief wants, and extending `verify()` with `expected_audience` as an optional kwarg rather than a required positional keeps the base `Auth` contract intact.

The plugin logic is solid, and the unit tests in `test_delegatable_auth.py` cover the happy path, cascading revocation, and all three attack classes. A few things to sort before it is merge-and-score ready:

**1. The adversarial checks are shipped as pytest, not as a registered validator.** The brief asks for a validator that "must FAIL against the default `jwt` plugin and PASS against your plugin," and the charter says the leaderboard re-runs your scenario under the adversarial validators. Your three attacks are covered thoroughly, but they live in `test_delegatable_auth.py`, and nothing is registered in `nest_core/validators.py` (no `"delegated_auth"` key in `VALIDATORS`, no `validate_delegated_auth_*` function). Your scenario already emits the `revocation_detected` / `verified` / `revoked` events a trace validator would key on, so wiring one up is mostly mechanical: add `validate_delegated_auth_*` functions keyed into `VALIDATORS["delegated_auth"]` that assert the scoped-branch-then-root cascade. If the pytest form is meant to satisfy the brief, please say so explicitly in the PR body.

**2. Audience binding is opt-in.** `verify(token, expected_audience=None)` only catches the audience-confusion attack when the caller passes `expected_audience`; omit it and a token presented by the wrong holder verifies fine. Your docstring is candid that this is the relying party's job, which is defensible since it mirrors JWT `aud`-claim checking, but attack 3 is a required defense here. Either make the check non-optional for delegated (non-root) tokens, or spell the trust assumption out in the PR body and threat model so a judge does not read it as a gap. The scenario itself does pass `expected_audience` on every verify, which is the right call; the concern is only the plugin API's default.

**3. Say what sets this apart.** Problem 04 is the most crowded slot in the pool right now (roughly twenty open PRs). Your body explains the mechanism well but not what differentiates it. Lead with the specific differentiator (HMAC-chain-as-key with cascading revocation for free, the 16-agent scoped-branch-vs-root tree, the byte-identical replay test) and make it load-bearing in a test, since novelty is scored against merged peers.

Housekeeping: the branch is currently in conflict with `main` (GitHub shows it as conflicting), so it needs a rebase before it can merge or be scored, most likely on the shared `nest_core/plugins.py` and `scenarios.py` registries that other problem-04 PRs also touch. No CI checks have run on the branch yet either; the judge panel does not score broken PRs, so after the rebase run `uv sync`, `ruff check`, `ruff format --check`, `pyright`, and `pytest -v` locally and confirm all green. Branch is `hackathon/rudra-auth-delegation`; the pattern is `hackathon/<handle>-<theme>`, so worth checking `rudra` against your handle `rudranaresh0201`.

Suggested path forward:

1. Rebase on `main` and resolve the registry conflicts.
2. Decide validator vs pytest for the adversarial checks; if validator, register `delegated_auth` in `VALIDATORS`.
3. Make audience binding non-optional or document the assumption in the threat model.
4. Add the differentiator paragraph to the PR body.
5. Get all five CI stages green.

Leaving this open. Ping me on the hackathon channel if the validator-registration path is unclear.

JamesCarnley:
Thank you, @rudranaresh0201. The strongest reusable case is mandatory leaf-audience presentation: omission or mismatch must reject independently of revocation, and a MAC-valid widened link must still fail at verification. Current Town binds a flat session grant to run, participant, permissions, expiry, and session key, but has no delegated leaf or child audience. This is a strong future acceptance profile, not a current flat-grant defect. I am closing the legacy PR while preserving audience, widening, deterministic-time, subtree/sibling, and cross-verifier cases for later authorization work. No contributor code was merged or executed during this review.

PR #103, opened 2026-07-08
hackathon/mwarsss-delegatable-auth

mwarsss:
## Summary

Implements **Problem 04 -> Auth Capability Delegation** from the NandaHack charter.  Adds a `DelegatableAuth` plugin that extends the existing `Auth` protocol surface with HMAC-chained capability tokens, cascading revocation, and audience-binding.

- **Plugin** -> `DelegatableAuth` (auth layer, name `delegatable`).  Extends the `Auth` protocol (`issue` / `verify` / `revoke`) with an additive `delegate(parent_token, audience, scopes, ttl) → Token` method.  No existing callers are modified.
- **Three attacks blocked** -> scope escalation, stale-parent revocation, audience confusion.  Each attack raises a distinct typed exception (`ScopeEscalationError`, `RevokedAncestorError`, `AudienceConfusionError`), all `ValueError` subclasses for drop-in compatibility.
- **HMAC chain** -> based on Birgisson et al. 2014 (Macaroon construction).  Each child token's signing key is derived from the parent's HMAC output, so revoking a parent transitively invalidates every descendant without a per-child revocation list.
- **Audience binding** -> `verify(token, *, caller: AgentId | None = None)` keyword-only arg; backward-compatible with all `Auth` protocol callers that don't supply `caller=`.
- **Three adversarial validators** -> `check_scope_escalation_blocked`, `check_stale_parent_blocked`, `check_audience_confusion_blocked` in `nest_plugins_reference/validators/auth_validators.py`, returning `ValidatorReport` (same type as `gossip_validators.py`).
- **Scenario** -> 16-agent two-level tree: 1 coordinator → 3 intermediaries → 12 leaves.  Scenario YAML at `scenarios/delegated_auth.yaml` with `layers.auth: delegatable`.
- **Wiring** -> `_BUILTINS` in `plugins.py`, `_try_load_builtin` in `scenarios.py`, and a `[project.entry-points."nest.plugins.auth"]` entry in `pyproject.toml`.

## Design choices

- **No types.py modification** ->`AuthContext` has no `metadata` field; audience is carried in the token payload itself and checked via the `caller=` keyword argument rather than polluting shared types.
- **Constructor mirrors `JwtAuth`** -> `__init__(self, secret: bytes, clock: float | None = None)`.  Auth is a single shared service (no per-agent instances needed), matching the existing pattern.
- **No new dependencies** -> uses only `hmac`, `hashlib`, `json`, `uuid` from the standard library.

## Files changed

| File | Change |
|------|--------|
| `packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable.py` | New plugin (~230 lines) |
| `packages/nest-plugins-reference/nest_plugins_reference/validators/auth_validators.py` | Three adversarial validators |
| `packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py` | Scenario factory (16 agents) |
| `scenarios/delegated_auth.yaml` | Scenario YAML |
| `packages/nest-plugins-reference/tests/test_delegatable_auth.py` | 32 example-based tests |
| `packages/nest-plugins-reference/tests/test_delegatable_auth_properties.py` | 5 Hypothesis property tests |
| `packages/nest-core/nest_core/plugins.py` | +1 line: `("auth", "delegatable")` in `_BUILTINS` |
| `packages/nest-core/nest_core/scenarios.py` | +4 lines: `delegated_auth` branch in `_try_load_builtin` |
| `packages/nest-plugins-reference/pyproject.toml` | +3 lines: `[project.entry-points."nest.plugins.auth"]` |

## Test plan

- [x] `uv sync` — clean install
- [x] `uv run ruff check .` — no errors
- [x] `uv run ruff format --check .` — no reformats needed
- [x] `uv run pyright` — 0 errors, 0 warnings
- [x] `uv run pytest -v` — 828 passed, 1 skipped, 0 failed (baseline: 377 tests; +37 new, rest unchanged, +414 from other PRs already in main)

Quick smoke test:

```python
from nest_plugins_reference.auth.delegatable import DelegatableAuth
from nest_core.types import AgentId
import asyncio

auth = DelegatableAuth(secret=b"demo")
async def demo():
    root  = await auth.issue(AgentId("coord"), ["admin", "read", "write"])
    child = await auth.delegate(root, AgentId("worker"), ["read"], ttl=600.0)
    ctx   = await auth.verify(child, caller=AgentId("worker"))
    assert ctx.scopes == ["read"]

asyncio.run(demo())
```

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -25,6 +25,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("registry", "in_memory"): f"{_REF}.registry.in_memory:InMemoryRegistry",
     ("registry", "gossip"): f"{_REF}.registry.gossip:GossipRegistry",
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
+    ("auth", "delegatable"): f"{_REF}.auth.delegatable:DelegatableAuth",
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
     ("trust", "agent_receipts"): f"{_REF}.trust.agent_receipts:AgentReceiptsTrust",
     ("payments", "prepaid_credits"): f"{_REF}.payments.prepaid_credits:PrepaidCredits",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -129,3 +129,7 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("failure_detection", failure_detection_factory)
+    elif name == "delegated_auth":
+        from nest_core.scenarios_builtin.delegated_auth import delegated_auth_factory
+
+        register_scenario("delegated_auth", delegated_auth_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py
@@ -1,0 +1,182 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Delegated-auth scenario — HMAC capability tokens flow through a trust hierarchy.
+
+A coordinator issues a root token and delegates scoped child tokens to three
+intermediaries.  Each intermediary re-delegates narrower tokens to its four
+leaf agents.  Leaves verify their tokens and report success back to the
+coordinator.  The scenario validates the delegation chain, scope narrowing,
+and audience binding end-to-end.
+
+Example::
+
+    agents = delegated_auth_factory(config, plugins)
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from nest_core.scenario import ScenarioConfig
+from nest_core.sim.agent import AgentContext, StateMachineAgent
+from nest_core.types import AgentId, Token
+
+_COORD = AgentId("coordinator-0")
+_NUM_INTER = 3
+_LEAVES_PER_INTER = 4
+
+_INTER_IDS = [AgentId(f"inter-{i}") for i in range(_NUM_INTER)]
+_LEAF_IDS = [AgentId(f"leaf-{i}-{j}") for i in range(_NUM_INTER) for j in range(_LEAVES_PER_INTER)]
+
+
+class CoordinatorAgent(StateMachineAgent):
+    """Issues a root capability token and delegates to every intermediary.
+
+    On start the coordinator issues a root token with scopes
+    ``["admin", "exec", "read", "write"]`` and forwards a scoped child token
+    (``["exec", "read", "write"]``) to each intermediary.
+
+    Example::
+
+        agent = CoordinatorAgent(AgentId("coordinator-0"))
+    """
+
+    def __init__(self, agent_id: AgentId) -> None:
+        self._id = agent_id
+        self._verified_count = 0
+
+    async def on_start(self, ctx: AgentContext) -> None:
+        """Issue root token and delegate to intermediaries.
+
+        Example::
+
+            await agent.on_start(ctx)
+        """
+        auth = ctx.plugins.get("auth")
+        if auth is None:
+            return
+
+        root: Token = await auth.issue(self._id, ["admin", "exec", "read", "write"], ttl=3600.0)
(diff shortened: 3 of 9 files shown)


dhve:
Thanks @mwarsss. This is one of the more carefully built Problem 04 submissions in the queue. The macaroon-style key chain in `delegatable.py` (each child's signing key is the parent's HMAC output, and revocation cascades through the ancestry walk in `verify()` with no per-child bookkeeping) is the right construction, the keyword-only `caller=` on `verify()` extends the `Auth` protocol without breaking a single existing caller, and `TestAdversarialValidators` does the fails-on-baseline pattern properly: all three validators are shown failing against `JwtAuth` and passing against `DelegatableAuth`. Most PRs on this problem skip that last part.

**1. The scenario crashes at startup: `ctx.plugins.get("auth")` is the class, not an instance.**

`_resolve_plugins()` in `nest_core/runner.py` puts plugin classes into the dict, and scenario factories instantiate shared services themselves (see `_instantiate_plugins` in `scenarios_builtin/empic_payments.py`, or the `isinstance(cls, type)` checks in `marketplace.py`). `delegated_auth_factory` never does this, so `CoordinatorAgent.on_start` calls `await auth.issue(...)` on the bare `DelegatableAuth` class and dies with a TypeError before the first tick. `scenarios/delegated_auth.yaml` cannot currently run, and a runnable scenario is the charter's core deliverable. None of your 37 tests runs the scenario, which is how this slipped through; copy the end-to-end pattern from `packages/nest-core/tests/test_receipt_reputation.py`.

**2. Determinism: `uuid.uuid4()` and the `time.time()` fallback sit in the simulation path.**

`_mint()` stamps every token with a random `tid`, and with no injected clock, `iat` and `exp` come from wall clock. Tokens are the message payloads here, so all of that lands in the trace: same seed, different bytes. Your YAML declares `tier: 1`, where byte-identical traces are non-negotiable. Fix: derive token ids deterministically (a monotonic counter, or an HMAC over the fields you already sign), instantiate the plugin with a fixed injected clock, and add a same-seed-twice sha256 test (pattern: `test_same_seed_identical_trace` in `test_receipt_reputation.py`).

**3. Audience binding is one `delegate()` call away from being bypassed.**

`delegate()` verifies the parent without `caller=` and mints a child for whatever audience the caller names. So the thief that `AudienceConfusionError` stops at `verify()` just runs `await auth.delegate(stolen, AgentId("intruder"), ["read"], ttl=60.0)` and verifies the child as themselves. Attenuation-by-holder is legitimate macaroon behavior, but then audience confusion should not be advertised as a blocked attack class in the module docstring and PR body. Either require the delegator to prove they are the parent's audience (a `caller=` kwarg on `delegate()`, checked against the parent's `aud`), or rewrite the threat model honestly. Add the steal-then-redelegate adversarial test either way; it is the first attack a judge tries.

**4. Rebase and branch name.**

The PR is currently in a CONFLICTING state against main (`_BUILTINS` in `plugins.py`, `_try_load_builtin` in `scenarios.py`, and the entry-point tables in `pyproject.toml` have all grown since you branched), and no CI checks have run on this branch. The judge panel does not score broken PRs. Separately, the charter's branch format is `hackathon/<handle>-<theme>` with a slash; yours is `hackathon-mwarsss-delegatable-auth`, and mismatched names are subject to auto-close. Renaming the branch in GitHub's web UI carries the open PR along.

**5. Say what differentiates you: Problem 04 is the most crowded in the queue.**

Roughly twenty open PRs implement delegatable capability tokens. Your PR body describes the design well but never positions it against that field. Your real differentiators (key-chain revocation with no per-child revocation list, typed `ValueError`-compatible exceptions, validators proven against the jwt baseline) deserve a short "why this one" paragraph; novelty is scored against the field.

**Suggested path forward:**

1. Instantiate `DelegatableAuth` in `delegated_auth_factory` (fixed clock, seed-derived secret) and add an end-to-end scenario test.
2. Replace `uuid4` token ids with a deterministic scheme and add the trace-hash determinism test.
3. Decide the audience story: enforce `caller=` on `delegate()` or reframe the docstring, and add the steal-then-redelegate test.
4. Rebase onto current main, fix the branch name, and confirm all five CI stages are green.
5. Add the differentiation paragraph to the PR body.

Leaving this open. Ping me on the hackathon channel if the instantiation pattern is unclear.

PR #104, opened 2026-07-08
hackathon/Implement delegatable capability tokens and scenario doctor

tanishaacodes:
# Nanda Town Hackathon Submission: Phase 1 & Phase 2

This PR contains the implementation of **Delegatable Capability Tokens with Cascading Revocation** (Phase 1) and the **Scenario Doctor FastAPI validation service** (Phase 2). 

---

### Phase 1: Auth Plugin & Verification
* [packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable.py](file:///Users/tani/nandatown/packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable.py):
  - Class `DelegatableAuth` implementing the `Auth` interface protocol.
  - Implements Macaroon-style HMAC chaining signature derivation for offline delegation.
  - Recursively checks the delegation chain during `verify()` and raises `RevokedAncestorError` if any parent/ancestor JTI is marked revoked.
  - Generates deterministic `jti` values using SHA-256 hashes of the payload to preserve simulation determinism.
* [packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py](file:///Users/tani/nandatown/packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py):
  - Simulation scenario setup containing scope escalation, stale parent, and audience confusion attacks.
  - Integrates a `VirtualClock` driven by the simulator's tick-based clock (`ctx.time`) for trace reproducibility across random seeds.
  - Dynamically patches standard `jwt` auth at runtime to mock delegation layouts while omitting security checks.
* [packages/nest-core/nest_core/validators.py](file:///Users/tani/nandatown/packages/nest-core/nest_core/validators.py):
  - Implements adversarial checks `validate_auth_no_scope_escalation`, `validate_auth_no_stale_parent`, and `validate_auth_no_audience_confusion`.
  - Correlates trace events using `(jti, presenter)` tuples to prevent collision of stolen tokens.
* [scenarios/delegated_auth.yaml](file:///Users/tani/nandatown/scenarios/delegated_auth.yaml):
  - Scenario configuration file using `auth: delegatable`.

### Phase 2: Scenario Doctor Service
* [apps/scenario-doctor/main.py](file:///Users/tani/nandatown/apps/scenario-doctor/main.py):
  - Standalone FastAPI application.
  - Endpoint `GET /health`: Basic health state check.
  - Endpoint `GET /layers`: Lists the 12 system layers and valid reference plugins (including `delegatable`).
  - Endpoint `POST /validate-scenario`: YAML structure checker validation.
  - Endpoint `POST /lint`: Verifies config layers and plugins against known entries.
* [apps/scenario-doctor/Dockerfile](file:///Users/tani/nandatown/apps/scenario-doctor/Dockerfile):
  - Multi-stage containerized setup using `python:3.12-slim` base image.
* [apps/scenario-doctor/railway.json](file:///Users/tani/nandatown/apps/scenario-doctor/railway.json):
  - Configures the builder to run under the `DOCKERFILE` driver.
* [docs/hackathon/solutions/scenario-doctor-skill.md](file:///Users/tani/nandatown/docs/hackathon/solutions/scenario-doctor-skill.md):
  - OpenClaw-compatible `SKILL.md` documenting curl invocation formats and schemas.

---

### `GET /health`
* **Response**: `{"status": "ok"}`

### `GET /layers`
* **Response**: Dictionary mapping of NANDA Town's 12 layers and plugins.

### `POST /validate-scenario`
* **Input Content-Type**: `text/yaml`
* **Input Example**:
  ```yaml
  name: test_scenario
  duration: 100


dhve:
Thanks @tanishaacodes. The core of this is good work. `DelegatableAuth` is the right shape for Problem 04: each child signature is HMAC-keyed on the parent's signature, scope subsets are enforced at both `delegate()` and `verify()` time, a child TTL past the parent's expiry is rejected at both ends, revocation walks the full `jti_list` and raises a typed `RevokedAncestorError` for ancestors, and `jti` values are derived from SHA-256 over the token fields rather than `uuid4`, which keeps traces reproducible. `test_delegated_auth.py` has exactly the pattern we ask for (validators fail on `jwt`, pass on `delegatable`, plus a byte-identical determinism check across three seeds), and keying validator interactions on `(jti, presenter)` so the stolen and legitimate presentations of the same token stay distinct is a nice touch.

**1. The audience-confusion result is decided by the scenario, not by your plugin.**

In `CoordinatorAgent.on_message` (`scenarios_builtin/delegated_auth.py`) you only enforce presenter-matches-subject when `self._auth.__class__.__name__ == "DelegatableAuth"`. So `delegatable` passes the audience-confusion check because the harness grants it a check it denies to `jwt`, while `DelegatableAuth.verify()` never sees the presenter and provides no audience protection itself. Any third auth plugin would be misclassified by name. This weakens both correctness and test rigor: the validator is measuring scenario branching, not plugin behavior. Suggested fix: give `verify()` an optional `presenter: AgentId | None = None` kwarg (this stays compatible with the `Auth` protocol), enforce the `aud` check inside the plugin, give the patched baseline's `dummy_verify` the same ignored kwarg, and have the coordinator pass `presenter=sender` uniformly. Then `jwt` genuinely fails because it ignores the presenter, and yours genuinely rejects.

**2. Repo-root leftovers from a different submission need to come out.**

`my_privacy_layer.py` and `test_privacy.py` are a trust-gated privacy layer, which already exists on main as `nest_plugins_reference/privacy/trust_gated.py` (merged in #115). Worse, the root `pyproject.toml` changes register `trust_gated = "my_privacy_layer:TrustGatedPrivacy"` under `nest.plugins.privacy`, colliding with that merged plugin's entry-point name, and `[tool.setuptools] py-modules = ["my_decaying_memory"]` references a module that does not exist anywhere in this PR. The Apache license classifier was also dropped from the root project. The charter says do not touch code outside your layer/scenario/validator; please revert all of these.

**3. Phase 2 (scenario doctor) does not belong in this PR.**

The charter scopes a submission to one problem: plugin plus adversarial validator plus scenario. The FastAPI app adds `fastapi`, `uvicorn`, and `httpx` to root dev-dependencies (charter: no new deps beyond pure Python and existing extras), adds a repo-root `.dockerignore` plus Docker and Railway deploy config under `apps/scenario-doctor`, and carves `apps/scenario-doctor` out of pyright in shared config. Also note `apps/scenario-doctor/tests/test_main.py` will never run in CI: pytest `testpaths` is `["packages", "scripts"]`. The service itself is a reasonable idea, but bundled here it reads as scope creep and dilutes the submission. Move it to its own repo or at minimum its own PR.

**4. Branch name and the one-PR rule.**

The branch is `hackathon/versioned-comms`: that names a different problem entirely (comms versioning, already merged as #18) and is missing your handle. The charter format is `hackathon/<handle>-<short-theme>`, and mismatched branch names are subject to auto-close. Renaming the branch on your fork through the GitHub UI will carry the PR along. Separately, you also have #109 and #122 open, and the charter is one problem per participant in a single PR. Please pick your primary submission or explain in each body how they relate.

**5. The patched `jwt` baseline does not do what its docstring says.**

`_patch_jwt_auth_instance` replaces `delegate`, `verify`, and `_now`, but not `revoke`. `JwtAuth.revoke()` stores the full token string in `_revoked`, while your `dummy_verify` looks up `jti` values, so under the patched baseline even direct revocation never triggers, despite the comment claiming direct revocation is checked with only the cascade missing. It does not change the validator outcome, but the baseline should behave as advertised: patch `revoke` to store the last block's `jti`. While you are in tests, a hypothesis property test (random delegation chains, revoking any ancestor implies `RevokedAncestorError`) would lift test rigor beyond example cases.

Three more notes. The branch currently has conflicts with `main`, so it needs a rebase before anything can land. No CI checks have reported on this branch, and the judge panel does not score broken PRs, so please run `make ci-local` and confirm all 5 stages green. And on novelty: there are roughly twenty open PRs on Problem 04, so state in the PR body what differentiates your take (the deterministic jti derivation and the `(jti, presenter)` correlation are real candidates) and make that differentiator load-bearing in tests.

**Suggested path forward:**

1. Strip the PR to Phase 1: `delegatable.py` and its tests, the `delegated_auth` scenario/factory/validators, the `plugins.py` and `scenarios.py` registrations, and the YAML. Revert root `pyproject.toml`, `.dockerignore`, `my_privacy_layer.py`, `test_privacy.py`, and `apps/scenario-doctor`. Also register `delegatable` under `[project.entry-points."nest.plugins.auth"]` in `packages/nest-plugins-reference/pyproject.toml`, the discovery path CONTRIBUTING documents.
2. Add the optional `presenter` kwarg to `verify()`, enforce `aud` in the plugin, and drop the class-name check in `CoordinatorAgent`.
3. Patch the baseline's `revoke` so the mock matches its docstring.
4. Rename the branch to `hackathon/tanishaacodes-delegatable-auth`.
5. Rebase onto current `main` to clear the conflicts, then run `make ci-local` and confirm all 5 stages green.
6. Update the PR body with your differentiator versus the other Problem 04 submissions.

Leaving this open. Ping me on the hackathon channel if the presenter-kwarg reshaping is unclear.

tanishaacodes:
hi vedh updated the repo under #165

PR #106, opened 2026-07-08
hackathon/dzaky-delegatable-auth: Problem 04 — Delegatable Capability Tokens

xDzaky:
### What this solves

Agents in NANDA Town currently have no way to mint bounded sub-credentials and pass them downstream without routing every authorisation request back through the original issuer. This PR adds a **macaroon-inspired capability-token auth layer** where agent A can carve a strictly-narrower token for agent B, B can further carve for C, and revoking A's root immediately invalidates the entire subtree — all in-process, with zero round-trips to an external authority.

---

### Design

```
root token (alice, scopes={read,write,admin}, max_depth=2)
 ├── child-1 (broker, scopes={read,write}, depth=1)
 │    └── leaf-1 (worker, scopes={read}, depth=2)
 └── child-2 (broker2, scopes={read}, depth=1)
```

Each token is HMAC-SHA256 signed using a chain anchored at the server secret:

```
root_sig  = HMAC(server_secret, root_claims)
child_sig = HMAC(root_sig,      child_claims)   ← cryptographically bound to parent
```

Revoking the root marks its `token_id` in an in-memory revocation set; `verify_capability` walks the ancestry chain and fails on the first revoked ancestor — **O(depth)** per verify, **O(n)** revoke-tree traversal.

---

### Files changed

| File | Role |
|---|---|
| `packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable.py` | Core plugin: `DelegatableAuth` implementing the `Auth` protocol |
| `packages/nest-plugins-reference/nest_plugins_reference/validators/auth_delegation_validators.py` | Event-list validator for scenario YAML |
| `packages/nest-core/nest_core/scenarios_builtin/auth_capability_delegation.py` | 16-agent simulation (1 coordinator, 3 intermediaries, 12 leaves) |
| `packages/nest-core/nest_core/scenarios.py` | Register new builtin scenario type |
| `packages/nest-core/nest_core/plugins.py` | Register `("auth", "delegatable")` in PluginRegistry |
| `packages/nest-plugins-reference/pyproject.toml` | Expose `DelegatableAuth` as entry-point plugin |
| `scenarios/auth_capability_delegation.yaml` | Root-level scenario YAML |
| `packages/nest-plugins-reference/nest_plugins_reference/scenarios/auth_delegation.yaml` | Event-list format for validator |
| `packages/nest-plugins-reference/tests/test_auth_delegation.py` | 12 unit tests |
| `packages/nest-plugins-reference/tests/test_auth_delegation_properties.py` | 8 Hypothesis property-based tests |
| `packages/nest-plugins-reference/tests/test_auth_delegation_scenario.py` | 4 scenario/integration tests |
| `docs/hackathon/solutions/04-auth-capability-delegation.md` | STRIDE threat model + attack tree |

---

### Test results

```
24 passed in 2.92s
ruff check   : 0 errors
ruff format  : pass
pyright      : 0 errors, 0 warnings
```

Coverage across:
- ✅ Root token issuance and round-trip verification
- ✅ Scope monotonicity (child scopes ⊆ parent scopes), property-tested
- ✅ Depth enforcement (delegation past `max_depth` raises `CapabilityError`)
- ✅ TTL bounding (child can never outlive parent), property-tested
- ✅ Cascading revocation — root revoke invalidates entire subtree
- ✅ Partial revocation — revoking one branch leaves sibling branches intact
- ✅ Adversarial inputs: NaN TTL, tampered payload, wrong issuer, scope escalation
- ✅ Structural replay determinism under same seed (Tier-1 reproducibility)

---

### Security properties

| Threat | Mitigation |
|---|---|
| Token forgery | HMAC-SHA256 with server secret; secret never serialised |
| Scope escalation | `delegate()` enforces `child_scopes ⊆ parent_scopes` at mint time |
| Audience confusion | Audience bound in signed payload; `verify_capability(audience=…)` required |
| TTL extension | Child `expires_at` clamped to `min(requested, parent.expires_at)` |
| Depth-limit bypass | `depth >= max_depth` check before any delegation |
| Post-revocation replay | O(depth) ancestry walk on every verify against in-memory revocation set |
| Float abuse (NaN/Inf) | `_require_positive_finite()` guards on every TTL/timestamp input |

---

### How agents use it

```python
auth = DelegatableAuth()
root = auth.issue_root(subject="alice", audience="nest",
                       scopes={"read", "write", "admin"},
                       ttl_seconds=3600.0, max_depth=2)

child = auth.delegate(root, subject="bob", scopes={"read", "write"})
leaf  = auth.delegate(child, subject="carol", scopes={"read"})

cap = auth.verify_capability(leaf, audience="nest", required_scopes={"read"})
# cap.subject == "carol", cap.scopes == frozenset({"read"})

auth.revoke_tree(root)   # invalidates root + child + leaf atomically
# auth.verify_capability(leaf, ...) now raises CapabilityError("revoked")
```

### DIFF
--- a/docs/hackathon/solutions/04-auth-capability-delegation.md
+++ b/docs/hackathon/solutions/04-auth-capability-delegation.md
@@ -1,0 +1,230 @@
+---
+title: Delegatable capability tokens — reference solution & threat model
+layer: auth
+problem: 04-auth-capability-delegation
+plugin: ("auth", "delegatable")
+difficulty: easy
+---
+
+# Delegatable capability tokens — reference solution & threat model
+
+## Solution overview
+
+The `DelegatableAuth` plugin replaces the default `JwtAuth` with a
+macaroon-inspired capability-token system. Three new primitives sit
+on top of the base `Auth` protocol:
+
+| Primitive | What it does |
+|---|---|
+| `issue_root(subject, audience, scopes, ttl_seconds, max_depth)` | Mint a root token with N levels of delegatable depth. Root tokens carry the full scope set and a `max_depth` ceiling. |
+| `delegate(parent, subject, audience, scopes, ttl_seconds)` | Carve a strictly-narrower child token from an existing one. Scopes must be a subset of the parent's, TTL must be <= parent's TTL, and the remaining depth budget must be > 0. Every delegation re-signs the child using the parent's HMAC context so revocation propagates transitively. |
+| `revoke_tree(token_str)` | Revoke a token and *every* descendant in a single O(n) traversal of the in-memory ancestry forest. The revoked set is returned so callers can audit how many tokens were invalidated. |
+| `verify_capability(token_str, audience, required_scopes, now)` | Verify signature, expiry, audience binding, scope sufficiency, and recursive revocation ancestry in one call. Raises `CapabilityError` with a typed reason string on every failure path. |
+
+### HMAC-chain construction
+
+```
+root_hash  = HMAC(server_secret,  root_claims)
+child_hash = HMAC(parent_hash,  child_claims)  ← anchored to parent
+```
+
+Each token carries `parent_id` and the parent's truncated hash in its
+claims. Verifying a token recomputes the expected chain from the known
+root — if any ancestor was revoked its `token_id` appears in the
+revocation set and verification fails with `revoked`.
+
+This means revoking the **root** token immediately invalidates every
+token in its subtree at the next `verify_capability` call, with zero
+per-child bookkeeping.
+
+---
+
+## Threat model
+
+### Assets
+
+1. **Token secrecy / integrity** — ability to forge, tamper, or replay a
+   capability token to gain unauthorised access.
+2. **Scope integrity** — ability to widen one's authority beyond what
+   the granting agent intended.
+3. **Audience binding** — ability to present a token issued for agent B
+   as if it were issued for agent A.
+4. **Liveness / timeliness** — ability to use an expired or revoked
+   token long past its intended expiry.
+5. **Delegation-tree integrity** — ability to delegate deeper than the
+   root allowed, or to create cycles / DAGs that confuse revocation.
+
+### Adversary model
+
+We assume an in-process, in-memory adversary that:
+
+- Can read *any* serialized token string it obtains (network sniffing,
+  shared memory, log files, side channels).
+- Can call the public `delegate()`, `verify_capability()`, `revoke_tree()`,
+  and `inspect()` APIs on the same `DelegatableAuth` instance.
+- **Cannot** read the `server_secret` (it lives in memory only, never
+  serialized).
+- **Cannot** forge a valid HMAC without the secret — so forging a root
+  token is infeasible.
+- **Can** attempt to delegate from a token it legitimately holds but
+  with escalated scopes, longer TTL, different audience, or deeper
+  depth.
+- **Can** attempt to verify a legitimate token after its parent was
+  revoked (observing a valid `parent_id` in the wild and replaying it).
+
+### Attack tree
+
+```
+1. FORGE_ROOT_TOKEN                               [IMPOSSIBLE]
+   └── 1.1 Guess/crack server_secret               [infeasible — HMAC-SHA256]
+   └── 1.2 Blind signature collision               [infeasible]
+   └── 1.3 NaN/Inf injection in claims             [BLOCKED — _check_finite]
+
+2. SCOPE_ESCALATION                               [BLOCKED]
+   └── 2.1 child scopes ⊈ parent scopes            [BLOCKED — strict subset check]
+   └── 2.2 child has broader scope than parent      [BLOCKED — max one scope set intersection]
+   └── 2.3 parent has no admin, child requests it  [BLOCKED — CapabilityError("scope")]
+
+3. TTL_ABUSE                                       [BLOCKED]
+   └── 3.1 child ttl > parent ttl                  [BLOCKED — clamped to parent.expires_at]
+   └── 3.2 child ttl = NaN / Inf / negative        [BLOCKED — _check_finite]
+   └── 3.3 token used after expiry                 [BLOCKED — _check_expiry → expires_at < now]
+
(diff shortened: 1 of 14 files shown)


dhve:
Thanks @xDzaky. This is one of the more thoroughly engineered Problem 04 submissions: `delegatable.py` is clean, `verify_capability` has a typed failure path for every check (signature, liveness, audience, scope, ancestry), the NaN/Inf guards via `_require_positive_finite` are a touch most entries skip, and the Hypothesis suite in `test_auth_delegation_properties.py` (particularly `test_revoking_intermediate_invalidates_only_its_subtree`) plus the threat-model doc with the attack tree are above the bar.

**1. The PR body claims HMAC chaining, but the code does not implement it.**

The body and the solutions doc both promise a parent-anchored chain (`child_hash = HMAC(parent_hash, child_claims)`, "cryptographically bound to parent"). But `_sign()` in `delegatable.py` signs every token, root and child alike, with `hmac.new(self.secret, payload, hashlib.sha256)`. The parent linkage is just a signed `parent` claim plus the server-side `_tokens` / `_children` maps. That is the anti-pattern the problem statement explicitly warns against (a delegated-from field in claims without HMAC chaining), and it means `delegate()` only works through the issuing instance, the one holding the secret and the `_tokens` map, which brushes against the other anti-pattern (delegation as central re-issuance). The fix is small because your structure is already right: in `delegate()`, key the child's HMAC with the parent's signature (`hmac.new(bytes.fromhex(parent.signature), child_payload, hashlib.sha256)`) and verify by recomputing the chain from the root. With roughly twenty open PRs on this problem, real macaroon-style chaining is also your clearest novelty differentiator. Right now the claim and the code diverge, which costs you on both correctness and docs.

**2. Determinism: uuid4 and wall-clock time leak into the trace.**

`_new_id()` is `uuid4().hex` and `_checked_now` falls back to `time()`, and the scenario agents call `issue_root` / `delegate` without `now`, so real timestamps and random token ids end up in message payloads. Your own replay test concedes this: `test_scenario_is_deterministic_under_replay` strips `payload`, `token`, `data`, `msg`, and `size` before comparing. The Tier-1 bar is byte-identical traces for the same seed, so a structural-only comparison is the test telling you the requirement is not met. Concrete fix: pass `now=ctx.time` (the simulator clock on `AgentContext`), replace uuid4 with a per-instance counter, and derive the secret from the scenario seed the way the `identity_rotation.py` builtin does (`identity_cls(aid, seed=b"identity-rotation:" + ...)`). Then delete the field-stripping and assert byte equality; that turns your weakest test into your strongest.

**3. The adversarial validator cannot actually run against the `jwt` baseline.**

Problem 04 requires the validator to FAIL against `jwt` and PASS against yours. `CapabilityDelegationValidator._apply` calls `issue_root`, `delegate`, and `verify_capability`, none of which exist on `JwtAuth`, and `AttributeError` is not in your caught exception tuple, so `CapabilityDelegationValidator(JwtAuth())` crashes instead of returning `passed=False`. The solution doc's verification table checks off "Validator fails against `JwtAuth`" as passing "by construction", but that failure is asserted, never demonstrated. `test_auth_adversarial_comparison.py` is good work, but it re-implements the attacks by hand rather than running the validator. Suggested: feature-detect in the validator (treat `delegate` events as plain `issue` when the plugin lacks the API) so that against `jwt` the post-revocation verify succeeds and the validator reports a real failure, then add one test asserting `CapabilityDelegationValidator(JwtAuth()).validate_events(...)` returns `passed=False`.

**4. The 16-agent scenario never revokes anything.**

The module docstring of `auth_capability_delegation.py` promises that at the end the coordinator's root is revoked and every leaf must fail verification, but no agent ever calls `revoke_tree` inside the simulation. `CoordinatorAgent.on_message` is `pass` and `LeafAgent.verified` is never read by anything. The flagship feature, cascading revocation, is only exercised in unit tests, not in the trace the judges replay. Have the coordinator revoke after the intermediary acks arrive (or on a scheduled tick), broadcast a re-verify request, and assert in the scenario test that all 12 leaves fail post-revocation.

**5. YAML placement is crossed, and the branch has conflicts.**

The problem asks for the delegation-tree scenario under `scenarios/`, but `scenarios/auth_capability_delegation.yaml` is a bespoke event-list format that no code in this PR loads (the validator test re-declares the same events inline in Python), while the runnable ScenarioConfig lives at `nest_plugins_reference/scenarios/auth_delegation.yaml`, whose `agents.count: 3` also disagrees with the 16 agents the factory builds. Swap the locations and load the event YAML from the test so the file is exercised. Separately, GitHub reports the branch as CONFLICTING with main, and no CI checks have reported on this branch yet. The judge panel does not score broken PRs.

**Suggested path forward:**

1. Rebase onto latest main and resolve `pyproject.toml` and `nest_core/scenarios.py` (main has grown a `nest.plugins.trust` entry-point table and new scenario branches since you branched).
2. Implement the parent-keyed HMAC chain in `_sign` / `delegate` so the code matches the PR body, or rewrite the body honestly around the revocation-list design.
3. Thread `now=ctx.time`, counter-based token ids, and a seed-derived secret through the scenario, then make the replay test byte-identical.
4. Make the validator runnable against `JwtAuth` and add the failing-baseline test.
5. Add in-simulation revocation to the coordinator, fix the YAML locations, and state in the PR body what differentiates this entry from the other open Problem 04 submissions.
6. Confirm `uv sync`, `ruff check`, `ruff format --check`, `pyright`, and `pytest -v` are all green locally before requesting re-review.

Leaving this open. Ping me on the hackathon channel if the chaining rework is unclear.

JamesCarnley:
Thank you, @xDzaky. The useful target is a portable holder-derived chain with verifier-enforced attenuation, finite bounded time/depth, byte-complete determinism, and a real post-revocation leaf attempt. At this head, the central holder-derived HMAC construction is absent, UUID/wall-clock nondeterminism is hidden by stripping token fields from replay comparisons, the validator can crash on the baseline, and the scenario does not exercise revocation. Current Town grants are flat and already test exact signatures, permissions, presentation, and finite times. I am closing this obsolete legacy implementation while retaining the portable-chain and non-vacuous attack cases for a future delegation profile. No contributor code was merged or executed during this review.

PR #94, opened 2026-07-08
[Hackathon] memory: add mv_register, a multi-value register CRDT that never drops a concurrent write

SwasthikaDev:
## Summary

Adds `mv_register`, a **Multi-Value Register CvRDT** for the memory layer. The existing `lww_register` is last-writer-wins: on two concurrent writes to a key it keeps one and **silently drops the other**. `mv_register` tags every write with a **version vector**, detects causal concurrency, and keeps concurrent writes as **siblings** — no concurrently-written value is ever lost. After N agents write one key concurrently, `lww_register` keeps 1 value and `mv_register` keeps N.

Ships as a plugin + adversarial validator + deterministic scenario + docs + 49 tests.

## What and where — exact file manifest

**New plugin** — `packages/nest-plugins-reference/nest_plugins_reference/memory/mv_register.py` (612 lines)
- `VersionVector` (**line 83**): immutable `{node: counter}` clock; `dominates` / `strictly_dominates` / `concurrent` / `merge`.
- `Value` (**line 213**): a payload tagged with the version vector that wrote it.
- `_merge_into_antichain` (**line 238**): incremental merge that keeps only the maximal (pairwise-concurrent) set; `O(incoming × current)`, not a from-scratch recompute.
- `MvRegisterMemory` (**line 280**): implements the `Memory` protocol (`read` / `write` / `cas` / `subscribe`) plus `values(key)` (**line 366**) for the sibling set and `export` / `merge` / `export_all` / `merge_all` (**lines 444 / 474 / 503**) for gossip.

**Registration (2 lines)**
- `packages/nest-core/nest_core/plugins.py:42` — `("memory", "mv_register")` builtin.
- `packages/nest-plugins-reference/pyproject.toml:38` — `nest.plugins.memory` entry point.

**Adversarial validator** — `packages/nest-plugins-reference/nest_plugins_reference/validators/mv_register_validators.py:71`
- `validate_mv_no_concurrent_loss(make_replica, values)`: N replicas each write a distinct value concurrently, gossip all-to-all, then assert every value survives on every replica. **Fails** against `lww_register` (N-1 values lost), **passes** against `mv_register`. Exported from `validators/__init__.py`.

**Trace validator + scenario**
- `packages/nest-core/nest_core/validators.py:1271` — `validate_mv_sibling_preservation`; registered in the `VALIDATORS` table at **line 4229** under `"mv_register_siblings"`.
[...]

PR #108, opened 2026-07-08
[Hackathon] distsys-reliability: or_set memory CRDT + honest-write-liveness red-team of merged lww_register

prokiller67-crypto:
# [Hackathon] distsys-reliability: OR-Set memory CRDT + honest-write-liveness red-team of `lww_register`

**Persona:** `distsys-reliability` — a distributed-systems reliability engineer.
My risk model is not "does it converge on a flaky network"; the merged
`lww_register` already survives that. It is **"do the convergence claims survive
an adversarial replica"** — a peer that speaks the protocol but lies. That gap
is where write-suppression, split-brain, and silent data loss actually come
from in production replicated state.

## Motivation

The reference memory layer has a register (`lww_register`) and a racing dict
(`blackboard`). A register has **no principled delete**: "remove" has to be
modelled as writing a tombstone sentinel and hoping every replica agrees that
value means "gone". A claim/release marketplace — ten agents concurrently
*claiming* and *releasing* slot ids and needing to agree on *which slots are
held* — is a **set**, not a register. It needs add and remove as first-class,
conflict-free operations with a defined answer for concurrent `add`||`remove`.

This PR adds `or_set`: a state-based **observed-remove set CvRDT** (Shapiro,
Preguiça, Baquero & Zawirski, "A comprehensive study of Convergent and
Commutative Replicated Data Types", INRIA RR-7506, 2011, §3.3.5). Every `add`
mints a unique `(node_id, counter)` tag; every `remove` tombstones only the tags
the remover has *observed*; an element is present iff it carries an add-tag that
is not tombstoned. Merge is the pairwise union of the add and tombstone sets —
commutative, associative, idempotent — and concurrent `add`||`remove` resolves
**add-wins**.

## Threat model

Two distinct fault classes, and the reference plugins only defend the first:

- **Network faults** (drop, reorder, duplicate). `lww_register` handles these:
  its merge is a semilattice join, so honest replicas converge regardless of
  delivery order. `validate_crdt_convergence` already proves it.
- **A Byzantine replica** — honest-looking, protocol-speaking, but exporting
  *forged* state. This is the reliability engineer's real adversary, and no
  shipped validator tests for it. My headline contribution is a validator that
  does, and that **fails a plugin already merged into `main`**.

## The finding: `lww_register.py:305`

```python
# packages/nest-plugins-reference/nest_plugins_reference/memory/lww_register.py:305
self._clock = max(self._clock, incoming.lamport)
```

`merge()` blindly adopts any incoming Lamport clock, and `Register.dominates`
picks the higher `(lamport, node, payload)` as the winner. So a Byzantine
replica exports a register forged with `lamport = 2**60`. Any honest write made
**before** an honest replica merges that forgery carries an ordinary small
clock; once states gossip to convergence the forged register wins and the honest
write is **silently discarded** — write-suppression across the entire
propagation window. Verified repro (also pinned as a test in
`test_or_set_adversarial.py::TestVerifiedVulnerabilityRepro`):

```python
honest = LwwRegisterMemory("honest")
await honest.write("slot", b"honest-claim")   # legitimate write, lamport = 1
await honest.merge("slot", forged)            # forged register, lamport = 2**60
assert await honest.read("slot") == b"BYZANTINE"   # the honest write is gone
```

`or_set` closes the gap structurally: **there is no global clock to forge.** A
Byzantine replica can inflate its own tag counters as high as it likes, but that
only mints Byzantine-owned tags for Byzantine-owned elements. To suppress an
honest claim it would have to tombstone an add-tag it never observed — which it
cannot, because tags are `(node_id, counter)` and the honest counter is small
and unpredictable to the attacker. Its fabricated tombstones (`(claimant-k,
2**60)`) target real honest nodes but with counters no honest add ever used, so
they tombstone nothing.

## Design & tradeoffs

- **Classic tag-set OR-Set, not the optimized (Bieniusa et al.) variant.** The
  optimized OR-Set compresses per-element tags into a per-node version vector
  and drops the explicit tombstone set. I deliberately chose the **classic
  tag-set** construction: every tag and every tombstone is individually visible
  in the exported JSON, so a converged replica's state is **auditable** — you
  can point at exactly which `(node, counter)` add is holding a slot and which
  tombstone tried to remove it. For a reliability tool whose whole job is
  *explaining* what a replicated store did under attack, auditability beats the
  constant-factor space win.
- **Tombstone growth** is the honest cost of that choice: the `removed` set
  grows monotonically with releases. That is fine for a bounded hackathon
  scenario, and I document it as **future work** (a causal-stability GC that
  drops tombstones once every replica has observed the corresponding add — the
  standard OR-Set garbage-collection story).
- **Structural, not random, tags.** The charter's headline anti-pattern is
  nondeterminism smuggled in via `uuid4` / `time.time()` / an unseeded RNG. Tags
  are `(node_id, per-node monotone counter)`, so uniqueness is structural and
  the plugin replays byte-for-byte under a fixed seed.
- **`write` op format + documented fallback.** `write` parses
  `{"op": "add"|"remove", "element": <json>}`; a non-op byte payload is treated
  as `add` of the UTF-8 string (drop-in for opaque-byte callers). A `remove`
  therefore always requires the explicit op form — an ambiguous byte string
  never silently deletes.
- **`read` is byte-deterministic.** It returns the present-element list as
  canonical JSON, sorted by canonical element key, so two converged replicas
  emit identical bytes (what the trace-level validator compares).
- **Observed-context CAS.** `cas(expected, new)` compares *causal context*, not
  a single value: it succeeds iff the caller's observed tag-set still covers the
  key's current tag-set, then applies `new`. Because merge unions add-tags, an
  admitted update is never lost and an unobserved concurrent add is never
  clobbered — lost-update-freedom, pinned by a Hypothesis property against a
  blackboard that races.

## Runnable verification

```bash
# 1. Run the scenario (10 honest claimants + 1 Byzantine forger, 10% drop):
nest run scenarios/memory_orset_claims.yaml

# 2. Validate the trace — convergence, honest-liveness, attacker-actually-ran:
python -c "
from pathlib import Path
for r in validate_trace(Path('traces/memory_orset_claims.jsonl'), 'memory_orset_claims'):
"
```

Output (identical under seeds 42, 7, 1337; trace is byte-deterministic):

```
PASS orset_claims_convergence     - all 11 replicas converged to identical state
PASS orset_claims_honest_liveness - 10 honest claim(s) survived the Byzantine attack: [slot-0 … slot-9]
PASS orset_claims_attacker_ran    - Byzantine forged element observed in converged state (attack actually ran)
PASS memory_liveness              - all 11 replicas reported a final state
```

The three adversarial validators discriminate exactly as the charter asks
(`test_or_set_adversarial.py`):

| validator                             | `lww_register` | `blackboard` | `or_set` |
|---------------------------------------|:--------------:|:------------:|:--------:|
| `validate_memory_honest_write_liveness` | **FAIL**     | **FAIL**     | **PASS** |
| `validate_crdt_add_wins_convergence`    | —            | **FAIL**     | **PASS** |

`validate_memory_honest_write_liveness` **fails a merged plugin** — that is the
point.

## Relationship to merged #22 and open #94 / #76

- **#22 (`memory_concurrent_writers` / `lww_register`, merged):** same layer,
  *different data type and different conflict semantics*. #22 is a
  last-writer-wins **register**; this is an observed-remove **set** with add-wins
  resolution and a real delete. I reuse #22's scenario shape (per-agent replica
  overrides, gossip rounds, `final:` records) and its `validate_memory_liveness`,
  and I do **not** touch its files or reuse its scenario filename. Crucially, I
  ship a validator (`validate_memory_honest_write_liveness`) that #22's
  `lww_register` **fails** — I catch a fault its design does not defend against.
- **#94 / #76 (open memory-layer PRs):** these are also memory-layer work with
  different data types and conflict rules (e.g. a semantic/embedding store).
  This PR is orthogonal: a different CRDT (state-based OR-Set), a different
  conflict rule (observed-remove add-wins), and — the reason it stands alone — a
  Byzantine honest-write-liveness validator that their register/similarity
  designs fail for the same reason `lww_register` does (they resolve conflicts by
  an attacker-influenceable scalar rather than by a union that cannot forge a
  removal).

## Test summary

49 new tests, all green:

- **29 example-based** (`test_or_set.py`): empty/single-element sets,
  add/remove/re-add, self-merge no-op, idempotent + order-independent merge,
  cross-key isolation, structured-op parsing + plain-bytes fallback,
  observed-context CAS (fresh/stale/malformed), and malformed-merge rejection
  without state corruption (non-JSON, wrong CRDT tag, bad tag, bool-counter).
- **7 property-based** (`test_or_set_properties.py`, Hypothesis): merge
  commutativity / associativity / idempotence; permutation invariance
  (any merge order → identical canonical bytes); add-wins under arbitrary
  interleavings; observed-context CAS never loses an update — with a blackboard
  contrast that provably does.
- **7 adversarial** (`test_or_set_adversarial.py`): the clock-forgery matrix
  (FAIL `lww_register` / FAIL `blackboard` / PASS `or_set`), the add-wins matrix,
  and the direct `lww_register.py:305` repro.
- **6 end-to-end** (`test_orset_scenario.py`): the scenario passes all validators
  and is byte-identical across seeds 42 / 7 / 1337; all ten honest claims
  survive; the attacker provably ran; and the validators fail against a
  non-OR-Set trace.

## CI status

`make ci-local` (uv sync, ruff check, ruff format --check, pyright strict,
pytest -v): **840 passed, 1 skipped — all 5 checks green.** No edits to other
participants' plugins, validators' logic, scenarios, simulator internals, the
rubric, seed bank, or scoreboard. Shared-file edits are purely additive: one
`_BUILTINS` row, one `pyproject.toml` entry point, one `scenarios.py` lazy-load
branch, new validator functions + one `VALIDATORS` dict entry (and the one-line
registry-test update every prior scenario PR made).

### DIFF
--- a/docs/layers/memory.md
+++ b/docs/layers/memory.md
@@ -80,6 +80,80 @@ for r in validate_trace(Path('traces/memory_concurrent_writers.jsonl'), 'memory_
 
 The trace is byte-identical under seeds 42, 7, and 1337.
 
+## CRDT plugin: `or_set`
+
+`or_set` -- a state-based **observed-remove set CvRDT** (Shapiro, Preguica,
+Baquero & Zawirski 2011, §3.3.5). Where `lww_register` gives each key a single
+last-writer-wins *value*, `or_set` gives each key a *set* with principled add
+**and** remove: exactly what a claim/release marketplace needs. Every `add`
+mints a unique `(node_id, counter)` tag; every `remove` tombstones only the tags
+the remover has *observed*; an element is present iff it has an add-tag that is
+not tombstoned. Merge is the pairwise union of the add and tombstone sets
+(commutative, associative, idempotent), and concurrent `add`||`remove` resolves
+**add-wins**.
+
+Source: [`nest_plugins_reference/memory/or_set.py`](../../packages/nest-plugins-reference/nest_plugins_reference/memory/or_set.py).
+
+`read` returns the present-element list as canonical JSON (sorted,
+byte-deterministic); `write` takes a structured op `{"op": "add"|"remove",
+"element": <json>}`, with a plain-bytes fallback treated as an add. The per-key
+state stays grep-able in a trace::
+
+    {"crdt": "or_set", "adds": {"\"slot-1\"": [["agent-0", 1]]}, "removed": [["agent-1", 4]]}
+
+```python
+a = OrSetMemory("a")
+b = OrSetMemory("b")
+await a.write("held", b'{"op": "add", "element": "slot-1"}')
+await b.write("held", b'{"op": "add", "element": "slot-2"}')
+await b.merge("held", a.export("held"))   # gossip a -> b
+await a.merge("held", b.export("held"))   # gossip b -> a
+assert await a.read("held") == await b.read("held")   # converged, any order
+```
+
+### Why an OR-Set: a Byzantine-liveness gap `lww_register` has
+
+`lww_register.merge` adopts the higher Lamport clock
+(`lww_register.py:305`, `self._clock = max(self._clock, incoming.lamport)`), so
+a Byzantine replica that exports a register forged with `lamport = 2**60`
+silently suppresses every honest write with a smaller clock. An OR-Set has no
+global clock to forge: a Byzantine replica can inflate its own tag counters, but
+that only mints Byzantine-owned tags for Byzantine-owned elements -- it cannot
+tombstone an add-tag it never observed, so it cannot suppress an honest claim.
+
+`nest_core.validators` ships three checks for `or_set`:
+
+- `validate_memory_honest_write_liveness(make_replica, forge=..., honest_op=...,
+  is_visible=...)` -- a Byzantine replica forges state, an honest replica writes
+  before observing it, both gossip to convergence. It **fails** `lww_register`
+  (forged clock wins) and `blackboard` (later clobber), and **passes** `or_set`.
+- `validate_crdt_add_wins_convergence(make_replica, add_op=..., remove_op=...,
+  present=...)` -- concurrent `add`||`remove` on one element must converge
+  byte-identically *and* resolve add-wins. **Fails** `blackboard`, **passes**
+  `or_set`.
+- `validate_orset_claims_convergence` / `validate_orset_claims_honest_liveness`
+  -- registered for the `memory_orset_claims` scenario; confirm every replica's
+  final state is identical and every honest claim survived the attack.
+
+### Demo scenario
+
+`scenarios/memory_orset_claims.yaml` -- 10 honest replicas claim/release slot
+ids in one shared OR-Set under 10% message drop while **one Byzantine replica**
+injects forged, inflated-counter state throughout the run:
+
+```bash
+nest run scenarios/memory_orset_claims.yaml
+python -c "
+from pathlib import Path
+from nest_core.validators import validate_trace
+for r in validate_trace(Path('traces/memory_orset_claims.jsonl'), 'memory_orset_claims'):
+    print(('PASS' if r.passed else 'FAIL'), r.name, '-', r.detail)
+"
+```
+
+Every replica converges and all ten honest claims survive; the trace is
+byte-identical under seeds 42, 7, and 1337.
+
 ## Writing your own
 
 See [`writing-a-plugin.md`](../writing-a-plugin.md). Register under
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -39,6 +39,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("negotiation", "pareto"): f"{_REF}.negotiation.pareto:ParetoNegotiation",
     ("memory", "blackboard"): f"{_REF}.memory.blackboard:Blackboard",
     ("memory", "lww_register"): f"{_REF}.memory.lww_register:LwwRegisterMemory",
+    ("memory", "or_set"): f"{_REF}.memory.or_set:OrSetMemory",
     ("privacy", "noop"): f"{_REF}.privacy.noop:NoopPrivacy",
     ("privacy", "hybrid_x25519"): f"{_REF}.privacy.hybrid_x25519:HybridX25519Privacy",
     ("datafacts", "datafacts_v1"): f"{_REF}.datafacts.datafacts_v1:DataFactsV1",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -87,6 +87,12 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("memory_concurrent_writers", memory_concurrent_writers_factory)
+    elif name == "memory_orset_claims":
+        from nest_core.scenarios_builtin.memory_orset_claims import (
+            memory_orset_claims_factory,
+        )
+
+        register_scenario("memory_orset_claims", memory_orset_claims_factory)
     elif name == "comms_versioning":
         from nest_core.scenarios_builtin.comms_versioning import comms_versioning_factory
 
--- a/packages/nest-core/nest_core/scenarios_builtin/memory_orset_claims.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/memory_orset_claims.py
@@ -1,0 +1,302 @@
+# SPDX-License-Identifier: Apache-2.0
+"""OR-Set claim/release marketplace -- convergence under a Byzantine forger.
+
+A task-claim marketplace: ``claimant-0 .. claimant-{N-1}`` each own a private
+OR-Set replica of one shared key (``claims``) and concurrently *claim* (add) and
+*release* (remove) slot ids while gossiping their replica to convergence under
+10% message loss. Alongside them runs exactly **one** Byzantine replica,
+``byzantine-0``, that injects a forged, inflated-counter state throughout the
+run -- the OR-Set analogue of the ``lww_register`` clock-forgery attack (a
+register forged with ``lamport = 2**60``).
+
+The forgery does two adversarial things at once:
+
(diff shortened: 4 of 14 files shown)


dhve:
Thanks @prokiller67-crypto. This is one of the most carefully engineered memory-layer submissions I've read. The `lww_register.py:305` clock-adoption finding is real (I verified `self._clock = max(self._clock, incoming.lamport)` on main, and `TestVerifiedVulnerabilityRepro` pins it cleanly), the classic tag-set OR-Set in `or_set.py` is textbook on the algebra with a genuine Hypothesis suite (commutativity, associativity, idempotence, permutation invariance), and shipping `validate_memory_honest_write_liveness` as a validator that fails a plugin already merged into main is exactly what the charter asks for. The scenario also mirrors the merged `memory_concurrent_writers` shape (per-agent overrides via `_agent_plugins`, `final:` records, `validate_memory_liveness` reuse) instead of inventing a parallel universe.

**1. The headline liveness claim falls to a predicted-counter tombstone forgery.**

The PR body argues a Byzantine replica cannot suppress an honest claim "because tags are `(node_id, counter)` and the honest counter is small and unpredictable to the attacker." It is small, but it is not unpredictable: node ids are the public agent ids and counters go 1, 2, 3 per local add. `OrSetMemory.merge` unions `incoming.removed` unconditionally (`target.removed |= incoming.removed`), `removed` is a flat tag set with no element binding, and a forger can always ship a self-consistent add+tombstone pair, so schema validation cannot save you. Concretely:

```python
honest = OrSetMemory("claimant-3")
await honest.write("claims", b'{"op": "add", "element": "slot-3"}')  # mints ("claimant-3", 1)
forged = json.dumps({"crdt": "or_set", "adds": {},
                     "removed": [["claimant-3", c] for c in range(1, 33)]}).encode()
await honest.merge("claims", forged)
assert await honest.read("claims") == b"[]"  # honest claim suppressed
```

`_advance_counter` only protects adds made after the merge; every add in the pre-observation window, which is precisely the window your liveness property is about, dies. Your scenario passes because `_forged_state` uses `FORGED_COUNTER = 2**60`, a deliberately weak adversary; forge tombstones over counters 1..32 for every claimant instead and I believe `orset_claims_honest_liveness` fails with zero survivors. Since `validate_memory_honest_write_liveness` is parameterized by `forge`, add this stronger forge to the adversarial matrix and it should fail `or_set` too. Suggested fix: mint tag counters from the simulator's per-agent seeded RNG (the problem statement's pointer about `_failure_rng` and seeding logical clocks is exactly this), so tags stay byte-deterministic under a fixed seed but unpredictable in-model, then rescope the docstring and PR-body claims to what actually holds. This is the correctness of the submission's centerpiece, so it is worth getting exactly right.

**2. Branch name does not match the charter pattern.**

The charter requires `hackathon/<your-handle>-<short-theme>` and states non-matching PRs are auto-closed. `hackathon/distsys-reliability-orset-liveness` carries the persona, not your handle. Rename the branch on your fork (e.g. `hackathon/prokiller67-crypto-orset-liveness`); GitHub keeps the PR attached.

**3. `cas` silently breaks the protocol's documented contract for base callers.**

`nest_core/layers/memory.py` documents `cas` as "update only if the current value matches expected," and `lww_register.cas` honors that (it compares the winning payload, i.e. what `read` returns). Your observed-context `cas` requires `expected` to be an `export()` blob: a base-protocol caller doing `expected = await mem.read(key)` then `cas(key, expected, op)` gets `False` unconditionally (`read` returns a JSON list, `_decode` raises `OrSetStateError`, you return `False`), so a standard read-then-cas retry loop spins forever. Problem 02 does bless implementing `cas` "in terms of the CRDT's natural merge semantics," so keep the causal-context path, but fall back to comparing `expected` against the current `read()` bytes when it does not decode as an or_set export, and document both modes in `docs/layers/memory.md` (your doc section currently does not mention `cas` at all).

**4. The scenario liveness validator tolerates 9 of 10 claims being suppressed.**

`validate_orset_claims_honest_liveness` passes on `bool(honest_claims)`, i.e. a single surviving `slot-*`, and only decodes `finals[0]`. Your e2e test compensates by substring-matching all ten slots inside the human-readable `detail` string, which is brittle. Derive the expected set from the trace's `start` events (every `claimant-i` implies `slot-i`), require all of them, and intersect present-elements across all final records rather than trusting the first.

**Suggested path forward:**

1. Add the predicted-counter forge to `test_or_set_adversarial.py`, confirm it fails `or_set` today, then fix tag minting (seeded per-agent RNG) so it passes, and rescope the claims in `or_set.py`, `docs/layers/memory.md`, and the PR body.
2. Rename the branch to include your handle.
3. Dual-mode `cas` plus a doc line, with a test covering the read-then-cas caller.
4. Tighten `validate_orset_claims_honest_liveness`; drop the detail-string assertion in `test_orset_scenario.py`.
5. Nit: `ClaimantAgent._num_slots` is stored but never used.
6. No CI checks have reported on this branch yet (likely awaiting first-run approval). The judge panel does not score broken PRs, so confirm the run goes green after your next push.

Worth saying: this space is crowded (merged #22, plus open CRDT-memory PRs #76, #113, #124, #137), and your differentiator is the Byzantine red-team validator. Fixing point 1 is what makes that differentiator real rather than a demo against a chosen-weak adversary; done right, this is a strong submission. Leaving this open. Ping me on the hackathon channel if the tag-minting tradeoff is unclear.

JamesCarnley:
Thank you, @prokiller67-crypto. This contribution correctly highlights that convergence to absence is not honest-write liveness. A future add-wins set must preserve every expected honest add under duplicate/reordered delivery and adversarial tombstones targeting predicted tags, with authenticated tag ownership and malformed-state atomicity. Current Town memory has no replica, merge, tombstone, CAS, or subscription contract, so there is no current OR-Set defect to repair. The frozen implementation also does not withstand the predicted-tag tombstone case. I am closing the legacy PR while retaining that falsifier and the full convergence/liveness matrix for a future replicated-memory profile. No contributor code was merged or executed during this review.

PR #109, opened 2026-07-08
hackathon/tanisha: implemented a custom Rate Limiter layer and reference plugins

tanishaacodes:
# Summary
Implementation of a custom **Rate Limiting Layer** (introducing the `rate_limit` layer, `TokenBucketRateLimiter`, and `LeakyBucketRateLimiter` reference plugins) to enforce traffic and request boundaries per agent.

All changes are fully verified, formatted, typechecked, and pass the local CI checklist.

---

### Layer Definition
* `packages/nest-core/nest_core/layers/rate_limiter.py`:
  - Defines the `RateLimiter` Protocol interface.
  - Exposes `consume(client, *, now, tokens)` and `allowance(client, *, now)`. All calculations take logical simulation time `now` as a keyword-only argument to ensure determinism.
* `packages/nest-core/nest_core/layers/__init__.py`:
  - Imports and exports `RateLimiter`.

### Reference Plugins
* `packages/nest-plugins-reference/nest_plugins_reference/rate_limit/token_bucket.py`:
  - Class `TokenBucketRateLimiter` refilling tokens linearly at a constant rate up to a max capacity limit.
* `packages/nest-plugins-reference/nest_plugins_reference/rate_limit/leaky_bucket.py`:
  - Class `LeakyBucketRateLimiter` modeling a queue/water level that leaks constant capacity over logical time.
* `packages/nest-plugins-reference/nest_plugins_reference/rate_limit/__init__.py`:
  - Exports both plugins.
* `packages/nest-core/nest_core/plugins.py`:
  - Registers the custom plugins in the built-in registry (`("rate_limit", "token_bucket")` and `("rate_limit", "leaky_bucket")`).

### Unit Tests
* `packages/nest-plugins-reference/tests/test_rate_limiter.py`:
  - Unit tests verifying token replenishment, capacity limits, drop behavior, and independent client rate isolation.

---

## 2. Verification

Run local checks:
```bash
# Run rate limiter unit tests
uv run pytest packages/nest-plugins-reference/tests/test_rate_limiter.py -v

# Run full project checks
make ci-local

### DIFF
--- a/packages/nest-core/nest_core/layers/__init__.py
+++ b/packages/nest-core/nest_core/layers/__init__.py
@@ -16,6 +16,7 @@ from nest_core.layers.memory import Memory
 from nest_core.layers.negotiation import Negotiation
 from nest_core.layers.payments import Payments
 from nest_core.layers.privacy import Privacy
+from nest_core.layers.rate_limiter import RateLimiter
 from nest_core.layers.registry import Registry
 from nest_core.layers.transport import Transport
 from nest_core.layers.trust import Trust
@@ -31,6 +32,7 @@ __all__ = [
     "Negotiation",
     "Payments",
     "Privacy",
+    "RateLimiter",
     "Registry",
     "Transport",
     "Trust",
--- a/packages/nest-core/nest_core/layers/rate_limiter.py
+++ b/packages/nest-core/nest_core/layers/rate_limiter.py
@@ -1,0 +1,49 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Rate-limiter layer interface: tracks and bounds traffic consumption per agent.
+
+A rate limiter provides a traffic oracle. It tracks request allowances for clients
+over logical simulation time 'now' and enforces limits to prevent message spam or
+resource starvation.
+
+Every method takes 'now' (the caller's virtual simulation time) as a keyword-only
+argument to ensure that traffic metrics and rate limiting remain fully deterministic
+and trace-reproducible.
+"""
+
+from __future__ import annotations
+
+from typing import Protocol, runtime_checkable
+
+from nest_core.types import AgentId
+
+
+@runtime_checkable
+class RateLimiter(Protocol):
+    """Traffic oracle: tracks request counts and enforces rate limits.
+
+    Example::
+
+        limiter: RateLimiter = TokenBucketRateLimiter(rate=2.0, capacity=10.0)
+        allowed = await limiter.consume(AgentId("agent-1"), now=ctx.time, tokens=1.0)
+    """
+
+    async def consume(self, client: AgentId, *, now: float, tokens: float = 1.0) -> bool:
+        """Attempt to consume 'tokens' for a client at logical time 'now'.
+
+        Returns True if the request is permitted, or False if the client is rate-limited.
+
+        Example::
+
+            if not await limiter.consume(AgentId("sender"), now=ctx.time):
+                raise Exception("Rate limit exceeded")
+        """
+        ...
+
+    async def allowance(self, client: AgentId, *, now: float) -> float:
+        """Return the current available token allowance for the client at time 'now'.
+
+        Example::
+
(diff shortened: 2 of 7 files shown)


dhve:
Thanks @tanishaacodes. The `RateLimiter` Protocol design is the strongest part of this PR: making `now` a keyword-only argument on both `consume` and `allowance` so all refill math runs on logical simulation time is exactly the right determinism posture for this codebase, and the two implementations (`TokenBucketRateLimiter`, `LeakyBucketRateLimiter`) are clean, symmetric, and properly per-client isolated (nice touch verifying that in `test_independent_clients`).

**1. This does not map to any of the ten problem briefs, and you have three open PRs.**

The charter says pick exactly one problem from `docs/hackathon/problems/` and ship a single PR. Rate limiting is not one of the ten, and you also have #104 (capability tokens) and #122 (simulation sandbox) open. The judge panel scores against a problem brief, so right now this PR has no rubric anchor. Please pick your primary submission and note it on the others. If this is the one, say in the PR body which problem's success criteria you are targeting, or how it composes with one of the ten (for example as the anti-spam guard in a comms or coordination scenario).

**2. Both mandatory deliverables are missing: adversarial validator and scenario YAML.**

Every submission must include an adversarial validator that catches an attack class the baseline plugin would fail, plus a scenario under `scenarios/` demonstrating your plugin passing under that validator. This PR has neither; it is interface plus two plugins plus unit tests. `scenarios/failure_detection.yaml` is the closest precedent to copy, since #46 also introduced a brand-new layer.

**3. Correctness bug: both limiters mint free capacity when `now` goes backwards.**

In `TokenBucketRateLimiter._update_tokens` you clamp `elapsed = max(0.0, now - last)` but then unconditionally set `self._last_updated[client] = now`. A rewound `now` credits no refill yet moves the baseline backwards, so the same interval gets re-counted on the next call. Concretely, with `rate=1.0, capacity=5.0`: drain the bucket at `now=10`, call `consume(client, now=0.0)` (denied, but the timestamp rewinds), then `consume(client, now=10.0, tokens=5.0)` succeeds. Repeat forever: unlimited tokens with logical time frozen at 10. `LeakyBucketRateLimiter._leak` has the same defect (water re-leaks). The fix is one line, `self._last_updated[client] = max(last, now)`, and the attack is a gift-wrapped candidate for your missing adversarial validator: a client replaying stale timestamps that a naive limiter admits and yours rejects.

**4. Entry-point wiring is incomplete for a new layer (api fit).**

Registering both plugins in `_BUILTINS` in `nest_core/plugins.py` makes `resolve()` work, but a new layer needs two more hooks: `"rate_limit"` added to the layer list in `_discover_entry_points` (it is absent, so no third-party rate_limit plugin can ever be discovered), and a `[project.entry-points."nest.plugins.rate_limit"]` table in `packages/nest-plugins-reference/pyproject.toml` alongside the existing groups. The `failure_detector` layer did all three; mirror it.

**5. Tests are example-only.**

What is there is correct, but sibling suites like `test_bonded_trust.py` use hypothesis. The natural property here: for any sequence of consumes with non-decreasing `now`, total granted tokens never exceeds `capacity + rate * elapsed`. Extend the `now` strategy to arbitrary (non-monotone) floats and that one test also catches bug 3 automatically.

**Suggested path forward:**

1. Decide which of your three PRs is the primary and note it on the others.
2. If it is this one, anchor it to a problem brief in the PR body (motivation, threat model, tradeoffs).
3. Fix the timestamp rewind in both buckets and add the hypothesis property test.
4. Add the time-rewind adversarial validator and a `scenarios/rate_limit_*.yaml` showing your limiter passing where a naive one fails.
5. Wire the entry points (discovery list plus pyproject table).
6. CI has not run on this branch yet. Run `make ci-local` and confirm all stages green; the judge panel does not score broken PRs.

Leaving this open. Ping me on the hackathon channel if the problem-mapping question is unclear.

PR #110, opened 2026-07-08
Add policy decision layer for agent actions

aparey:
## Problem

NANDA Town has layers for identity, auth, privacy, payments, trust, memory, coordination, and failure detection, but it does not yet have a small deterministic policy decision point for agent actions.

Before an agent publishes data, spends credits, calls an external tool, or changes authority, it should be able to ask: is this action permitted, denied, or does it need approval?

This PR adds that missing policy layer.

## What is in the diff

| Path | Purpose |
| --- | --- |
| `packages/nest-core/nest_core/layers/policy.py` | New `Policy` Protocol plus `PolicyRequest`, `PolicyDecision`, and `PolicyEffect`. |
| `packages/nest-plugins-reference/nest_plugins_reference/policy/strict.py` | Reference deny-by-default policy. Permits safe public reads, denies sensitive public exports, requires approval for high-value payments, and denies unknown actions. |
| `packages/nest-plugins-reference/nest_plugins_reference/policy/allow_all.py` | Unsafe baseline plugin used as the foil. |
| `packages/nest-core/nest_core/scenarios_builtin/policy_guard.py` | Deterministic adversarial scenario that exercises safe, sensitive, high-value, and unknown actions. |
| `scenarios/policy_guard.yaml` | Runnable scenario config. |
| `packages/nest-core/nest_core/validators.py` | Four policy validators registered under `policy_guard`. |
| `packages/nest-plugins-reference/tests/test_policy.py` | Unit, integration, adversarial, and determinism tests. |
| `docs/layers/policy.md` | Documentation and usage notes. |

## Why this is useful

It fills a practical security gap. Auth answers “who are you?” Privacy answers “who can read this?” Payments answer “can value move?” A policy layer answers “should this agent be allowed to do this action right now?”

It is intentionally small and composable. The layer does not execute actions or replace other layers. It only returns a replayable decision that other layers or agents can respect.

It is testable, not just illustrative. The `strict_rules` plugin passes the scenario validators. The `allow_all` baseline fails the restrictive checks, proving the validators catch unsafe behavior instead of merely checking that messages exist.

## Verification

Ran the full local CI gate:

```bash
make ci-local

### DIFF
--- a/docs/layers/policy.md
+++ b/docs/layers/policy.md
@@ -1,0 +1,58 @@
+# Policy
+
+The policy layer is the decision point an agent calls before it uses a tool,
+publishes data, spends credits, or changes authority.  It turns a structured
+`PolicyRequest` into a deterministic `PolicyDecision`: `permit`, `deny`, or
+`approval_required`.
+
+This layer is intentionally small.  It does not execute the action and it does
+not replace auth, privacy, payments, or trust.  It answers one question: given
+the actor, action, resource, data classes, and optional amount, may the agent
+continue?
+
+## Interface
+
+Implementations satisfy `nest_core.layers.policy.Policy`:
+
+```python
+decision = await policy.decide(request, now=ctx.time)
+```
+
+`now` is the simulator's logical time.  Implementations should not read wall
+clock time, random sources, or process-local mutable state unless that state is
+derived from previous replayable observations.
+
+## Reference Plugins
+
+`strict_rules` is deny-by-default.  It permits declared low-risk reads, denies
+sensitive public exports, requires approval for high-value transfers, and denies
+unknown actions.
+
+`allow_all` is a deliberately unsafe baseline.  The `policy_guard` scenario uses
+it as the foil: the same trace validators that pass under `strict_rules` fail
+under `allow_all`.
+
+## Scenario
+
+Run the built-in scenario:
+
+```bash
+uv run nest run scenarios/policy_guard.yaml
+```
+
+Then validate the trace:
+
+```bash
+uv run python - <<'PY'
+from pathlib import Path
+from nest_core.validators import validate_trace
+
+for result in validate_trace(Path("traces/policy_guard.jsonl"), "policy_guard"):
+    print(result)
+PY
+```
+
+The scenario emits canonical JSON request and decision events for a safe read, a
+sensitive public export, a high-value payment, and an undeclared admin action.
+Validators assert that the policy permits the safe read, blocks public sensitive
+data, requires approval for high-impact spend, and denies undeclared actions.
--- a/packages/nest-core/nest_core/layers/__init__.py
+++ b/packages/nest-core/nest_core/layers/__init__.py
@@ -15,6 +15,7 @@ from nest_core.layers.identity import Identity
 from nest_core.layers.memory import Memory
 from nest_core.layers.negotiation import Negotiation
 from nest_core.layers.payments import Payments
+from nest_core.layers.policy import Policy, PolicyDecision, PolicyEffect, PolicyRequest
 from nest_core.layers.privacy import Privacy
 from nest_core.layers.registry import Registry
 from nest_core.layers.transport import Transport
@@ -30,6 +31,10 @@ __all__ = [
     "Memory",
     "Negotiation",
     "Payments",
+    "Policy",
+    "PolicyDecision",
+    "PolicyEffect",
+    "PolicyRequest",
     "Privacy",
     "Registry",
     "Transport",
--- a/packages/nest-core/nest_core/layers/policy.py
+++ b/packages/nest-core/nest_core/layers/policy.py
@@ -1,0 +1,105 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Policy layer interface: deterministic action decisions for agent tools.
+
+Agents need a small, replayable gate before they call tools, spend money, or
+publish data outside the town.  A policy implementation receives a structured
+request and returns one of three effects:
+
+* ``permit`` -- the action may proceed.
+* ``deny`` -- the action must not be attempted.
+* ``approval_required`` -- the action is allowed only after an external approval.
+
+The request is plain data and every decision takes ``now`` as a keyword-only
+argument, so policy checks remain deterministic under simulator replay.
+
+Example::
+
+    request = PolicyRequest(actor=AgentId("a1"), action="read", resource="catalog/public")
+    decision = await policy.decide(request, now=ctx.time)
+"""
+
+from __future__ import annotations
+
+import enum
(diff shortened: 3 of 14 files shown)

### CODE REVIEW

**Comment by aparey** [comment]
PTAL Thanks in advance!

**Comment by dhve** [comment]
Thanks @aparey. This is one of the cleanest-built submissions in the queue: `strict.py` is a genuinely tidy deny-by-default engine with sensible rule ordering (sensitive-public export first, then spend thresholds, then the high-impact catch-all, then declared safe reads, then default deny), `test_policy_guard_allow_all_fails_adversarial_validators` proves your validators can actually fail instead of rubber-stamping, and the byte-determinism test plus SPDX headers, `from __future__ import annotations`, and `Example::` blocks everywhere show you read the house style closely.

**1. This is an off-menu problem, and the core footprint is larger than it needs to be.**

The charter asks for exactly one problem from `docs/hackathon/problems/`, and a policy layer is not one of the ten. There is precedent for an off-menu layer (the merged failure-detector work), but note how it landed on the core side: it added the Protocol, the `_BUILTINS` entries, and the entry-point discovery line, and stopped there. It never touched `LayerConfig` or `runner._resolve_plugins`; the plugin choice comes from `task.config` (`fd_plugin`) inside the scenario factory, injected per-agent via `plugins["_agent_plugins"]`. Your diff instead edits `scenario.py` and `runner.py`, which changes plugin resolution for every existing scenario, and the drive-by `failure_detector: str = "phi_accrual"` field you added to `LayerConfig` creates a second config surface that the existing failure_detection scenario ignores (it reads `task.config`). Suggest reverting the `scenario.py` and `runner.py` hunks entirely and selecting the policy plugin via `task.config` in `policy_guard_factory`, mirroring `fd_plugin`. Also worth saying plainly in the PR body that this is an off-menu layer proposal; CONTRIBUTING.md asks for an issue first for new layers, so open one and link it.

**2. A payment with no declared amount is autonomously permitted.**

In `StrictPolicy.decide`, `amount = request.amount or 0` means an adversarial agent that simply omits `amount` on a `pay` or `transfer` request gets `permit-low-value-transfer`. That is the exact evasion class this layer exists to stop, and neither the tests nor the `policy_guard` scenario cover it (your scenario only probes `amount=900`). Treat `None` as unknown and return `approval_required` (or deny), add a unit test, and consider a fifth scenario case plus validator for it. A hypothesis property here would also lift test_rigor: for all amounts above a permitted one, the decision is never more permissive.

**3. The branch conflicts with main and CI never ran on the head commit.**

`mergeable` is CONFLICTING right now. `validators.py` and `test_validators.py` are hot files that nearly every merged PR appends to, so this will keep happening; rebase onto latest main and re-resolve. There are also zero check runs on your head SHA, and the judge panel does not score broken PRs, so getting a green run up is step one.

**4. Two small cleanups.**

In `policy_guard_factory`, the first branch (`if policy_cls is not None and not isinstance(policy_cls, type): plugins["policy"] = policy_cls`) assigns the value back to itself; collapse to a single `isinstance(policy_cls, type)` instantiation. And `metrics: [policy_decisions]` in `scenarios/policy_guard.yaml` is a silent no-op: `compute_metrics` skips names missing from `_METRIC_FUNCS` and the PR never registers one. Either implement the metric or drop the entry.

**Suggested path forward:**

1. Rebase onto main and resolve the validators conflicts.
2. Open the CONTRIBUTING-suggested issue framing this as a new-layer proposal, and link it from the PR body.
3. Revert `scenario.py` and `runner.py`; move plugin selection to `task.config` in the factory and update `test_policy.py` accordingly.
4. Close the `amount=None` bypass with a test and a scenario case.
5. Fix the factory no-op branch and the phantom metric.
6. Run `make ci-local` and confirm all stages green before requesting re-review.

Leaving this open. Ping me on the hackathon channel if the footprint reshaping is unclear.

**Comment by aparey** [comment]
Thanks for the detailed review. I pushed a revised head at `11997ae` that addresses the requested shape and safety fixes.

Issue for the off-menu/new-layer proposal: https://github.com/projnanda/nandatown/issues/154

Changes made:

- Rebased onto latest `main`; PR is mergeable again.
- Removed the global `LayerConfig` / `ScenarioRunner` policy resolution footprint.
- Moved policy plugin selection into `policy_guard_factory` via `task.config.policy_plugin`.
- Removed the phantom `metrics: [policy_decisions]` entry.
- Fixed the missing-amount bypass: `pay` / `transfer` with `amount=None` now returns `approval_required`.
- Added a fifth scenario case and validator for `unknown_amount_payment`.
- Added direct unit coverage for the missing amount case.
- Added a monotonic spend-permission test.
- Preserved upstream validator/scenario additions from the rebase.

Verification:

```bash
uv sync --all-packages --all-extras --dev
uv run ruff check .
uv run ruff format --check .
uv run pyright packages scripts
uv run pytest packages/nest-plugins-reference/tests/test_policy.py -q
uv run pytest packages/nest-core/tests/test_validators.py -q
uv run nest run scenarios/policy_guard.yaml
uv run python -c "from pathlib import Path; from nest_core.validators import validate_trace; results=validate_trace(Path('traces/policy_guard.jsonl'), 'policy_guard'); print('\n'.join(f'{r.name}: {r.passed} {r.detail}' for r in results)); raise SystemExit(0 if all(r.passed for r in results) else 1)"
uv run pytest -q
```

Local results:

- `ruff check`: passed
- `ruff format --check`: passed
- `pyright packages scripts`: 0 errors
- focused policy tests: 14 passed
- validator tests: 125 passed
- policy_guard scenario validators: all passed
- full pytest: 1164 passed, 1 skipped, 1 deselected

Note: I used `pyright packages scripts` locally because this checkout contains an unrelated nested `skillguard/` repo that is not part of PR #110; GitHub CI will not have that local untracked directory.

aparey:
Hi @dhve, quick deadline ping for PR #110. I addressed your review: rebased onto main, removed the LayerConfig/runner footprint, moved policy selection into task.config, fixed the amount=None bypass with unit + scenario validator coverage, removed the phantom metric, opened issue #154 for the off-menu layer proposal, and posted verification results. The PR is mergeable now, but checks have not run yet, likely awaiting maintainer workflow approval. Thanks again.

JamesCarnley:
Thank you, @aparey. The reusable requirement is a total, deterministic, replayable policy decision with explicit permit, deny, or approval-required outcomes; missing inputs and unknown actions must fail closed, and a decision must remain separate from approval and effect. Current Town has narrow Run Grant permission checks for join, claim, send, and acknowledge, but no adopted general business-policy layer for exports, payment thresholds, or administration. Importing this ruleset would choose new product policy rather than fix a current defect, so I am closing the legacy PR and retaining its missing-input, threshold, safe-read, and effect-separation cases for a future policy profile. No contributor code was merged or executed during this review.

PR #112, opened 2026-07-08
[Hackathon] Expand behavioral test coverage for the score_average trust plugin

shreyad2806:
## Summary

Adds additional behavioral tests for the reference `score_average` trust plugin.

These tests improve coverage of the existing implementation without changing production behavior.

## Tests Added

- running average calculation
- byzantine evidence handling
- unknown evidence fallback
- confidence progression
- confidence saturation
- placeholder attestation
- stake accumulation
- stake isolation
- score isolation between agents

## Validation

- [x] Existing plugin tests pass
- [x] New tests pass
- [x] Ruff checks pass
- [x] Pyright checks pass

### DIFF
--- a/packages/nest-plugins-reference/tests/test_score_average.py
+++ b/packages/nest-plugins-reference/tests/test_score_average.py
@@ -1,0 +1,132 @@
+# SPDX-License-Identifier: Apache-2.0
+# pyright: reportPrivateUsage=false
+"""Unit tests for score_average trust plugin."""
+
+from __future__ import annotations
+
+import pytest
+from nest_core.types import AgentId, Claim, Evidence
+
+
+class TestScoreCalculation:
+    @pytest.mark.asyncio
+    async def test_score_treats_byzantine_as_negative(self) -> None:
+        from nest_plugins_reference.trust.score_average import ScoreAverageTrust
+
+        trust = ScoreAverageTrust()
+        ev = Evidence(reporter=AgentId("a2"), subject=AgentId("a1"), kind="byzantine")
+        await trust.report(AgentId("a1"), ev)
+
+        score = await trust.score(AgentId("a1"))
+        assert score.score == 0.0
+        assert score.sample_count == 1
+
+    @pytest.mark.asyncio
+    async def test_score_returns_running_average(self) -> None:
+        from nest_plugins_reference.trust.score_average import ScoreAverageTrust
+
+        trust = ScoreAverageTrust()
+        pos = Evidence(reporter=AgentId("a2"), subject=AgentId("a1"), kind="positive")
+        neg = Evidence(reporter=AgentId("a3"), subject=AgentId("a1"), kind="negative")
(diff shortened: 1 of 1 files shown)


dhve:
Thanks @shreyad2806. The tests are clean: I checked every assertion against `score_average.py` and all match the implementation (byzantine mapping to 0.0, the `min(1.0, n/100)` confidence curve, the unknown-kind 0.5 fallback), and the file follows house conventions (SPDX header, `from __future__ import annotations`).

The issue is scope, not quality. As it stands this is not a scoreable hackathon submission. Three things:

**1. It does not map to any problem in `docs/hackathon/problems/`.**

The charter asks you to pick exactly one of the ten problems and ship a plugin, scenario, or validator against its success criteria, plus a mandatory adversarial validator and a scenario YAML under `scenarios/`. A tests-only PR has none of those pieces, so the judge panel has almost nothing to grade on correctness, novelty, or persona fidelity. Also, the charter calls the reference plugins "deliberately simplified testing scaffolding", so a test like `test_attest_returns_placeholder_without_identity` (pinning `b"attestation"` and `algorithm="none"`) freezes placeholder behavior we want free to change.

**2. Some of this coverage already exists.**

`tests/test_plugins.py` has a `TestScoreAverageTrust` class covering the default score, positive reports, and the running average, so `test_score_returns_running_average` retreads it. The new ground here is byzantine handling, the unknown-kind fallback, confidence saturation, and stakes.

**3. Example-only tests score low on test rigor.**

The rubric favors property-based tests and `hypothesis` is already a dependency (see the `*_properties.py` files). One property, "for any mix of positive, negative, and byzantine reports, score equals positives/total and confidence equals min(1, n/100)", would subsume four of these tests.

**Suggested path forward:**

1. Pick one problem from `docs/hackathon/problems/`. Your instincts around byzantine evidence and confidence saturation transfer well to several of them.
2. Reuse this work as the seed of your mandatory adversarial validator and its tests, property-based where you can.
3. Add the scenario YAML showing your plugin passing under that validator.
4. Run the five CI stages locally (`uv sync`, `ruff check`, `ruff format --check`, `pyright`, `pytest -v`). CI has not run on this branch, and the panel does not score red PRs.

Leaving this open. Ping me on the hackathon channel if picking a problem to anchor this to is unclear.

JamesCarnley:
Thank you, @shreyad2806. These tests usefully emphasize explicit handling of unknown evidence kinds and subject-isolated score state. They target the retired score_average implementation, however, including confidence, stake, and a placeholder unsigned attestation that current Town deliberately does not expose. Rebuilt Town uses receipt-attributed +1/-1 reputation and its evaluator now checks prior unique receipts, observer/subject/outcome/order/run binding, exact arithmetic, and unknown outcomes. I am closing this obsolete test-only PR while retaining unknown-kind and subject-isolation cases for any future scoring profile. No contributor code was merged or executed during this review.

PR #113, opened 2026-07-08
[Hackathon] jmuslu: PN-Counter memory with basis-fusion invariants

jmuslu:
## Layer 10 — Memory

This PR adds a `memory:pn_counter` reference plugin, a `memory:basis_gated`
wrapper plugin, and three memory scenarios that test signed evidence
aggregation, basis-restricted memory fusion, and context-saturation rejection.

## Why

Convergence alone is not always the right invariant. A last-writer-wins
register can converge while discarding concurrent evidence. For agent memory,
especially under noisy or adversarial context, we often need a state object
that **preserves contributions** and only acts on information that can **legally
fuse** with the current task structure.

This PR treats memory as an applied mathematical object:

- **PN-Counter memory** preserves signed evidence as positive and negative
  grow-only coordinates. Merge is pointwise max, so it is commutative,
  associative, and idempotent. The derived value is `sum(positive) - sum(negative)`.
- **Basis-gated memory** only accepts evidence that restricts onto declared
  basis dimensions. Natural-language saturation, code-shaped saturation, and
  off-basis structured claims are ignored rather than fused into meaningful state.

The distinction that motivates the whole PR:

- **Convergence** as agreement on state, versus
- **Convergence** as preservation of meaningful evidence under legal fusion rules.

### `memory:pn_counter`

`PnCounterMemory` implements the existing `Memory` protocol (`read`, `write`,
`subscribe`, `cas`) and exposes CRDT-style replication helpers (`export`,
`merge`, `export_all`, `merge_all`).

Writes accept signed deltas such as:

```json
{"op": "inc", "amount": 2}
{"op": "dec", "amount": 1}
{"delta": -3}
```

The implementation rejects malformed state and non-integer coordinates
(including fractional values like `1.5`), so the PN-Counter stays a genuine
integer-coordinate CRDT.

**Self-coordinate defense.** Each replica tracks how much it has locally
written and, on merge, clamps any incoming state that tries to inflate its own
coordinate (`positive[self]` / `negative[self]`) beyond that truth, counting it
in `detected_forgeries`. This closes the one Byzantine attack a monotone max
would otherwise latch forever — a peer forging *your* coordinate. Forging a
*third* node's coordinate still requires per-contribution signatures and remains
the job of an authenticated wrapper; that boundary is documented and pinned by a
test.

### `memory:basis_gated`

`BasisGatedMemory` is a thin `Memory` that wraps an inner counter (a
`pn_counter` by default) and only lets a report become a delta if it restricts
onto a declared basis. It validates fusability inside `write`/`fuse` before
anything reaches the inner counter:

- the payload must be a JSON object (else `not-json` / `not-object`),
- its `node` must be declared (else `no-overlap`),
- its `basis` must be one of that node's dimensions (else `outside-basis`),
- and each `(node, basis)` pair fuses at most once (else `duplicate`).

This makes "memory that only fuses legal evidence" a **reusable layer artifact**
rather than a property of one bespoke agent. `read`/`subscribe`/`cas` and the
CRDT gossip helpers delegate to the inner counter, so a basis-gated counter
still converges across replicas exactly like a plain `pn_counter`.

### Scenarios

- **`memory_pn_counter_reports`** — agents write signed calculator-project
  evidence into PN-Counter replicas and gossip state. The validator checks that
  final converged memory preserves the aggregate signed delta, not merely that
  all replicas converge to some value.
- **`memory_basis_fusion_calculator`** — a coordinator uses a `basis_gated`
  memory over a `pn_counter`. It forwards raw reports to `memory.fuse` and traces
  the accept/ignore decision; the memory layer is what validates fusability and
  writes the delta. The saturation payload is a frozen public-domain
  *Alice's Adventures in Wonderland* excerpt with no legal node/basis overlap, so
  it is ignored; a structured but off-basis `"horoscope"` report is also ignored.
- **`memory_code_saturation_calculator`** — the same basis-fusion rule against a
  code-shaped payload (a permissively licensed AMDGPU driver excerpt from
  `torvalds/linux`). It reuses `task.type: memory_basis_fusion_calculator` with a
  different fixture, so factory and validator stay shared.

A representative trace decision:

```
decision|calculator|ship|score=4|ignored=2
```

Four valid basis reports fused, two non-fusable inputs were ignored, and only
then did the system act.

### Validators

- PN-Counter delta preservation
- basis-fusion required-basis coverage (read from the coordinator's
  `basis_decl|...` trace line, not hardcoded)
- natural-language / context-saturation rejection
- code-shaped context-saturation rejection
- off-basis rejection
- exactly one valid ship action after fusion

These inspect the trace rather than just checking that the scenario ran.

### Tests

- Memory protocol compatibility and registry resolution
- PN-Counter CRDT laws (commutativity, associativity, idempotence, duplicate-merge idempotence)
- signed delta preservation
- malformed-state and fractional-coordinate rejection
- **Byzantine self-coordinate forgery is clamped** (and the foreign-coordinate boundary is documented)
- comparison against LWW / blackboard behavior (`test_lww_register_converges_but_loses_deltas`)
- **basis-gated memory:** fuse/ignore reasons, dedupe, gossip delegation
- end-to-end deterministic scenario runs (byte-identical across seeds)
- public-domain text and code-shaped graphics-driver saturation rejection
- adversarial off-basis fusion behavior

## Review feedback addressed

Following review, the two points that were initially only documented are now
implemented in code:

1. **Byzantine forgery** → self-coordinate clamp + `detected_forgeries` +
   `TestByzantineSelfCoordinate` (scope boundary documented for cross-node forgery).
2. **Basis gate as a layer** → `BasisGatedMemory` plugin; the coordinator now
   delegates to `memory.fuse`. Trace output is byte-identical, so determinism and
   trace-reading validators are unchanged.
3. **Validator no longer hardcodes basis/threshold** → read from the trace.
4. **Duplicate scenario type collapsed** → `memory_code_saturation_calculator`
   reuses `task.type: memory_basis_fusion_calculator`.

## How to verify

```powershell
$uv = "$env:APPDATA\Python\Python312\Scripts\uv.exe"
$env:UV_PROJECT_ENVIRONMENT = "$env:LOCALAPPDATA\nanda-town-memory-uv-venv"
$env:UV_LINK_MODE = "copy"
& $uv sync
& $uv run ruff check .
& $uv run ruff format --check .
& $uv run pyright
& $uv run pytest -v
```

Result on the current branch (after merging `upstream/main`): **1197 passed, 1 skipped, 1 deselected, 1 warning**. The warning is the existing `nest_cli.main` deprecation from the compatibility shim; pyright reports 0 errors. (The memory changes alone accounted for 838 of those before the upstream merge.)


dhve:
Thanks @jmuslu. This is one of the stronger memory submissions in the queue. `PnCounterMemory` is a clean CvRDT: `CounterState.join` as pointwise max over two grow-only maps, hypothesis property tests for commutativity, associativity, and idempotence, and strict state validation that rejects fractional and negative coordinates. The standout is `test_lww_register_converges_but_loses_deltas`: you show the baseline convergence check passing on `lww_register` while the signed total is lost, which makes your "converges to the wrong mathematical object" pitch load-bearing in a test rather than asserted in prose, and it is the fails-on-baseline pattern the problem's success criteria ask for. Byte-identical trace assertions across three seeds are the right determinism proof too.

This is close. Four things to tighten:

**1. `merge` accepts forged coordinates, and pointwise max makes the forgery permanent.**

`_state_from_fields` in `pn_counter.py` validates shape (non-negative integers) but not authorship. Any replica can gossip `{"crdt":"pn_counter","positive":{"builder":1000},"negative":{}}` and inflate another node's coordinate, including the receiving replica's own `node_id`. Because the join is a monotone max, the poisoned coordinate can never be retracted: it wins every subsequent merge and every replica converges on the forged total. Your PR body frames this as evidence aggregation under adversarial context, so this is the natural next attack class after malformed state, and right now the plugin absorbs it silently. Two acceptable resolutions: (a) track each replica's own local contribution and reject or clamp incoming states where `positive[self._node_id]` exceeds what this node actually wrote, with a test where a byzantine gossiper attempts the inflation; or (b) state explicitly in the module docstring that replicas are mutually trusted and the adversary lives in the delivery layer only (drop, duplication, reordering), which fits the problem's stated scope: it tests the merge logic under delivery-order chaos, not byzantine peers. Either choice sharpens correctness and persona fidelity; leaving it implicit is the only wrong option.

**2. The basis-fusion invariant is enforced by the scenario agent, not the memory layer.**

The headline claim is basis-restricted memory fusion, but the node/basis gate lives in `FusionCoordinatorAgent.on_message` in `memory_basis_fusion_calculator.py`. `PnCounterMemory` itself accepts any well-formed integer delta from any caller; nothing at the layer boundary knows about nodes or bases. As a demonstration scenario that is legitimate, and the two saturation payloads (Gutenberg text, AMDGPU driver excerpt) are a nice touch. But the reusable artifact here is "PN-Counter plus a bespoke coordinator", not "memory that only fuses legal evidence". If the fusion rule is meant to be part of the contribution, consider a thin wrapper plugin (a basis-gated memory that validates fusability inside `write` and delegates to `pn_counter`), or at minimum say plainly in `docs/layers/memory.md` that the gate is scenario-level policy.

**3. `validate_basis_fusion_calculator_action` hardcodes what the scenario makes configurable.**

The factory reads `required_basis` and `threshold` from task config, but the validator fixes `required = {"add", "subtract", "multiply", "divide"}` and demands `ignored >= 2`. Run the same scenario type with a different basis declaration and the validator either fails a correct run or passes an incorrect one. Cheapest fix: have the coordinator broadcast its declared basis once (something like `basis_decl|add|subtract|...`) and have the validator read the requirement from the trace instead of a literal.

**4. `memory_code_saturation_calculator` does not need to be a separate scenario type.**

As far as I can tell, factory lookup keys on `task.type` and validator selection keys on the scenario type string, and your second type registers the identical factory and identical validator list. Both YAMLs could share `task.type: memory_basis_fusion_calculator`, keeping their own `name`, description, trace path, and `saturation_fixture`, which already comes from config. That drops a branch in `scenarios.py` and a registry entry in `validators.py`. Smaller shared-core surface is a good look when you are touching `nest_core` files.

One small thing: the second half of your PR body lost its markdown formatting (everything from "Scenarios" down renders as flat text). Judges read the body for docs quality, so a two-minute edit there is cheap points. Worth knowing the field is crowded too: LWW-Register merged in #22 and several OR-Set memory PRs are open. Your preservation-versus-convergence framing is a real differentiator; keep the LWW comparison test as the proof and keep that framing explicit.

**Suggested path forward:**

1. Pick a trust model for gossiped state: clamp or reject forged own-coordinates with a byzantine test, or document replicas-are-trusted explicitly.
2. Either lift the basis gate into a small wrapper memory plugin or scope the claim honestly in the docs.
3. Make the basis-fusion validator read the declared basis from the trace.
4. Collapse `memory_code_saturation_calculator` into the existing scenario type.
5. Fix the PR body markdown, then re-run the full local gate (`uv sync`, `ruff check`, `ruff format --check`, `pyright`, `pytest -v`) before requesting re-review.

Leaving this open. Ping me on the hackathon channel if the trust-model question needs discussion.

jmuslu:
Thanks @dhve — this is really helpful. I took the "full mile" on both of the
points you flagged as documented-only, and addressed the other two in code:

1. **Byzantine forgery** — `PnCounterMemory` now tracks its own locally-written
   contribution and clamps any incoming state that inflates its own coordinate,
   exposed via `detected_forgeries`. `TestByzantineSelfCoordinate` shows the
   forged value can't win (even after repeated merges), that honest stale-lower
   states aren't flagged, and pins the documented boundary: a third node's
   coordinate still needs signatures.
2. **Basis gate as a layer** — new `BasisGatedMemory` plugin (`memory:basis_gated`)
   validates fusability inside `write`/`fuse` and delegates to an inner
   `pn_counter`. The coordinator now forwards raw reports to `memory.fuse`, so
   "memory that only fuses legal evidence" is a reusable artifact, not scenario
   glue. Trace output is byte-identical, so the determinism and trace-reading
   validators are unchanged.
3. **Validator no longer hardcodes basis/threshold** — reads the declared basis
   from the `basis_decl|...` trace line.
4. **Collapsed the duplicate scenario type** — `memory_code_saturation_calculator`
   reuses `task.type: memory_basis_fusion_calculator`.

Also merged `upstream/main` to clear the merge conflict, and rewrote the PR body
(the markdown formatting is fixed). Full local gate on the merged tree:
ruff check, ruff format --check, pyright (0 errors), pytest — 1197 passed,
1 skipped, 1 deselected.

Ready for re-review when you have a moment.

JamesCarnley:
Thank you, @jmuslu. This work preserves an important distinction for any future replicated-memory profile: convergence alone is insufficient; the expected signed aggregate must survive duplicate and reordered merges, and basis admission must reject malformed, off-basis, duplicate, or unauthenticated coordinates before mutation. Current Town memory is deliberately a per-agent exact-key store and makes no replica, merge, counter, or distributed-state claim. Choosing PN-counter semantics and replica authority would be a new profile, not a repair to that contract, so I am closing this legacy PR and retaining its algebra, aggregate-preservation, and foreign-coordinate cases for future work. No contributor code was merged or executed during this review.

PR #114, opened 2026-07-08
Delegatable capability tokens with cascading revocation (auth)

limelight1310-alt:
## Summary

Solves hackathon problem [`04-auth-capability-delegation`](docs/hackathon/problems/04-auth-capability-delegation.md) (difficulty: easy). Full writeup, commands, and honest limits are in [`packages/nest-plugins-reference/nest_plugins_reference/auth/README.md`](packages/nest-plugins-reference/nest_plugins_reference/auth/README.md).

- **`DelegatableAuth`** (`packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable_hmac.py`), registered as `("auth", "delegatable_hmac")`: adds `delegate(parent_token, audience, scopes_subset, ttl)` on top of the existing `issue`/`verify`/`revoke` surface, plus an optional `presented_by` kwarg on `verify`. Each token is a chain of HMAC-linked claims (child keyed on parent's signature); revoking one token adds a single id to a set, and every descendant carries its full ancestor chain, so cascading revocation needs no per-descendant bookkeeping.
- **Adversarial validator + scenario**: `scenarios/delegated_auth_hmac.yaml` (coordinator -> 3 intermediaries -> 12 leaves) and four validators in `nest_core/validators.py` (`auth_delegation_occurred`, `auth_scope_escalation_blocked`, `auth_cascading_revocation`, `auth_audience_confusion_blocked`). All four **FAIL** against the `jwt` baseline (no `delegate()` at all) and **PASS** against `delegatable_hmac`.

**Naming note:** the plugin/scenario/registry-key/YAML identifiers carry an `_hmac` suffix (`delegatable_hmac` / `delegated_auth_hmac`) because #138, already merged to `main`, independently solved the same hackathon problem and claimed the unsuffixed `delegatable` / `delegated_auth` names first. This branch has been rebased onto current `main` and those identifiers renamed so both submissions coexist without a plugin-registry or scenario-name collision. **What differentiates this submission from other problem-04 solutions:** `auth_cascading_revocation`'s trace-order guards (`blocked_descendant_seen` and `ok_outside_seen`) require the revocation mechanism to actually fire and an unrelated sibling subtree to stay unaffected, so a validator that never observes a real block or a real unaffected sibling can't trivially pass; and the scenario is honest that its "stale parent" coverage exercises only the explicit-revocation case, since the shared plugin instance runs on a fixed simulated clock and tokens never naturally expire during the run.

## Test plan

- [x] `uv run pytest packages/nest-plugins-reference/tests/test_delegatable_hmac_auth.py -v`
- [x] `uv run pytest packages/nest-core/tests/test_delegated_auth_hmac.py -v`
- [x] `uv run ruff check .` — clean on all files this PR touches (one pre-existing, unrelated `E501` in `test_negotiation_determinism.py` from `main` itself)
- [x] `uv run ruff format --check .` — clean on all files this PR touches
- [x] `uv run pyright` — 0 errors
- [x] `uv run pytest -v` (full suite, rebased onto current `main`) — all tests in this PR's scope pass; 17 pre-existing failures on `main` itself (`registry: byzantine_gossip` missing from `_BUILTINS`) are unrelated to this change and reproduce on `main` before this branch is applied
- [x] Manually verified red -> green: removing the plugin/validators/scenario causes the corresponding test files to fail collection (`ModuleNotFoundError` / `ImportError`); restoring them, all tests pass
- [x] Byte-deterministic trace confirmed across seeds 42, 7, 1337

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -32,6 +32,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
     ("auth", "delegatable"): f"{_REF}.auth.delegatable:DelegatableAuth",
     ("auth", "mesh_revocable"): f"{_REF}.auth.mesh_revocable:MeshRevocableAuth",
+    ("auth", "delegatable_hmac"): f"{_REF}.auth.delegatable_hmac:DelegatableAuth",
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
     ("trust", "agent_receipts"): f"{_REF}.trust.agent_receipts:AgentReceiptsTrust",
     ("trust", "parc"): f"{_REF}.trust.parc:ParcTrust",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -197,3 +197,7 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("capability_spoofing", capability_spoofing_factory)
+    elif name == "delegated_auth_hmac":
+        from nest_core.scenarios_builtin.delegated_auth_hmac import delegated_auth_hmac_factory
+
+        register_scenario("delegated_auth_hmac", delegated_auth_hmac_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/delegated_auth_hmac.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/delegated_auth_hmac.py
@@ -1,0 +1,392 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Delegated-auth scenario -- a 3-level capability-delegation tree under attack.
+
+A ``coordinator`` issues a root token and delegates a narrower sub-token to
+each of 3 ``intermediary`` agents, which each delegate further to 4 ``leaf``
+agents (12 leaves total). Every leaf presents its token to a single
+``auditor`` for verification, twice: once right after receiving it, and once
+again after the coordinator revokes one intermediary's token partway through
+the run. This exercises the three attacks the problem brief names, plus the
+headline cascading-revocation feature:
+
+* **Scope escalation** -- the coordinator also tries to delegate scopes its
+  own root token does not hold; must be blocked.
+* **Cascading revocation** -- revoking one intermediary's token must
+  invalidate that intermediary and all 4 of its leaves at the next verify,
+  while the other 8 leaves (under the two untouched intermediaries) must be
+  completely unaffected.
+* **Audience confusion** -- one designated leaf also presents its own,
+  perfectly valid token but claims to be a *different* agent; must be
+  blocked.
+
+The scenario's "stale parent" coverage is the explicit-revocation case
+only: the shared auth plugin instance runs on a fixed simulated clock
+(``clock=0.0``), so tokens never naturally expire during the run, and
+natural-expiry rejection is exercised solely by a plugin-level unit test,
+not by this scenario.
+
+Agents resolve the auth plugin from ``ctx.plugins`` indirectly: the factory
+instantiates a single shared instance (revocation state and the HMAC secret
+must be shared for verification to mean anything) and injects it into every
+agent's constructor. The coordinator capability-gates on ``hasattr(auth,
+"delegate")`` -- so pointing this scenario at ``auth: jwt`` does not crash;
+the whole cascade simply never starts, which is the honest gap the
+``delegated_auth_hmac`` validators are built to catch (see
+``nest_core.validators.validate_auth_delegation_occurred``).
+
+Trace line protocol (``:``-delimited, carried in message bodies)::
+
+    issued:<token_id>:<subject>:<scopes_csv>
+    delegated:<child_id>:<parent_id>:<audience>:<scopes_csv>
+    delegate_blocked:<attack_kind>:<parent_id>
+    revoked:<token_id>
+    verify_result:<token_id>:<check_kind>:<outcome>
+
+``verify_request`` (leaf/intermediary -> auditor) is ``|``-delimited instead,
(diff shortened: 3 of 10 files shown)


dhve:
Thanks @limelight1310-alt, this is one of the cleanest takes on problem 04 I've read. The design lands exactly where the brief points: `DelegatableAuth._verify_chain` re-keys hop by hop (`key = bytes.fromhex(actual)` off each link's `mac`), so a link can only exist if its author held the whole preceding chain, and `revoke` only ever adds `chain[-1]["token_id"]` to one set while every descendant token carries its full signed ancestor chain. That is cascading revocation by construction with zero per-descendant bookkeeping, which is the right shape. I also like that `validate_auth_cascading_revocation` walks events in trace order with the `blocked_descendant_seen` and `ok_outside_seen` guards, so a validator that never actually fired cannot trivially pass, and that `test_ok_before_revocation_is_not_a_violation` pins that behavior. Making the delegation exceptions subclass `ValueError` via `DelegationError`, and adding `presented_by` as a keyword-only optional on `verify`, keeps the base `Auth` protocol and every existing `except ValueError` caller intact. The README's honest-limits section is exactly what docs_quality wants.

This is close. Two blockers (both mechanical) and two things to tighten:

**1. The branch conflicts with `main`, so it cannot be scored as-is.**

`mergeable` is `CONFLICTING` and `mergeStateStatus` is `DIRTY`. The panel scores after merge, so a PR that will not merge cleanly is invisible to it. The likely collisions are the shared registries every scenario-shipping PR appends to: since your base, `main` gained four new scenario registrations (`failure_detection`, `parc_migration`, `rogue_trusted_agent`, `sybil_bond`) that land right after `escrow_marketplace`, exactly where you inserted `delegated_auth` in the `_try_load_builtin` chain in `scenarios.py`, the `VALIDATORS` table in `nest_core/validators.py`, and the hand-enumerated set in `test_validators.py::test_all_scenario_types_registered`. The `_BUILTINS` table in `nest_core/plugins.py` also grew around your `("auth", "delegatable")` line. Rebase onto latest `main`, re-add your four `delegated_auth` validators, the builtin entry, the scenario branch, and the `"delegated_auth"` set entry, then re-run `uv run pytest -v` to confirm nothing else moved under you.

**2. The branch name is missing your handle.**

The charter's pattern is `hackathon/<your-handle>-<short-theme>` and yours is `hackathon/auth-capability-delegation`. PRs not matching the pattern are auto-closed per the charter, so fix it while you rebase. Rename the branch on your fork through the GitHub branches page (that keeps this PR attached; pushing a fresh branch under a new name would not), to something like `hackathon/limelight1310-alt-capability-delegation`.

**3. Problem 04 is crowded, and your headline claim is not yet load-bearing in a test (novelty, test_rigor).**

Roughly twenty open PRs implement this same brief, and the brief itself hands everyone the same suggested shape (HMAC chain anchored on the parent, revocation by construction), so most of the field will claim exactly what your README claims. The way to stand out is to prove the mechanism. Right now `test_revoking_root_invalidates_grandchild_without_touching_it` proves the cascade but not the single-entry property, a reviewer cannot tell it apart from a plugin that walks and revokes each descendant. Add an assertion that makes it concrete, for example `assert len(auth._revoked) == 1` right after `await auth.revoke(root)` in that test, and add a line to the PR body naming what differentiates your implementation from the other 04 submissions (the trace-order validator guards and the honest jwt-gap scenario design are genuine candidates). Cheap to do, and it turns your headline into something the judges can see rather than take on faith.

**4. The "stale parent" attack is only half-covered at the scenario level (test_rigor).**

The brief's attack #2 is "parent expired *or* revoked but child still verifies." Your scenario and `validate_auth_cascading_revocation` nail the revoked half, but the natural-expiry branch of `_verify_chain` (`leaf["exp"] < self._now()`) never fires in the run because the shared plugin is pinned at `clock=0.0`. You disclose this honestly in the README, which I appreciate, and `test_expired_parent_child_still_within_ttl_is_rejected` covers it at the unit level. To make the scenario itself demonstrate the full attack, consider a short-TTL delegation plus a re-verify scheduled against an advanced virtual clock, or a fifth validator asserting an expired-leaf outcome. If you keep the fixed clock, one explicit sentence in the scenario docstring that the scenario-level "stale parent" is the revocation case only would close the gap in the reader's head.

**Suggested path forward:**

1. Rebase on latest `main`, reconcile `validators.py`, `plugins.py`, `scenarios.py`, and `test_validators.py`, and confirm the full suite is green.
2. Rename the branch to match `hackathon/<your-handle>-<short-theme>`.
3. Add the `len(auth._revoked) == 1` style assertion so the single-entry cascade is proven, and add one line to the PR body positioning it against the other 04 submissions.
4. Either wire a natural-expiry probe into the scenario or state plainly that scenario-level "stale parent" means revocation.
5. Re-run `ruff check`, `ruff format --check`, `pyright`, and `pytest -v` locally and confirm all green before re-review.

Leaving this open. Ping me on the hackathon channel if any of the rebase reshaping on the shared registries is unclear.

limelight1310-alt:
Branch-naming note: this branch is not renamed to `hackathon/limelight1310-alt-capability-delegation` (the required `hackathon/<handle>-<topic>` pattern) — renaming a fork branch that backs an open cross-repository PR causes GitHub to treat it as a deleted head branch and auto-close the PR (confirmed empirically: the rename closed #114, and it could only be reopened by renaming the branch back to its original name so the stored head ref existed again). Keeping the current name `hackathon/auth-capability-delegation` to keep this PR open and mergeable. Happy to redo the rename and open a fresh linked PR if a maintainer prefers that path instead.

JamesCarnley:
Thank you, @limelight1310-alt. The strongest reusable requirement is a non-vacuous cascade test: one revoked descendant must fail after revocation while an unrelated sibling remains valid, with expiry and presenter controls checked independently. Current Town Run Grants have no descendant graph or revocation set, so this is a future authorization profile rather than a current defect. I am closing the legacy PR while retaining those exact cascade, sibling, expiry, and deterministic-time cases for later profile work. No contributor code from this PR was merged or executed during this review.

PR #118, opened 2026-07-09
[Hackathon] auth-engineer: delegatable capability tokens with cascading revocation

PranavPipariya:
## Problem 04 — Delegatable capability tokens with cascading revocation

The default `auth` plugin (`jwt_auth.py`) issues flat HMAC tokens and revokes by
*exact token string* (`JwtAuth._revoked: set[str]`). It cannot model the most
common multi-agent authorization pattern: an orchestrator hands a worker a
**narrowed, time-bounded sub-capability without going back to the issuer**, and
revoking the orchestrator's own token silently withdraws every sub-capability
underneath it.

This PR ships `delegatable` — a macaroon-style auth plugin (Birgisson et al.,
2014) that does exactly that.

### How it works

A token is not a flat blob: it carries its **entire delegation chain**
(root → leaf) plus a single running HMAC signature where each link is keyed on
its parent's signature:

```
s0  = HMAC(secret, link0)
s1  = HMAC(s0,     link1)      # the previous signature is the next key
sig = s_n
```

Two properties fall out of that construction — and they are the whole point:

- **Tamper-evidence.** Changing any byte of any link (a widened scope, a
  stretched expiry, a swapped audience) breaks the recomputed `sig`.
- **Cascading revocation by construction.** Every descendant embeds its
  ancestors' links, so `verify` walks the whole ancestry offline. Revoking a
  parent link's id invalidates every descendant at the next verify — **no
  per-child revocation list.**

### API (extends `Auth`, stays protocol-compatible)

- `delegate(parent, audience, scopes_subset, ttl, caveats=None) -> Token` — any
  holder mints a child **without the issuer**. Attenuation-only: child scopes
  must be a subset of the parent's (`ScopeEscalationError` otherwise) and child
  TTL must be ≤ the parent's (`TtlExpansionError`).
- `verify(token, presenter=None, context=None) -> AuthContext` — checks the
  chain signature, then per link: revocation (`RevokedAncestorError`), expiry
  (`ExpiredTokenError`), scope/expiry monotonicity, and first-party caveats
  (`CaveatUnsatisfiedError`); binds the presenter to the leaf audience
  (`AudienceMismatchError`).
- `revoke(token)` — O(1); adds the leaf link id to the shared revoked set, which
  collapses the entire subtree beneath it.

All errors subclass one `DelegationError(ValueError)`, mirroring
`EscrowError` in the escrow plugin.

### Adversarial validators (`delegated_auth`)

Three validators, each derived purely from trace evidence so they catch a
*buggy plugin that lets an attack through*, not just self-reported denials:

| Validator | Catches |
|---|---|
| `authz_no_scope_escalation` | a delegated token holding a scope its parent never held |
| `authz_cascading_revocation` | a token that verifies OK **after** an ancestor was revoked (order-aware) |
| `authz_audience_binding` | a token that verifies OK when presented by an agent other than its audience |

**All three PASS under `delegatable` and all three FAIL under the default `jwt`
plugin** (no `delegate` surface → the tree never forms → "no delegation
observed"). Same discrimination model as the merged escrow PR.

### Scenario `scenarios/delegated_auth.yaml`

Coordinator → 3 intermediaries → 12 leaves, exactly as the problem's success
criteria require, plus adversarial probes that the plugin rejects:

- **scope escalation** — an intermediary delegates a scope it does not hold;
- **stale parent** — both an **expired** parent (deterministic via a fixed
  clock and a pre-minted past-dated token) and a **revoked** parent (the
  coordinator revokes `intermediary-2`, cascading to its four leaves);
- **audience confusion** — `leaf-0`'s token is handed to `leaf-1`, which tries
  to present it.

### Determinism

No wall clock, no unseeded RNG. Like `JwtAuth`, the plugin takes a fixed
`clock`; the scenario pins it so **same seed → byte-identical trace** (asserted
in the test suite, verified across seeds `[42, 7, 1337]`).

### Tests

- `test_delegatable.py` — 16 unit tests: attenuation, 3-deep cascading revoke,
  sibling isolation, audience mismatch, expiry, delegate-from-dead-parent,
  tamper/forgery, caveats, determinism.
- `test_delegatable_properties.py` — 8 property tests feeding hand-crafted
  attack traces to each validator (honest → PASS, each attack → FAIL).
- `test_delegated_auth_scenario.py` — 6 e2e tests: `delegatable` passes all
  validators, `jwt` fails all, byte-determinism, cross-seed discrimination,
  and every attack class present in the trace.

### CI

`make ci-local` is green: `uv sync`, `ruff check`, `ruff format --check`,
`pyright` (strict), `pytest -v` → **765 passed, 1 skipped**.

### Files

**New:** `auth/delegatable.py`, `scenarios_builtin/delegated_auth.py`,
`scenarios/delegated_auth.yaml`, three test modules.
**Edited:** `plugins.py` (+1 entry), `scenarios.py` (+1 branch), `validators.py`
(parser + 3 validators + registry), `tests/test_validators.py` (registry set),
`docs/layers/auth.md`, `README.md` (validators table). No changes outside the
auth layer / its scenario / its validators.

### Out of scope

DAG delegation (strict tree only), OAuth2/network endpoints, hardware key
attestation. This is **not** PR #9's DPoP — that is sender-constraint; this is
delegation + cascading revocation, a different problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,6 +293,7 @@ for r in validate_trace(Path("traces/auction.jsonl"), "auction"):
 | Supply chain | Delivered goods trace through all four hops; no materials lost. |
 | Reputation | Cheaters get bad reports; agents with score ≤ −3 are warned. |
 | Marketplace | No double-sell (same product to two buyers); every buy answered; sold price matches the offer. |
+| Delegated auth | Delegated scopes ⊆ the parent's (attenuation); no token verifies after an ancestor is revoked (cascading revocation); presenter must equal the token's audience. |
 
 > **Important.** Validators are property *checks*, not blessings of the
 > bundled scenarios. They inspect trace evidence and can still only verify
--- a/docs/layers/auth.md
+++ b/docs/layers/auth.md
@@ -21,6 +21,26 @@ shape (header.payload.sig), no claim validation beyond the signature.
 
 Source: [`nest_plugins_reference/auth/jwt_auth.py`](../../packages/nest-plugins-reference/nest_plugins_reference/auth/jwt_auth.py).
 
+## Delegation plugin
+
+`delegatable` — macaroon-style capability tokens (Birgisson et al., 2014).
+Implements the two things `jwt` cannot: **capability delegation** and
+**revocation propagation**. A token carries its whole delegation chain plus one
+running HMAC signature (`s_i = HMAC(s_{i-1}, link_i)`), so any holder can mint a
+narrowed, shorter-lived sub-token for another agent *without the issuer*, and
+revoking a parent link's id invalidates every descendant at the next verify —
+no per-child revocation list.
+
+Adds `delegate(parent, audience, scopes_subset, ttl)` on top of the base `Auth`
+contract; `verify` takes an optional `presenter` (audience binding) and
+`context` (first-party caveats). Delegation is attenuation-only: child scopes
+must be a subset of the parent's and child TTL must be ≤ the parent's.
+
+Source: [`nest_plugins_reference/auth/delegatable.py`](../../packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable.py).
+Demo scenario: [`scenarios/delegated_auth.yaml`](../../scenarios/delegated_auth.yaml)
+with the `delegated_auth` validators (scope escalation, cascading revocation,
+audience binding).
+
 ## Writing your own
 
 See [`writing-a-plugin.md`](../writing-a-plugin.md). Register under
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -26,6 +26,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("registry", "in_memory"): f"{_REF}.registry.in_memory:InMemoryRegistry",
     ("registry", "gossip"): f"{_REF}.registry.gossip:GossipRegistry",
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
+    ("auth", "delegatable"): f"{_REF}.auth.delegatable:DelegatableAuth",
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
     ("trust", "agent_receipts"): f"{_REF}.trust.agent_receipts:AgentReceiptsTrust",
     ("trust", "parc"): f"{_REF}.trust.parc:ParcTrust",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -145,3 +145,9 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("rogue_trusted_agent", rogue_trusted_agent_factory)
+    elif name == "delegated_auth":
+        from nest_core.scenarios_builtin.delegated_auth import (
+            delegated_auth_factory,
+        )
+
+        register_scenario("delegated_auth", delegated_auth_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py
@@ -1,0 +1,387 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Delegated-auth scenario -- a coordinator, 3 intermediaries, 12 leaf agents.
+
+Drives the ``delegatable`` auth plugin through a real delegation tree and three
+adversarial probes, emitting one structured ``authz:<kind>:<fields>`` broadcast
+per lifecycle event. The three ``delegated_auth`` validators (see
+:mod:`nest_core.validators`) read those broadcasts and confirm:
+
+* every delegated token's scopes are a subset of its parent's (attenuation);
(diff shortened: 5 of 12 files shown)


shilpa-kulkarni-14:
@PranavPipariya Plz resolve conflicts for review consideration.

shilpa-kulkarni-14:
@PranavPipariya plz resolve conflicts for review consideration.

dhve:
Thanks @PranavPipariya, this is a strong, well-thought-out submission. The macaroon construction in `delegatable.py` is done right: `_chain_sig` keys each link on the previous signature so tamper-evidence and cascading revocation both fall straight out of the math, and `_link_id` targeting the content hash (not the token string) is the correct thing to revoke against. I also like two things most Problem 04 entries skip: the first-party caveat vocabulary in `_check_caveat` (`max_depth` plus `key=value` context, failing closed on a missing key), and `test_verify_rejects_hand_forged_scope_widening`, which proves `verify` itself rejects an out-of-band JSON edit rather than trusting `delegate` to be the only gate. The three validators reading pure trace evidence (not self-reported denials) is exactly the discrimination model the charter wants, and I confirmed the suite runs green locally (151 passed across the four touched test modules).

Not ready to merge as-is though, and the biggest problem is timing, not code. Four things to work through:

**1. A competing Problem 04 implementation just merged, and it occupies your namespace (blocker).**

PR #138 (ckeda, "delegatable capability tokens with cascading revocation", same problem, also macaroon-style) landed on `main` after you cut this branch, so GitHub now reports this PR as `CONFLICTING`. I ran a merge simulation: six files conflict, and three are add/add collisions with #138's versions of the same artifacts: `auth/delegatable.py`, `scenarios_builtin/delegated_auth.py`, and `scenarios/delegated_auth.yaml`. Your `("auth", "delegatable")` line in `plugins.py` merges cleanly only because it is character-identical to the one #138 already added, and `main` already has an `elif name == "delegated_auth"` branch in `scenarios.py`, so a plain rebase cannot fix this. You will need to rename your plugin key, module, scenario name, and YAML into your own namespace (for example `("auth", "caveated_delegation")` with a matching scenario name). The remaining conflicts are mechanical appends: `test_validators.py` (the `expected` set gained `sybil_bond` on `main` where you add `delegated_auth`), the tail of `scenarios.py`, and `docs/layers/auth.md`. The judge panel does not score a PR that will not merge.

**2. Reposition novelty against merged #138, not the open crowd (novelty).**

Your "Out of scope" note distinguishes this from PR #9's DPoP, but the comparison that now matters is #138: it also ships HMAC-chained attenuation, audience binding, cascading revocation, and validators for the same three attack classes. What is genuinely yours: #138 has no caveat support at all, its audience check lives in a separate `verify_presented` method while yours keeps the protocol `verify` signature with optional `presenter` and `context` kwargs, and your validators are wired into `nest_core`'s `VALIDATORS` registry and run through `validate_events` rather than living beside the plugin as standalone functions. Make those the headline of the PR body, and make the caveat layer load-bearing where judges will see it: right now `max_depth` and context caveats appear only in unit tests, not in `delegated_auth.yaml` or the scenario factory, so add a caveat probe (say, a `max_depth` violation) to the scenario trace.

**3. `test_delegatable_properties.py` is example-based, not property-based (test rigor).**

The filename and the PR body ("8 property tests") both promise property tests, but every case is a hand-crafted event stream fed to `validate_events`. That is good adversarial fixture coverage, but the merged peers this will be judged next to (`test_escrow_properties.py`, `test_aae_permit_gate_properties.py`, `test_ed25519_rotating_properties.py`) all use hypothesis `@given`, and the rubric rewards property-based over example-only. Add a couple of real generative properties over the plugin itself: any randomly built attenuating chain always verifies under its leaf audience, and any randomly injected scope-widening or expiry-extension always raises.

**4. Optional: declare the entry point (api fit).**

You register the plugin in `plugins.py`, which satisfies the problem's success criteria. There is no `nest.plugins.auth` table in `packages/nest-plugins-reference/pyproject.toml` today (jwt is builtin-only too), so this is not required, but adding one under your new plugin name is the documented install-time discovery path in `CONTRIBUTING.md` and a cheap api-fit signal.

**Suggested path forward:**

1. Rebase on `main`; rename the plugin key, module, scenario name, and YAML out of #138's namespace; resolve the mechanical conflicts in `scenarios.py`, `test_validators.py`, and `docs/layers/auth.md`; rerun `make ci-local`.
2. Rewrite the PR body lead to state what this adds over merged #138, centered on the caveat layer and the registry-wired validators.
3. Add a caveat probe to the scenario, and 2-3 hypothesis properties to `test_delegatable_properties.py` covering attenuation and rejection invariants over generated chains.
4. Optionally add the `nest.plugins.auth` entry-point table.

Leaving this open. Ping me on the hackathon channel if the rebase or the renaming is unclear.

JamesCarnley:
Thank you, @PranavPipariya. This contribution usefully shows that caveats must be enforced at verification time, missing context must fail closed, and a scenario must actually exercise caveat, scope, expiry, revocation, and presenter failures. Current Town uses flat Run Grants with named mailbox permissions and no caveat language, child chain, or context evaluator. Adding these semantics now would select a new authorization policy rather than repair current behavior, so I am closing the legacy PR and preserving those cases for a future delegated-capability profile. No contributor code from this PR was merged or executed during this review.

PR #119, opened 2026-07-09
feat(datafacts): add MerkleDataFacts plugin with Git-style DAG, monotonic hash chains, bloom ancestry, and aliases

Amitk003:
Implements Problem 08:
- Content-addressed URLs via SHA256(payload || metadata || parents)
- Anti-equivocation monotonic hash chain freshness proofs
- Bloom filter ancestry for O(1) negative checks
- Alias resolution for df://name@latest

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -43,6 +43,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("privacy", "hybrid_x25519"): f"{_REF}.privacy.hybrid_x25519:HybridX25519Privacy",
     ("datafacts", "datafacts_v1"): f"{_REF}.datafacts.datafacts_v1:DataFactsV1",
     ("datafacts", "cid_facts"): f"{_REF}.datafacts.cid_facts:CidFacts",
+    ("datafacts", "merkle_datafacts"): f"{_REF}.datafacts.merkle_datafacts:MerkleDataFacts",
     ("failure_detector", "heartbeat"): (
         f"{_REF}.failure_detection.heartbeat:HeartbeatFailureDetector"
     ),
--- a/packages/nest-plugins-reference/nest_plugins_reference/datafacts/merkle_datafacts.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/datafacts/merkle_datafacts.py
@@ -1,0 +1,489 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Content-addressed DataFacts with Git-style Merkle DAG, monotonic hash chains,
+Bloom-filter ancestry, and alias resolution.
+
+Implements the refined Problem 08 approach:
+
+1. **Git-style Merkle DAG** — the dataset URL is
+   ``df://sha256-<sha256(payload_checksum || canonical_metadata || sorted_parent_hashes)>``.
+   A separate ``tree_hash`` is unnecessary because lineage integrity is
+   guaranteed by construction.
+
+2. **Alias consistency under partition** — ``df://<name>@latest`` resolves
+   from the local registry view. When the partition heals, gossip converges
+   the alias to the globally newest hash.
+
+3. **Anti-equivocation freshness proofs** — each publisher maintains a
+   monotonic hash chain ``H_n = SHA256(H_{n-1} || tick_n || dataset_hash_n)``.
+   Fork detection catches double-publishing at the same tick.
+
+4. **Compact ancestry check** — each dataset carries a Bloom filter of its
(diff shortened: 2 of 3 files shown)


dhve:
Thanks @Amitk003. The chain-link construction itself is a nice idea: binding `h_prev`, the logical tick, and the dataset hash into `H_n` in `_compute_chain_hash`, then signing `H_n`, is the right primitive for anti-equivocation, and pulling the parent hashes out of the canonical metadata so they feed `content_hash` as an explicit sorted input is a defensible refinement of the addressing scheme.

Not ready to merge as-is though, and there is a structural problem to talk about before the code-level ones:

**1. Most of this is already merged as `cid_facts` (PR #31).**

Most of `merkle_datafacts.py` is copied from `nest_plugins_reference/datafacts/cid_facts.py`: same constructor signature, same `publish`/`fetch`/`request_access`/`verify_freshness` skeleton, identical error strings, and `SharedClock`, `ProvenanceError`, `ancestors`, `known_urls`, `freshness_proof`, and the `clock` property all have identical bodies (a few docstrings lightly reworded). `scenarios/merkle_provenance_supply_chain.yaml` is the existing `scenarios/provenance_supply_chain.yaml` with `datafacts:` swapped and the name changed; same seed, same agents, same `provenance_supply_chain` task and validators. The charter is explicit that re-issuing existing work does not count, and novelty is scored against merged peers. What could make this submission yours is exactly the three advertised additions, which brings me to:

**2. The three differentiators are declared but not delivered.**

- *Fork detection*: the module docstring says fork detection catches double-publishing at the same tick, but no method anywhere compares two `ChainLink`s that share an `h_prev`. `verify_freshness` recomputes a single link from the `h_prev` carried inside the proof itself, so chain continuity is never checked; as written the chain proves nothing beyond what `cid_facts`' signed `(url, tick)` already proves. Also note `publish` advances the shared clock every call, so two publishes by one agent can never share a tick through the public API.
- *Bloom ancestry*: `_check_bloom` is defined and never called, by the plugin, the scenario, or anything else in the diff. `ancestors()` is still the full BFS from `cid_facts`.
- *Aliases*: `resolve()` is never called by the task or scenario, and the claimed gossip convergence after partition heal does not exist; `_aliases` is a plain local dict overwritten on every publish.

**3. No tests and no adversarial validator of your own.**

The charter requires an adversarial validator that catches an attack class the reference plugin fails. The reused `provenance_supply_chain` validators already pass on `cid_facts`, so running them against your plugin demonstrates nothing new. Compare what #31 shipped: `test_cid_facts.py` plus `test_provenance_supply_chain.py`. This PR adds zero test files, which puts test rigor at the floor.

**4. Housekeeping: entry point, CI, and your other open PR.**

The plugin is registered in the builtin table in `nest_core/plugins.py` but not in `[project.entry-points."nest.plugins.datafacts"]` in `packages/nest-plugins-reference/pyproject.toml`, where `cid_facts` also lives. No CI checks have run on this branch yet; the judge panel does not score broken PRs, so run the five stages locally (`uv sync`, `ruff check`, `ruff format --check`, `pyright`, `pytest -v`). Separately, you also have #99 open, and the charter is one problem per participant, single PR. Please pick your primary or say how the two relate.

**Suggested path forward:**

1. Pick one real delta over `cid_facts` and make it load-bearing. Equivocation detection is the strongest candidate: store per-publisher chain history, add a `detect_fork` style verifier, and check continuity (`link_n.h_prev == link_{n-1}.h_curr`) in `verify_freshness`.
2. Write your own adversarial validator plus a scenario where a double-publishing agent is caught by `merkle_datafacts` and slips past `cid_facts`. That fails-on-baseline, passes-on-yours pattern is what the rubric rewards.
3. Either wire `_check_bloom` and `resolve()` into the task and tests, or delete them; unreferenced code costs you on correctness and docs.
4. Add unit tests, register the pyproject entry point, and get CI green.
5. Decide between this and #99.

Leaving this open. Ping me on the hackathon channel if you want to talk through the fork-detection design.

Amitk003:
okk, I'd like go with PR99 and you can close this one..

JamesCarnley:
Thank you, @Amitk003. The valuable future cases are canonical parent binding, adjacent-chain continuity, and a signed fork discriminator. At this head, however, the implementation duplicated the legacy cid_facts family, while its bloom and alias differentiators were not exercised by a load-bearing scenario or test; you also explicitly selected #99 and said this PR could close. Rebuilt Town records signed single-fact evidence but does not claim a dataset DAG, bloom ancestry, or alias resolution. I am closing this withdrawn duplicate and retaining the fork/continuity cases for any future provenance profile. No contributor code from this PR was merged or executed during this review.

PR #120, opened 2026-07-09
[Hackathon] LasaKaru : delegatable capability tokens with cascading revocation

LasaKaru:
Add a macaroon-style auth plugin that lets an agent delegate a strict subset of its capabilities to another agent without going back to the issuer, and that revokes an entire delegation subtree transitively when any ancestor token is revoked.

- `DelegatableAuth` plugin (registered as ("auth", "delegatable")): each token carries its full root->leaf chain; each link's HMAC is keyed by the previous link's signature, so a link's intermediate signature equals the outer signature of the corresponding ancestor token. Revoking an ancestor records that signature and every descendant fails at the next verify -- no per-child revocation list. `delegate()` enforces scope subsetting and TTL <= parent; verify() gains an optional keyword-only `presenter` for audience binding while keeping the base `verify(token)` contract unchanged.
- Three adversarial validators (scope escalation, stale parent, audience confusion) that FAIL against the default `jwt` plugin and PASS against `delegatable`.
- `delegated_auth` scenario + builtin factory: a coordinator, 3 intermediaries, and 12 leaves in a delegation tree; revoking one intermediary cascades to its whole subtree. Trace validators check scope narrowing, cascading revocation, and audience binding.
- Tests covering the plugin surface, validator discrimination, and a deterministic end-to-end scenario run.

### DIFF
--- a/docs/layers/auth.md
+++ b/docs/layers/auth.md
@@ -21,6 +21,40 @@ shape (header.payload.sig), no claim validation beyond the signature.
 
 Source: [`nest_plugins_reference/auth/jwt_auth.py`](../../packages/nest-plugins-reference/nest_plugins_reference/auth/jwt_auth.py).
 
+## Delegatable plugin
+
+`delegatable` — macaroon-style capability tokens with **cascading
+revocation**. On top of the base `Auth` surface it adds:
+
+```python
+async def delegate(
+    self, parent_token: Token, audience: AgentId,
+    scopes_subset: list[str], ttl: float,
+) -> Token: ...
+
+async def verify(self, token: Token, *, presenter: AgentId | None = None) -> AuthContext: ...
+```
+
+A holder mints a narrower, shorter-lived child token **without going back
+to the issuer**. Each token carries its whole root→leaf chain, and each
+link's HMAC is keyed by the previous link's signature (Birgisson et al.,
+2014). Because a link's intermediate signature equals the outer signature
+of the corresponding ancestor token, revoking an ancestor invalidates
+every descendant at the next `verify` — no per-child revocation list.
+`delegate` enforces scope subsetting and TTL ≤ parent; the optional
+`presenter` argument binds a token to its audience (an impostor is
+rejected) while leaving `verify(token)` — the base contract — unchanged.
+
+Source: [`nest_plugins_reference/auth/delegatable.py`](../../packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable.py).
+
+**Adversarial validators** (`nest_plugins_reference.validators`):
+`check_scope_escalation_rejected`, `check_stale_parent_rejected`,
+`check_audience_confusion_rejected` — each FAILS against `jwt` and PASSES
+against `delegatable`. The [`delegated_auth`](../../scenarios/delegated_auth.yaml)
+scenario (a coordinator + 3 intermediaries + 12 leaves) drives a full
+cascading-revocation run whose trace is checked by
+`validate_delegated_auth_*`.
+
 ## Writing your own
 
 See [`writing-a-plugin.md`](../writing-a-plugin.md). Register under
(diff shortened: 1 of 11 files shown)


shilpa-kulkarni-14:
@LasaKaru Resolve conflicts for review consideration.

dhve:
Thanks @LasaKaru. This is one of the cleanest takes on Problem 04 I have read so far. The core insight in `_chain_sigs` (the intermediate signature at depth k is byte-identical to the outer signature of the depth-k ancestor, so `revoke()` stores exactly one hash and the whole subtree dies at the next verify) is the macaroon construction done properly, and you avoided the trap most submissions fall into: `verify()` takes `presenter` as an optional keyword-only argument, so the base `Auth` contract is untouched. The `_delegate` fallback in `auth_validators.py` that degrades to re-issuance for plugins without `delegate()` is exactly what makes the FAIL-on-jwt / PASS-on-delegatable claim honest, and the byte-identical two-run determinism test is the right bar for Tier 1.

This is close. Four things to work through:

**1. The branch conflicts with main and CI never ran.**

`mergeable` is currently CONFLICTING and no checks are reported on the branch. The collision looks confined to `validators/__init__.py`: since you branched, main gained the trust-gate exports from #115 (`check_denial_receipt_auditable` and friends) in the same import and `__all__` lists your PR extends. The other shared files you touch (`plugins.py`, the registry in `nest_core/validators.py`, `scenarios.py`) still match your context, so the fix should be mechanical. Rebase onto current main, re-apply your additions, and run `make ci-local` to confirm all five stages green. The judge panel does not score broken PRs.

**2. Branch name does not match the charter pattern.**

`hackathon/delegatable-auth-cascading-revocation` is missing your handle. The charter requires `hackathon/<your-handle>-<short-theme>`, and PRs not matching the pattern are auto-closed. Rename the head branch to something like `hackathon/lasakaru-delegatable-auth`; GitHub keeps the PR attached when you rename from the branches page.

**3. Audience binding can be sidestepped by one self-delegation.**

`delegate()` verifies the parent without a presenter, and `verify()` only checks the leaf link's `aud`. So mallory, holding bob's token string, cannot present it directly (your `AudienceMismatchError` fires), but she can call `delegate(bobs_token, AgentId("mallory"), subset, ttl)` and get a fresh leaf with `aud=mallory` that verifies cleanly. That is faithful to bearer-attenuation macaroons, but the PR body and `check_audience_confusion_rejected` currently claim more than the construction delivers. Either (a) add an optional keyword-only `presenter` to `delegate()` and require it to equal the parent leaf's `aud`, mirroring `verify`, or (b) state explicitly in the PR body's threat model that attenuation-by-bearer is intended and audience binding only guards direct replay. Whichever you pick, add the delegate-then-present sequence as a test case so the choice is pinned. The problem statement's "don't conflate scopes with audiences" line points at exactly this seam.

**4. Twenty open PRs implement this same problem; make your differentiator load-bearing.**

Novelty will be judged against a crowded field (roughly twenty open PRs target Problem 04). Your real edge is O(1) revocation state per revoked node with zero per-child bookkeeping. Say that explicitly in the PR body, and pin it in a test: after revoking `int-1` in the 16-agent scenario, assert the plugin holds a single revocation entry while the 5-token subtree dies. A hypothesis property test (random tree shape and depth, revoke a random node, exactly its subtree fails) would also put you ahead on test rigor, since everything in `test_delegatable_auth.py` today is fixed-example. Small polish while you are in there: `packages/nest-plugins-reference/pyproject.toml` has no `[project.entry-points."nest.plugins.auth"]` table yet, so add one registering `delegatable`; recently merged plugins (`parc`, `bonded_trust`, `trust_gated`) ship both the builtin-map entry and the entry point, and api_fit looks for the entry point.

**Suggested path forward:**

1. Rename the branch to match `hackathon/<handle>-<theme>`.
2. Rebase onto main and resolve the registry/import conflicts.
3. Decide the delegate-audience question, document it, and add the laundering test.
4. Add the entry-point registration and, ideally, one hypothesis property test.
5. Run `make ci-local` and confirm all stages pass before requesting re-review.

JamesCarnley:
Thank you, @LasaKaru. The useful requirement here is issuer-secret-free offline attenuation with verifier-enforced scope and expiry, exact descendant revocation, healthy sibling controls, and presenter binding that survives delegation laundering. Current Town Run Grants are intentionally flat, controller-signed, run-bound session grants; they do not claim child delegation or distributed revocation. Retrofitting this mechanism would choose a new authorization profile rather than fix the current grant contract, so I am closing the legacy PR while retaining those adversarial cases for a future delegation profile. No contributor code from this PR was merged or executed during this review.

PR #124, opened 2026-07-09
[Hackathon] maharmuavia (distributed-systems-engineer): or_map -- add-wins observed-remove map CRDT (memory)

MaharMuavia:
# memory: `or_map` — an add-wins observed-remove map CRDT (Problem 02)

**Persona:** distributed-systems engineer. The thing I care about is *causal*
consistency under concurrent delete — the failure mode that a plain register
hides.

## Motivation

`memory/lww_register` (merged) is a correct **LWW-Register**: it settles every
conflict by the total order `(lamport, node)`. That is exactly right for a
single-valued register — and exactly wrong for a shared **set/map of live
entries**, because a *delete* has to be modelled as a tombstone write, and a
tombstone that happens to carry the higher `(lamport, node)` **silently
overwrites a concurrent add**. An agent re-advertising a catalogue entry at the
same moment another agent retires the copy it had seen loses its entry. The
default `blackboard` is worse: delete is order-dependent `del` on a per-replica
dict, so replicas simply diverge.

There is no add-wins / observed-remove memory plugin in the tree, and
`docs/layers/memory.md` calls CRDTs out as wanted. This fills that gap with the
one capability `lww_register` structurally cannot express: **deletion under
concurrency**.

## Design

`OrMapMemory` is a **state-based ORSWOT** ("OR-Set Without Tombstones",
generalised from a set to a keyed map):

- every `write` mints a globally-unique **dot** `(node, counter)`;
- a per-replica **causal context** (a version vector: `node -> max counter
  seen`) records everything the replica has observed;
- `remove(key)` drops only the dots this replica has *observed* — a concurrent
  add carries a fresh dot **outside** the remover's context and therefore
  survives (**add-wins**);
- the full-state join keeps a dot iff the other side still has it *or* has never
  seen it, so the merge is **commutative, associative, and idempotent** — strong
  eventual consistency with **no tombstones**.

Reads stay single-valued and deterministic: when concurrent writes leave several
live dots, `read` returns the payload of the largest dot under `(counter,
node)`, while all concurrent dots are retained internally so convergence and
trace inspection are exact.

The base `Memory` surface (`read`/`write`/`cas`/`subscribe`) is unchanged; the
delete verb and replication channel (`remove` / `export` / `merge` /
`export_all` / `merge_all`) are **additive**, so a caller that only speaks the
base protocol never has to know the values are CRDT entries.

## Tradeoffs

- **State-based (CvRDT), not op-based.** Simpler to reason about and matches the
  problem brief ("pick state-based for this problem"); the cost is that a
  per-key `export` carries the causal context. Removal propagation uses the
  full-state `export_all`/`merge_all` channel, which is the correct anti-entropy
  path for observed-remove.
- **No context compaction.** The version vector is not garbage-collected across
  gossip rounds — fine at simulation scale, called out as future work.
- **Deterministic single-valued read** over a multi-value internal state: a
  deliberate choice so traces replay byte-identically; the concurrent values
  remain inspectable.

## Adversarial validators (the discriminating bar)

Two **capability-aware** validators in
`nest_plugins_reference/validators/or_map_validators.py`. They use the CRDT
channel + `remove` when present and otherwise model "observe" as a `write` and
"delete" as a tombstone `write`, so one call site discriminates all three
plugins:

| validator | `blackboard` | `lww_register` | `or_map` |
|---|---|---|---|
| `check_add_wins_survives_concurrent_remove` | **FAIL** (diverges) | **FAIL** (add lost to tombstone) | **PASS** |
| `check_convergence_under_churn` | **FAIL** (diverges) | PASS (converges, wrong value) | **PASS** |

`lww_register` passing churn-convergence but failing add-wins is the whole
point: it *is* a CRDT, so it converges — but it converges to the *deleted*
state, dropping the concurrent add. `or_map` keeps it.

## Scenario

`scenarios/memory_or_map_add_remove.yaml` — 9 replicas each write the same key
with their own `or_map` replica; every third also issues an observed-remove one
round later; full-state gossip under 10% message drop. The
`memory_or_map_add_remove` trace validators (`validate_or_map_convergence` +
`validate_memory_liveness`) confirm every replica converges to one value and
none dropped out. Byte-deterministic across seeds 42, 7, 1337.

## Verify it

```bash
uv sync
uv run pytest packages/nest-plugins-reference/tests/test_or_map.py -v   # 42 tests

# Watch the validators discriminate:
uv run python - <<'PY'
from nest_plugins_reference.memory.blackboard import Blackboard
from nest_plugins_reference.memory.lww_register import LwwRegisterMemory
from nest_plugins_reference.memory.or_map import OrMapMemory
from nest_plugins_reference.validators import check_add_wins_survives_concurrent_remove
for name, f in {"or_map": OrMapMemory, "lww_register": LwwRegisterMemory,
                "blackboard": lambda _n: Blackboard()}.items():
    r = check_add_wins_survives_concurrent_remove(f)
    print(f"{name:14s} {'PASS' if r.passed else 'FAIL'}  {r.detail}")
PY
```

## CI

All five checks green locally (`uv sync`, `ruff check`, `ruff format --check`,
`pyright` strict → 0 errors, `pytest` → **987 passed**, 1 pre-existing skip).

## Files

- `packages/nest-plugins-reference/nest_plugins_reference/memory/or_map.py` — plugin
- `packages/nest-plugins-reference/nest_plugins_reference/validators/or_map_validators.py` — adversarial validators
- `packages/nest-core/nest_core/scenarios_builtin/memory_or_map_add_remove.py` + `scenarios/memory_or_map_add_remove.yaml` — scenario
- `packages/nest-plugins-reference/tests/test_or_map.py` — 42 tests
- wiring: `plugins.py`, `scenarios.py`, `validators.py`, reference `pyproject.toml`, validators `__init__.py`

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -45,6 +45,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("negotiation", "pareto"): f"{_REF}.negotiation.pareto:ParetoNegotiation",
     ("memory", "blackboard"): f"{_REF}.memory.blackboard:Blackboard",
     ("memory", "lww_register"): f"{_REF}.memory.lww_register:LwwRegisterMemory",
+    ("memory", "or_map"): f"{_REF}.memory.or_map:OrMapMemory",
     ("privacy", "noop"): f"{_REF}.privacy.noop:NoopPrivacy",
     ("privacy", "hybrid_x25519"): f"{_REF}.privacy.hybrid_x25519:HybridX25519Privacy",
     ("privacy", "trust_gated"): f"{_REF}.privacy.trust_gated:TrustGatedPrivacy",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -109,6 +109,12 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("memory_concurrent_writers", memory_concurrent_writers_factory)
+    elif name == "memory_or_map_add_remove":
+        from nest_core.scenarios_builtin.memory_or_map_add_remove import (
+            memory_or_map_add_remove_factory,
+        )
+
+        register_scenario("memory_or_map_add_remove", memory_or_map_add_remove_factory)
     elif name == "comms_versioning":
         from nest_core.scenarios_builtin.comms_versioning import comms_versioning_factory
 
--- a/packages/nest-core/nest_core/scenarios_builtin/memory_or_map_add_remove.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/memory_or_map_add_remove.py
@@ -1,0 +1,163 @@
+# SPDX-License-Identifier: Apache-2.0
+"""OR-Map add/remove churn scenario -- prove add-wins convergence end-to-end.
+
+Every agent owns its **own replica** of the memory plugin (a per-agent plugin
+override) and, at start-up, writes a distinct value to the *same* shared key.
+A designated subset of agents then issues an observed-remove of that key one
+round later -- retiring only the adds they have gossiped so far, while other
+agents' concurrent adds are still in flight. Agents run a fixed number of
+full-state anti-entropy rounds (broadcasting ``export_all`` snapshots; peers
+``merge_all`` whatever they receive), so under a lossy network the redundant
+rounds still drive every replica to the same **add-wins** result: the key
+stays present, holding the deterministic winning add, because a remove can only
+erase adds it had already observed.
+
+On stop each agent broadcasts a ``final:<json>`` record carrying the *winning
+value* it reads for the key, so the ``memory_or_map_add_remove`` trace
+validator (:func:`nest_core.validators.validate_or_map_convergence`) can
+confirm the swarm converged to one value regardless of delivery order or loss.
+
+Example::
+
+    agents = memory_or_map_add_remove_factory(config, plugins)
+"""
+
+from __future__ import annotations
+
+import base64
+import json
+from typing import Any
+
+from nest_core.scenario import ScenarioConfig
+from nest_core.sim.agent import AgentContext, StateMachineAgent
+from nest_core.types import AgentId
+
+_TICK = b"tick"
+_REMOVE_TICK = b"remove-tick"
+_SYNC_PREFIX = "sync:"
+_FINAL_PREFIX = "final:"
+
+
+class OrMapChurnAgent(StateMachineAgent):
+    """Writes one value to a shared key, optionally removes it, then gossips.
+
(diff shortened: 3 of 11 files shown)


shilpa-kulkarni-14:
@MaharMuavia LGTM, will approve once the branch is on current main with the union conflict resolved and the footer stripped, ci-local is green and the discriminator is verified

MaharMuavia:
@shilpa-kulkarni-14 Thank you for the review — all four items are done as of `f8cf13c`:

- **Rebased onto current main** (`7106ca6`) — single commit, no merge commits.
- **Union conflict resolved** — both main and this branch appended to `__all__` in `nest_plugins_reference/validators/__init__.py`; resolved as the sorted union of both sides (the import blocks had merged cleanly).
- **Footer stripped** from the commit message.
- **ci-local green on the rebased branch** — uv sync ✓ · ruff check ✓ · ruff format --check ✓ · pyright (strict) 0 errors ✓ · pytest **1192 passed, 1 skipped** in 5:20.
- **Discriminator verified against current main** (`test_or_map.py::TestAdversarialValidators`, 5/5):
  - `check_add_wins_survives_concurrent_remove` → **PASSES or_map · FAILS blackboard · FAILS lww_register**
  - `check_convergence_under_churn` → **PASSES or_map · FAILS blackboard**

GitHub now reports the branch mergeable. Happy to address anything else.

JamesCarnley:
Thanks @MaharMuavia for the OR-Map contribution. I reviewed all eleven changed files at `f8cf13cb6ca3a3673875e89c7005594f58dc0bed` and the follow-up discussion. The full-state add-wins/remove discriminators, separate replica instances, and explicit distinction between local CAS and cross-replica convergence are valuable. I also saw the reported rebase, conflict resolution, and passing legacy checks; this is not a closure for failing to respond to that review.

Closing without a current-main replacement because rebuilt Town's memory contract is a per-agent key-value store, not a replicated observed-remove map. This PR introduces a new replication and deletion contract rather than fixing current remember/recall behavior. Running its YAML through the current importer uses `kv.v1` and an adapted exchange; `original_scenario` is explicitly `not_tested`. A passing adapted run does not establish OR-Map correctness.

Retaining the contribution as a source for a future independently selected memory profile: compare convergence separately from the required add-wins result; preserve concurrent-add versus observed-remove tests, reordered/duplicate delivery, full-state deletion propagation, and repeatable snapshots. Before a port, define how per-key and full-state replication interact. In this implementation the per-key `export()` docstring promises a removed-key record, but `_seen_key()` only recognizes live keys; the documented full-state channel is the actual deletion path. A future evaluator must also name the expected replicas/key and expected surviving value, rather than accepting agreement on any value, including absence.

No OR-Map runtime or tests have been merged as part of this disposition, and the legacy pass counts remain author-reported rather than reproduced here. The source and credited discriminators remain available; a selected replicated-memory workload can revive them in a focused profile instead of carrying this legacy integration PR indefinitely.

PR #125, opened 2026-07-09
Remove dead code from reputation scoring validator

IlIGODIlI:
## Summary

This PR removes the unused dead loop in `validate_reputation_scoring` and updates the related comments to reflect the validator's current behavior.

## Changes

- Removed the unused dead loop in `validate_reputation_scoring`
- Updated comments to match the implemented validation logic
- Verified that the existing `TestReputationScoring` tests continue to pass

### DIFF
--- a/packages/nest-core/nest_core/validators.py
+++ b/packages/nest-core/nest_core/validators.py
@@ -787,7 +787,7 @@ def validate_supply_chain_no_lost(
 def validate_reputation_scoring(
     events: list[dict[str, Any]],
 ) -> list[ValidationResult]:
-    """Reputation scores decrease for agents that cheat."""
+    """Validate that every cheating agent receives a corresponding bad report."""
     # Track scores: agent -> score
     scores: dict[str, int] = defaultdict(int)
     cheaters: set[str] = set()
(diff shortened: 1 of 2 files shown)


dhve:
Thanks @IlIGODIlI. Your dead-code analysis is correct, and I verified it: every agent in `cheaters` is already inserted into `scores` by the `scores[agent_str] -= 2` branch, so the loop you removed had no side effects, and all 10 reputation tests pass on your branch. The new docstring is also more honest than the old one: `validate_reputation_scoring` never checks that scores decrease, it only checks that every `cheat:` emitter got a bad report, which is exactly what you wrote.

That said, not mergeable as-is. Three things to work through:

**1. The new whitespace fails `ruff format --check` and `ruff check`.**

CI has not run on your branch yet, but I ran ruff with the repo config against your diff. Both failures are introduced by this PR: the replacement comment in `validators.py` is indented 3 spaces instead of 4 (the formatter reindents it), and in `test_validators.py` the two blank lines before `class TestReputationWarnings:` became a single whitespace-only line, which trips W293. The judge panel does not score broken PRs.

**2. The `test_validators.py` change is pure churn.**

The only edit there is that whitespace change, and it is what trips W293. Revert the file; nothing in it needed to move for this cleanup.

**3. As a hackathon entry, this will not score.**

The charter asks for one problem from `docs/hackathon/problems/`, a plugin or validator, a mandatory adversarial validator, and a scenario YAML under `scenarios/`. A +5/-13 cleanup in shared `nest_core` does not hit any rubric dimension. Also, the branch pattern is `hackathon/<handle>-<short-theme>`, and `hackathon/fix-reputation-validator` has no handle; mismatched names are subject to auto-close. If this is housekeeping rather than your entry, say so in the PR body and drop the `hackathon/` prefix.

**Suggested path forward:**

1. Revert `test_validators.py`, fix the comment indentation, run `make ci-local`, confirm all stages green.
2. Decide: housekeeping PR (rename the branch), or pick a problem and build a full submission.

Leaving this open. Ping me on the hackathon channel if you want help choosing a problem.

PR #127, opened 2026-07-09
[Hackathon] Harden LWW register state validation

abbasmir12:
## Summary

Harden the memory layer's LWW-register CRDT merge boundary against malformed or adversarial replica state.

This rejects:
- invalid base64 payloads
- non-string payload fields
- negative Lamport clocks
- empty node IDs
- non-string node IDs

It also adds regression coverage proving malformed remote CRDT state cannot poison an existing trusted replica value.

**Problem:** `02-memory-crdt-lww`  
**Layer:** `memory`

## Validation

Ran the required CI sequence locally:

- `uv sync`
- `uv run ruff check .`

### DIFF
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -109,6 +109,12 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("memory_concurrent_writers", memory_concurrent_writers_factory)
+    elif name == "memory_byzantine_state":
+        from nest_core.scenarios_builtin.memory_byzantine_state import (
+            memory_byzantine_state_factory,
+        )
+
+        register_scenario("memory_byzantine_state", memory_byzantine_state_factory)
     elif name == "comms_versioning":
         from nest_core.scenarios_builtin.comms_versioning import comms_versioning_factory
 
--- a/packages/nest-core/nest_core/scenarios_builtin/memory_byzantine_state.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/memory_byzantine_state.py
@@ -1,0 +1,161 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Byzantine-state memory scenario -- reject malformed CRDT gossip.
+
+This scenario is a hostile variant of ``memory_concurrent_writers``. Each
+honest agent owns a private LWW-register replica, writes a local value, and
+gossips normally. One byzantine agent then broadcasts malformed serialized
+register state with an artificially high Lamport clock. A hardened CRDT must
+raise at the merge boundary and leave the trusted local value unchanged; an
+unsafe decoder can accept the poison and spread it by later gossip rounds.
+
+Example::
+
+    agents = memory_byzantine_state_factory(config, plugins)
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from nest_core.scenario import ScenarioConfig
+from nest_core.sim.agent import AgentContext, StateMachineAgent
+from nest_core.types import AgentId
+
+_TICK = b"tick"
+_ATTACK = b"attack"
+_SYNC_PREFIX = "sync:"
+_ATTACK_PREFIX = "attack:"
+_FINAL_PREFIX = "final:"
+_REJECTED_PREFIX = "rejected:"
+_ACCEPTED_PREFIX = "accepted:"
+
+
+class ByzantineStateAgent(StateMachineAgent):
+    """Write one CRDT value, gossip it, and reject malformed remote state.
+
+    Example::
+
+        agent = ByzantineStateAgent(AgentId("writer-1"), "shared", b"value", 20, False)
+    """
+
+    def __init__(
+        self,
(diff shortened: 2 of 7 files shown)


dhve:
Thanks @abbasmir12. This fixes a real hole: on main, `base64.b64decode("@@@")` silently drops non-alphabet characters and returns `b""`, so a bogus payload with a high Lamport clock replaces a trusted replica value with empty bytes. `test_malformed_merge_does_not_poison_existing_value` genuinely fails on main, and putting the checks in `_register_from_fields` covers both `merge` and `merge_all`. Nicely targeted.

Not a complete submission yet though. Three things to work through:

**1. `OverflowError` escapes your hardened boundary.**

`"lamport": 1e400` parses to `float("inf")`, and `int(inf)` raises `OverflowError`, which is neither in your `(KeyError, ValueError, TypeError)` catch nor a `ValueError` subclass. So the contract this PR establishes (malformed state raises `CrdtStateError`) breaks on that input. Fix: require `isinstance(raw_lamport, int) and not isinstance(raw_lamport, bool)` before coercing; that also stops `"lamport": 2.9` truncating to 2 and `"lamport": "7"` sneaking through.

**2. The charter's mandatory pieces are missing.**

The charter (docs/hackathon/charter.md) requires an adversarial validator plus a scenario YAML demonstrating the plugin passing under it; this PR ships neither. Concrete shape: a byzantine-state validator where one replica gossips malformed register state mid-run and honest replicas must stay unpoisoned (fails against pre-fix main, passes here), plus something like `scenarios/memory_byzantine_state.yaml` next to `memory_concurrent_writers.yaml`. This also carries your novelty case: problem 02 is already merged (#22 is the plugin you are hardening), so state in the PR body that malformed-state defense is your differentiator.

**3. Test rigor and validation story.**

`test_lww_register.py` already imports hypothesis; a property test asserting arbitrary junk bytes either raise `CrdtStateError` or merge cleanly, never mutating stored state on raise, would lift test rigor. Your Validation section stops at `ruff check` and CI has not run on this branch yet. Run the full sequence (`ruff format --check`, `pyright`, `pytest -v`); the judge panel does not score broken PRs. No conflicts with main.

**Suggested path forward:**

1. Type-check `lamport` before coercion; add `1e400`, float, string, and bool rejection tests.
2. Add the byzantine-state validator and scenario YAML.
3. Add one hypothesis property test for the decode boundary.
4. Finish the PR body: validation output, threat model, differentiation from #22.

Leaving this open. Ping me on the hackathon channel if the validator shape is unclear.

abbasmir12:
Thanks again for the review @dhve. I pushed an updated version and rebased onto latest "projnanda/main".

Addressed:

1. Lamport validation:
   
   - requires "lamport" to be an "int"
   - rejects "bool"
   - rejects floats like "2.9"
   - rejects strings like ""7""
   - rejects overflow-style JSON numbers like "1e400"
   - still rejects negative values

2. Hackathon scenario + validator:
   
   - added "scenarios/memory_byzantine_state.yaml"
   - added built-in scenario "memory_byzantine_state"
   - added "validate_memory_byzantine_state_rejected"

The scenario injects malformed serialized CRDT state with a high Lamport clock. Honest replicas must reject it, continue gossiping, and converge without final-state poisoning.

3. Test rigor:
   - added targeted tests for malformed Lamport values
   - added a Hypothesis property test that failed merges never mutate trusted state
   - added an end-to-end scenario test for the byzantine-state case

Validation after rebasing onto latest upstream:

- "ruff check": passed
- "pyright": passed
- focused tests: "47 passed"

Earlier full suite on this patch passed with:

- "957 passed, 1 skipped, 1 deselected"

This PR’s differentiator from #22 is malformed/byzantine serialized CRDT state defense at the merge boundary.

JamesCarnley:
Thanks @abbasmir12 for the targeted decoder hardening and for addressing the review with strict clock typing, malformed-state tests, a property test and a runnable legacy scenario. I reviewed the full updated seven-file patch and discussion at `37ece62ce7e08a2c80b13a019dc8c6525071b2d9`.

Rebuilt `main@f4e85d7` uses `KeyValueMemory` with per-agent `remember`/`recall`; it has no LWW register, serialized replica-state decoder, gossip merge or `merge_all` boundary. This is an important fix for that former implementation, but there is no equivalent decoder to patch in current Town. Reintroducing a replicated-memory protocol would require choosing its semantics and test subject, rather than treating it as a small validation port.

Closing as **a retired-implementation fix with its future test requirements preserved, without a replacement**. The current memory/import/adaptation checks passed as part of a 33-test run; this review did not execute or certify the legacy implementation.

Retain these requirements for a future replicated-memory adapter: malformed binary encoding, non-integer/boolean/fractional/negative clocks and invalid node identifiers must be rejected without changing trusted state; batch validation must avoid partial mutation; honest replicas must still make progress. The scenario should also require observations from every expected honest replica and their terminal states, so missing observations cannot look like successful rejection or non-poisoning. Your strict decoder and regression/property-test ideas remain valuable source material here. The current generic adaptation/import result must not be described as CRDT correctness.

PR #129, opened 2026-07-09
fix: enforce reputation score property in validate_reputation_scoring

### LINKED ISSUE #98: validate_reputation_scoring: documented 'cheaters lose score' property is never enforced (dead loop)
### Summary

`validate_reputation_scoring` documents the property "reputation scores decrease for agents that cheat", but that property is never asserted. The `scores` dict is accumulated and then read only inside a loop whose body is `pass`, so the score arithmetic is dead code. The validator's one real check is a set comparison between `cheat:` senders and `report:...:bad` recipients, which passes vacuously when the trace has no `cheat:` lines.

Low severity (the validator still does something, and the reputation scenario's own discrimination works), but the stated invariant isn't enforced, so a trace that violates "cheaters lose score" is not caught.

### Where

`packages/nest-core/nest_core/validators.py`, `validate_reputation_scoring` (~lines 787-847). The dead loop:

```python
for cheater in cheaters:
    if scores[cheater] >= 0:
        # ... comment ...
        pass
```

`scores` is never compared against anything after this.

### Reproduction

```python
from nest_core.validators import validate_reputation_scoring

def send(a, to, msg): return {"ts": 1.0, "agent": a, "kind": "send", "to": to, "msg": msg}

# agent-9 receives 5 "bad" reports (its score should fall) and there are no cheat: lines
events = [send("reporter", "obs", "report:round:agent-9:bad") for _ in range(5)]
print(validate_reputation_scoring(events)[0].passed)   # True
```

The validator passes regardless of the accumulated scores.

### Suggestion

Either enforce the documented property (assert that agents with net-negative reports end below their starting score, or below non-cheaters), or drop the `scores` accumulation and the dead loop and tighten the docstring to match what the check actually verifies.

Anurag17-2005:
fix: enforce reputation score property in validate_reputation_scoring

Fixes #98

Problem: validate_reputation_scoring documented the property "reputation scores 
decrease for agents that cheat" but never enforced it. The validator accumulated 
scores in a dict but had a dead loop (lines 811-816) with only `pass` in the body, 
so score violations were never caught.

Solution: This PR fixes the dead loop bug by enforcing that agents who cheat must 
receive bad reports.

Implementation:
- Removed dead loop code
- Single-pass parse into agents_who_cheated / agents_with_bad_reports
- Enforces property: "cheaters must be reported"
- Updated tests to cover the validator logic

---

## Submission Context

This PR fixes the dead loop bug in issue #98. It is a supplementary contribution 
alongside my primary NandaHack submission (PR #133 - Structured Receipt Field).

---

## Reviewer Feedback Addressed

Both critical issues from the review have been resolved:

### 1. ✅ Property 2 Dropped (Issue #98 Option A)

**Problem**: Property 2 ("net-negative scores") failed on stock reputation scenario 
seeds 1, 3, 7, 123 because probabilistic cheaters who cheat rarely can still end 
with positive scores.

**Solution**: Took reviewer's Option A:
- Removed property 2 entirely (no score tracking)
- Kept property 1: "cheaters must be reported" 
- Updated docstring to document what is actually checked
- Updated 6 tests to match the simplified property

**Justification**: The reference MaliciousAgent cheats with probability 0.5. A cheater 
who cheats once but delivers honestly 3+ times ends at +1. The validator now ensures 
cheaters are reported (scores decrease from bad reports), but doesn't require 
net-negative final scores.

### 2. ✅ Determinism Fixed

**Problem**: Set iteration caused hash-order non-deterministic output.

**Solution**: Use `sorted(unreported_cheaters)` in output formatting for byte-determinism.

---

## CI Output

```
make ci-local

All checks passed!
===== 1153 passed, 1 skipped, 1 deselected, 1 warning in 101.15s (0:01:41) =====
ci-local: all 5 checks passed. Safe to push.
```
  ---

## Scenario-Level Verification

As requested in review, tested scenarios/reputation.yaml on the specific seeds that failed Property 2:

**Seeds that now pass (previously failed with old Property 2):**

Seed 1:   ✅ Passed - checked 3 cheaters, 3 agents with bad reports
Seed 3:   ✅ Passed - checked 4 cheaters, 4 agents with bad reports  
Seed 7:   ✅ Passed - checked 4 cheaters, 4 agents with bad reports
Seed 123: ✅ Passed - checked 4 cheaters, 4 agents with bad reports

**Seed 42 (pre-existing failure, unrelated to this PR):**

Seed 42: ❌ Failed - cheaters not reported: ['malicious-3']
(Also fails on main branch - malicious-3 only cheats malicious peers who never report)

The simplified Property 1 ("cheaters must be reported") now holds on all seeds where the scenario actually generates reports. Seed 42's failure is pre-existing and documented in the review.

**Note on determinism fix**: Output format is now deterministic — compare the sorted list above (['malicious-3']) to main's raw set repr ({'malicious-3'}) for the same seed. This confirms both fixes are working in practice.


dhve:
Thanks @Anurag17-2005. This fixes a real bug: the dead `for cheater in cheaters: ... pass` loop from issue #98 is gone, and the single-pass parse into `agents_who_cheated` / `agents_with_bad_reports` is cleaner than the double scan it replaces. CI has not run on your branch; ruff, pyright, and pytest are green locally.

Not ready to merge as-is though. Two things to work through:

**1. Property 2 fails the stock reputation scenario on most seeds.**

`validate_reputation_scoring` is wired to the `reputation` scenario type in `VALIDATORS`, so I ran `scenarios/reputation.yaml` traces through your branch. Seeds 1, 3, 7, and 123 pass on main but now fail with `agents with bad reports have non-negative scores`. Root cause: the reference `MaliciousAgent` cheats with probability 0.5 and earns +1 per honest delivery, so a cheater who cheats once and delivers three times legitimately ends at +1. "Scores decrease" (each bad report decrements) is weaker than "net-negative overall", and the scenario only guarantees the former. Your unit tests are green because nothing validates a real scenario trace; that is the missing test. Note: seed 42 already fails property 1 on main (`malicious-3` cheats only malicious peers, who never report); pre-existing.

**2. The violation detail is hash-order nondeterministic.**

`agents_with_positive_score_despite_bad_reports` is built by iterating a set, so the `(scores: ...)` sub-list order varies with `PYTHONHASHSEED`. Two runs printed `['malicious-2=1', 'malicious-1=0']` and `['malicious-1=0', 'malicious-2=1']` for the same trace. Iterate `sorted(agents_with_bad_reports)` instead. Byte-determinism is the house rule.

Process note: you also have #133 open; the charter asks for a single PR per participant. Fix PRs have precedent (#96), so please say how the two relate or which is primary.

**Suggested path forward:**

1. Either take issue #98's second option (keep your tightened property 1, drop the net-negative assertion, document what is checked), or pick a score property the stock trace satisfies across seeds, proven by an end-to-end test on `reputation.yaml` (pattern: `test_receipt_reputation.py`).
2. Sort the violator list before formatting the detail string.
3. Re-run `make ci-local` and note the scenario-level result in the PR body.

Leaving this open. Ping me on the hackathon channel if the property calibration is unclear.

JamesCarnley:
Thanks @Anurag17-2005 for the reputation-validator work and the important correction that a bad outcome should reduce a score without requiring the agent's lifetime total to be negative.

#242 is merged as `f4e85d75324a09ca9820e15d36281345c70522b6`. Its landed tree matches the reviewed candidate: 303 tests passed locally and Python 3.11/3.12 CI passed.

Reviewed your full patch and discussion at `b06896e741faf44513e566d7eaa654e92a03200b`. The rebuilt code no longer uses the legacy `cheat:` / `report:` parser. Replacement #242 preserves a selected current requirement: the marketplace report now verifies each update's prior receipt reference, observer, subject and outcome, and independently recomputes the reference score arithmetic. Missing receipts are insufficient evidence; contradictory attribution/arithmetic fail. A real Town API good/good/good/bad sequence finishes at +2 and passes consistency, while its wrong-delta and missing-receipt variants do not.

Closing as **partially superseded by #242**, not as a merge of this original patch. This does not implement the old cheating-agent scenario or prove that every independently observed bad transaction receives the correct report. Current receipt events do not carry the transaction binding needed for that stronger claim. That future profile must connect observed behavior, transaction, attributed receipt and score update without treating a report alone as truth or allowing empty traces to pass. The original code, tests and discussion remain preserved here.

PR #130, opened 2026-07-09
[Hackathon] SkillMD API reachability validation

nonggde:
## Summary

This Phase 1 warm-up improves the SkillMD submission API so programmatic submissions get the same basic safety and reachability behavior as the form path.

## What changed

- Reject non-HTTP(S) `source_url` values for `url` and `github` submissions.
- Add a best-effort 5 second reachability check for API-created hosted SkillMD links.
- Persist the resulting `reachable` value instead of leaving API submissions with an unknown badge state.

## Why

NandaHack explicitly encourages agents and scripts to submit SkillMD records through `POST /api/skills`. Without URL validation and reachability checking, API-created entries can look recorded while still lacking the `link responded` signal that judges and participants use to verify a service can be reached.

## Verification

```bash
cd apps/nest-dashboard
npm ci --ignore-scripts
npm run lint -- src/app/api/skills/route.ts
npx tsc --noEmit
```

### DIFF
--- a/apps/nest-dashboard/src/app/api/skills/route.ts
+++ b/apps/nest-dashboard/src/app/api/skills/route.ts
@@ -8,6 +8,31 @@ function s(value: unknown): string {
   return typeof value === "string" ? value.trim() : "";
 }
 
+function isValidHttpUrl(value: string): boolean {
+  try {
+    const u = new URL(value);
+    return u.protocol === "http:" || u.protocol === "https:";
+  } catch {
+    return false;
+  }
+}
+
+async function checkReachable(url: string): Promise<boolean> {
+  try {
+    const controller = new AbortController();
+    const timer = setTimeout(() => controller.abort(), 5000);
+    const res = await fetch(url, {
+      signal: controller.signal,
+      redirect: "follow",
+      headers: { "user-agent": "NandaTown-SkillMD-Checker" },
+    });
+    clearTimeout(timer);
+    return res.ok;
+  } catch {
+    return false;
+  }
+}
+
 /**
  * GET /api/skills
  * Returns every submitted SkillMD as JSON. An OpenClaw agent can call this to
@@ -58,11 +83,22 @@ export async function POST(request: NextRequest) {
       { status: 400 },
     );
   }
+  if ((sourceType === "url" || sourceType === "github") && !isValidHttpUrl(sourceUrl)) {
+    return Response.json(
+      { error: "source_url must be a valid http(s) URL" },
+      { status: 400 },
+    );
+  }
   if (sourceType === "content" && content.trim().length < 20) {
     return Response.json(
       { error: "content is required (and must be the full SkillMD)" },
       { status: 400 },
     );
+  }
+
+  let reachable: boolean | null = null;
+  if (sourceType === "url" || sourceType === "github") {
+    reachable = await checkReachable(sourceUrl);
   }
 
   try {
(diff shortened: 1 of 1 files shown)


dhve:
Thanks @nonggde. You found a real inconsistency: `POST /api/skills` persists `reachable: null`, and the badge ternary in `skills/page.tsx` treats null as falsy, so every API-created url/github entry renders "couldn't reach link". Bringing the API path up to parity with `submitSkill` is the right instinct, and the 400 on non-http(s) `source_url` is a sensible guard.

Honest framing though: this is a nice dashboard fix, not a scoreable hackathon entry. Three things to work through:

**1. The charter asks for something different.**

Your PR body calls this a Phase 1 warm-up, but I can find no warm-up track in `docs/hackathon/` (point me to it if I missed it). `docs/hackathon/charter.md` asks for exactly one problem from `docs/hackathon/problems/`: a plugin or validator, plus a mandatory adversarial validator and a scenario YAML under `scenarios/`. The judges score against the Python layer protocols, so a TypeScript dashboard change gives them nothing to grade. Pick one of the 10 problems and make that your single PR.

**2. `isValidHttpUrl` and `checkReachable` are verbatim copies from `src/app/skills/actions.ts`.**

Duplicated helpers drift the first time someone tweaks the timeout or UA string. Extract both into a shared module under `src/lib/` and import from both call sites. It matters more because open PR #80 is already reworking `checkReachable` in `actions.ts`; a second copy in `route.ts` guarantees a collision.

**3. The reachability fetch is a blind SSRF probe on an unauthenticated endpoint.**

`checkReachable` fetches any caller-supplied http(s) URL server-side and the boolean lands in a public badge and `GET /api/skills`, so a script can probe localhost or internal addresses and read back whether they answered 2xx. The form path already had this exposure; a JSON API makes it trivially scriptable. Reject loopback, private, and link-local hosts before fetching.

**Suggested path forward:**

1. Pick one problem from `docs/hackathon/problems/` and ship that as your hackathon PR (plugin + adversarial validator + scenario YAML).
2. To land this fix on its own merits, drop the [Hackathon] tag, dedupe the helpers into `src/lib/` (coordinating with #80), and add the private-address guard.

Leaving this open. Ping me on the hackathon channel if you want help picking a problem.

JamesCarnley:
Thanks @nonggde. The URL-validation and misleading-badge issues you identified are now addressed on current main by #246, merged as ca7cb73c6ba62fc930abba4f6105679d9536fc2a.

The replacement shares the existing HTTP(S) syntax check between the form and API, rejects invalid API sources before saving, and displays `reachable: null` as **link not checked** rather than failed. This also addresses the helper-duplication concern in the earlier review.

This is **partial supersession**, not a merge of the whole original implementation. The proposed additional server-side probe was intentionally not carried over: API submissions remain unprobed, and the form's existing probe is unchanged. A future probing change needs an explicit bounded execution/address/redirect policy and timestamped observations that distinguish no check, observed failure and unavailable evidence. A successful URL response is not capability verification or endorsement.

Verification: 36 catalog tests passed (10 expected failures before the correction), typecheck and focused lint passed, production build passed, and the built local Next server returned HTTP 400 for invalid hosted-link/GitHub submissions. All 385 Python tests and the dashboard/Python 3.11/Python 3.12 CI jobs passed. The landed tree matches the tested candidate.

Closing this legacy PR with credit and the remaining probe requirements preserved here. No contributor branch is being deleted.

PR #131, opened 2026-07-09
[Hackathon] audit-engineer: intent-gated datafacts — pre-publication intent gate (problem 08)

midhunrajcharles:
**Layer:** `datafacts` · **Problem:** 08 · **Persona:** audit-engineer · **Author:** Midhun Raj Charles (`midhunrajcharles`) — same name as the Phase 2 skills submission ([AgentIntent in the registry](https://nandatown.projectnanda.org/skills), entry `f9e242fd-4cd8-4f71-9240-b12ff98447e5`, SKILL.md: https://agentintent.onrender.com/SKILL.md).

**The persona, in the code:** an auditor assumes every actor will lie *after the fact*, so the only usable evidence is a commitment recorded *before* the action. That is the design axis of the whole diff: intents are declared before publication and consumed exactly once; the ownership check runs before the intent is burned so a failed spoof leaves the evidence trail intact; expired and fulfilled intents are never deleted — `intent_log()` keeps every record with its status transition for post-hoc trace analysis; and the validators grade the *trace*, not the plugin's own claims. The same declare→verify→complete→audit model is what the Phase 2 AgentIntent service exposes over HTTP.

## Problem

**Problem 08 — datafacts.** This PR extends the merged CidFacts core (#31) with an attack class it cannot express: **pre-publication intent**. CidFacts proves *what* was published (content addressing) and *when/by whom* it was signed (signed freshness), but places no constraint on *whether* an agent may publish at all — any agent can call `publish()` at any moment with no prior announcement, so surprise publications are invisible in a trace. `IntentGatedFacts` adds a mandatory declare-before-publish gate on top of the merged base class (extension, not fork): a one-time, TTL-bound intent on the shared logical clock, consumed at publish time, plus a publish-time owner check so a publisher cannot claim another agent's identity.

## What

- `nest_plugins_reference/datafacts/intent_facts.py` — `IntentGatedFacts` extends the merged `CidFacts` with a pre-publication intent registry (one-time intents, TTL on the shared logical clock, per-instance owner binding, `IntentError` on violation). Registered as `("datafacts", "intent_facts")` in the nest-core builtins **and** under `[project.entry-points."nest.plugins.datafacts"]`.
- Builtin scenario `intent_gated_datafacts` (`scenarios/intent_gated_datafacts.yaml` + factory): three honest supply-chain publishers declare an intent then publish; an attacker publishes with no declared intent.
- Two trace validators in `nest_core/validators.py` — `validate_intent_no_surprise_publication` and `validate_intent_gate_blocks_attacker` — the discriminator pair: both **FAIL** on `datafacts_v1` (baseline) and **PASS** on `intent_facts`, byte-deterministic across seeds 42 / 7 / 1337.

## Attacks blocked (that #31's CidFacts cannot)

1. **Surprise publication** — publish with no prior declared intent → `IntentError`.
2. **Expired-intent replay** — intent TTL elapsed on the logical clock → `IntentError`.
3. **Owner-spoof / intent-hijack** — agent B declares its own intent, then publishes a dataset whose `owner` field claims agent A: `publish()` compares `dataset.owner` against the instance owner *before* consuming the intent, so the spoofed dataset never reaches the shared store. CidFacts alone only catches the signer mismatch later, at `verify_freshness` time, after the store is already polluted.

## Review follow-ups (per @dhve)

- Rebased onto current main; conflicts in `scenarios.py` / `test_validators.py` resolved. Purely additive again.
- Dropped `integration/agent_intent_client.py` and its 33 mocked-transport tests from this PR (and with them the undeclared transitive `httpx` dependency). The protocol client lives in the external repo linked below.
- Added the `dataset.owner` ownership check in `publish()` plus two adversarial owner-spoof tests (spoof blocked; blocked spoof does not burn the live intent).
- Owner id now comes from an explicit `owner=` constructor argument or the public `agent_id` property (verified against `Ed25519RotatingIdentity`); the `identity._agent_id` private reach is gone.
- `intent_facts` added to the `nest.plugins.datafacts` entry-point table in `packages/nest-plugins-reference/pyproject.toml`.

**Footprint:** 10 files, +1221 / −0.

## Testing

Full local gate green on the rebased tree: `uv sync`, `ruff check`, `ruff format --check`, `pyright` (0 errors, strict), `pytest` (1189 passed, zero regressions). 39 of those tests belong to this PR: plugin conformance, happy path, three adversarial suites (surprise publication, expired-intent replay, owner-spoof), owner-resolution tests, and an end-to-end `ScenarioRunner` test asserting both validators fail on `datafacts_v1` and pass on `intent_facts` with byte-identical traces per seed.

**Reproduce (from repo root):**

```bash
uv sync
# the PR's own tests: plugin conformance + adversarial suites + end-to-end scenario (39 passed)
uv run pytest packages/nest-plugins-reference/tests/test_intent_facts.py \
              packages/nest-core/tests/test_intent_gated_datafacts_scenario.py -q
# discriminator, gated side (both validators PASS):
uv run nest run scenarios/intent_gated_datafacts.yaml --seed 42 -o trace.jsonl
# baseline side (both validators FAIL): edit the yaml, set layers.datafacts: datafacts_v1, rerun.
# trace sha256[:16] reproduces as c6dc69a6... (intent_facts) / ec93269f... (datafacts_v1) on seeds 42/7/1337.
```

## Links

- Live service: https://agentintent.onrender.com
- SKILL.md: https://agentintent.onrender.com/SKILL.md
- Protocol client + service repo: https://github.com/midhunrajcharles/AgentIntent

**Branch-name note:** I attempted the charter rename to `hackathon/midhunrajcharles-intent-facts`, but renaming the head branch of a cross-repo PR deletes its head ref and auto-closes the PR (this PR was briefly closed and reopened in the process). I kept the original branch name to preserve this review thread — happy to open a freshly-named PR instead if you prefer.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -50,6 +50,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("privacy", "trust_gated"): f"{_REF}.privacy.trust_gated:TrustGatedPrivacy",
     ("datafacts", "datafacts_v1"): f"{_REF}.datafacts.datafacts_v1:DataFactsV1",
     ("datafacts", "cid_facts"): f"{_REF}.datafacts.cid_facts:CidFacts",
+    ("datafacts", "intent_facts"): f"{_REF}.datafacts.intent_facts:IntentGatedFacts",
     ("failure_detector", "heartbeat"): (
         f"{_REF}.failure_detection.heartbeat:HeartbeatFailureDetector"
     ),
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -175,3 +175,9 @@ def _try_load_builtin(name: str) -> None:
         from nest_core.scenarios_builtin.sybil_bond import sybil_bond_factory
 
         register_scenario("sybil_bond", sybil_bond_factory)
+    elif name == "intent_gated_datafacts":
+        from nest_core.scenarios_builtin.intent_gated_datafacts import (
+            intent_gated_datafacts_factory,
+        )
+
+        register_scenario("intent_gated_datafacts", intent_gated_datafacts_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/intent_gated_datafacts.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/intent_gated_datafacts.py
@@ -1,0 +1,192 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Intent-gated datafacts scenario: publish only after a pre-declared intent.
+
+Three honest agents each **register a publish intent** for their dataset and then
+publish it; a fourth agent is an **attacker** that skips the intent step and tries
+to publish a surprise dataset directly. Every step is reported as a ``|``-delimited
+trace message so ``validate_trace(..., "intent_gated_datafacts")`` can read it.
+
+The scenario is a discriminator. Point ``layers.datafacts`` at ``intent_facts``
+(the ``IntentGatedFacts`` plugin) and both validators pass: honest publishes are
+each preceded by a registered intent, and the attacker's un-declared publish is
+blocked. Point it at ``datafacts_v1`` -- which has no intent concept -- and both
+flip to FAIL: there is no intent gate, so the attacker's surprise publication
+succeeds silently and no publish is backed by a declared intent.
+
+Example::
+
+    agents = intent_gated_datafacts_factory(config, plugins)
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from nest_core.scenario import ScenarioConfig
+from nest_core.sim.agent import AgentContext, StateMachineAgent
+from nest_core.types import AgentId, DatasetMetadata
+
+
+class HonestPublisher(StateMachineAgent):
+    """Declares a publish intent, then publishes its dataset within the TTL.
+
+    Against a gated plugin the intent is registered first (emitting
+    ``intent_registered``) and the publish then consumes it (``publish_ok``).
+    Against a plugin with no intent concept the registration is silently
+    skipped -- the ``publish_ok`` then has no backing intent, which is exactly
+    what the validator catches.
+
+    Example::
+
+        pub = HonestPublisher(AgentId("supplier-0"), name="prices")
+    """
+
+    def __init__(self, agent_id: AgentId, name: str) -> None:
+        self._id = agent_id
+        self._name = name
+
+    async def on_start(self, ctx: AgentContext) -> None:
+        """Register intent (if the plugin supports it), then publish the dataset.
+
+        Example::
+
+            await pub.on_start(ctx)
+        """
+        facts = ctx.plugins.get("datafacts")
+        if facts is None:
+            return
+        register = getattr(facts, "register_publish_intent", None)
+        if callable(register):
+            register(self._name)
+            await ctx.send(self._id, f"intent_registered|honest|{self._name}|{self._id}".encode())
+        dataset = DatasetMetadata(name=self._name, owner=self._id)
+        try:
+            url = await facts.publish(dataset)
+        except ValueError as exc:  # IntentError (a ValueError) if the gate blocks it
+            await ctx.send(
+                self._id, f"publish_blocked|honest|{self._name}|{self._id}|{exc}".encode()
+            )
+            return
+        await ctx.send(self._id, f"publish_ok|honest|{self._name}|{self._id}|{url}".encode())
+
+
+class AttackerPublisher(StateMachineAgent):
+    """Skips the intent step and attempts a surprise publication.
+
+    A gated plugin raises (a ``ValueError``) because no intent was declared, and
+    the attempt is reported as ``publish_blocked``. A plugin with no gate lets
+    the surprise land as ``publish_ok`` -- the failure the discriminator surfaces.
+
+    Example::
+
+        atk = AttackerPublisher(AgentId("attacker-0"), name="surprise_release")
+    """
+
+    def __init__(self, agent_id: AgentId, name: str) -> None:
+        self._id = agent_id
+        self._name = name
+
+    async def on_start(self, ctx: AgentContext) -> None:
+        """Attempt to publish without declaring any intent first.
+
+        Example::
+
+            await atk.on_start(ctx)
+        """
+        facts = ctx.plugins.get("datafacts")
+        if facts is None:
+            return
+        dataset = DatasetMetadata(name=self._name, owner=self._id)
+        try:
+            url = await facts.publish(dataset)
+        except ValueError as exc:  # IntentError (a ValueError) when the gate blocks it
+            await ctx.send(
+                self._id, f"publish_blocked|attacker|{self._name}|{self._id}|{exc}".encode()
+            )
+            return
+        await ctx.send(self._id, f"publish_ok|attacker|{self._name}|{self._id}|{url}".encode())
+
+
+def _build_datafacts_handles(
+    datafacts_cls: type[Any],
+    identities: dict[AgentId, Any],
+    all_ids: list[AgentId],
+) -> dict[AgentId, Any]:
+    """Give each agent its own datafacts handle over shared storage where possible.
+
+    Plugins that take an ``Identity`` plus ``datasets``/``proofs``/``clock``
+    keyword arguments (``intent_facts``, ``cid_facts``) get one handle per agent
+    over the same dicts and logical clock, so each agent signs and registers
+    intents as itself while sharing the published-dataset store. Plugins with a
+    no-argument constructor (``datafacts_v1``) get a single shared instance.
+
+    Example::
+
+        handles = _build_datafacts_handles(cls, identities, all_ids)
+    """
+    shared_datasets: dict[Any, Any] = {}
+    shared_proofs: dict[Any, Any] = {}
+    shared_clock: Any = None
+    shared_instance: Any = None
+    handles: dict[AgentId, Any] = {}
+
+    for aid in all_ids:
+        try:
+            kwargs: dict[str, Any] = {"datasets": shared_datasets, "proofs": shared_proofs}
+            if shared_clock is not None:
+                kwargs["clock"] = shared_clock
+            handle = datafacts_cls(identities[aid], **kwargs)
+            shared_clock = getattr(handle, "clock", shared_clock)
+            handles[aid] = handle
+        except TypeError:
+            if shared_instance is None:
+                shared_instance = datafacts_cls()
+            handles[aid] = shared_instance
+    return handles
+
+
+def intent_gated_datafacts_factory(
+    config: ScenarioConfig,
+    plugins: dict[str, Any],
+) -> dict[AgentId, StateMachineAgent]:
+    """Create three honest publishers plus one surprise-publishing attacker.
+
+    Instantiates per-agent identities (so each agent registers and signs as
+    itself) and wires the resolved ``datafacts`` plugin class into per-agent
+    handles via :func:`_build_datafacts_handles`.
+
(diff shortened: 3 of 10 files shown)


midhunrajcharles:
Update: the intent gate is now wired as a **runnable discriminator**, matching the datafacts scenario convention (`provenance_supply_chain` / `cid_facts`).

- New builtin scenario `intent_gated_datafacts` (3 honest publishers declare a publish intent then publish; an attacker skips the intent step), registered in `scenarios._try_load_builtin`.
- Two trace-level validators registered under `intent_gated_datafacts` in `nest_core/validators.py`: `validate_intent_no_surprise_publication` and `validate_intent_gate_blocks_attacker`.

**Discriminator, verified end-to-end** on seeds 42/7/1337 (byte-deterministic per seed):

```
layers.datafacts = intent_facts
  PASS intent_no_surprise_publication - all 3 publications were backed by a declared intent
  PASS intent_gate_blocks_attacker    - attacker's intent-less publish was blocked by the gate

layers.datafacts = datafacts_v1
  FAIL intent_no_surprise_publication - supplier/manufacturer/retailer/attacker published with no declared intent
  FAIL intent_gate_blocks_attacker    - attacker published 'surprise_release' with no declared intent
```

Rebased onto current `main` (conflict-free). Full Definition of Done green on the rebased tree: `ruff check`, `ruff format --check`, `pyright` (0 errors), `pytest` (1016 passed). Purely additive.

dhve:
Thanks @midhunrajcharles. The core of this is genuinely well built. `IntentGatedFacts` extends the merged `CidFacts` instead of forking it, the TTL runs on the shared logical clock so there is no wall-clock anywhere in the sim path, and `test_intent_gated_datafacts_scenario.py` is the strongest part of the PR: it runs the real `ScenarioRunner` both ways, asserts both validators fail against `datafacts_v1` and pass against `intent_facts`, and checks byte-identical traces across seeds 42, 7, and 1337. That is exactly the fails-on-baseline, passes-on-yours pattern the charter asks for, and very few submissions actually test it end to end. The `_agent_plugins` override channel in the factory matches how `gossip_registry` and `empic_payments` do per-agent handles, so the wiring is idiomatic too.

**1. The branch conflicts with main, so CI has never run.**

GitHub reports `mergeable: CONFLICTING` and no checks on the branch. Merged against today's main, the conflicts land in `nest_core/scenarios.py` (the `_try_load_builtin` chain every scenario PR appends to) and `tests/test_validators.py` (the expected set in `test_all_scenario_types_registered`); your `validators.py` and `plugins.py` hunks auto-merge cleanly. Rebase onto main and re-push; until then the pipeline cannot even start, and the judge panel does not score broken PRs.

**2. The AgentIntent HTTP client is disconnected from the submission and leans on a dependency nothing declares.**

Nothing in the plugin, scenario, or validators imports `integration/agent_intent_client.py`; it only exists alongside its mocked-transport tests. And no workspace package declares `httpx`: it is importable today only because `nest-shell` hard-depends on `openai`, which drags `httpx` in transitively. The `pytest.importorskip("httpx")` guard in `test_agent_intent_integration.py` already hints at that fragility; if that transitive edge ever changes, an undeclared import inside `nest-plugins-reference` breaks. The bigger issue is scope: the charter asks for one problem per PR (plugin plus adversarial validator plus scenario), and a REST client for an external service is a second project living in the same diff. Its 33 tests exercise `_FakeTransport` route matching, not the live service, so they add bulk without adding evidence for the datafacts story. My honest recommendation: cut `integration/` and `test_agent_intent_integration.py` from this PR and keep them in your AgentIntent repo (link it in the body, as you already do). The plugin + scenario + validators stand perfectly well on their own.

**3. The intent gate never checks `dataset.owner` against the instance owner.**

`_consume_intent` keys on `(self._owner_str, dataset.name)` and neither your `publish()` nor `CidFacts.publish()` compares `dataset.owner` to the publishing identity. So agent B can call `register_publish_intent("weather")` on its own instance and then `publish(DatasetMetadata(name="weather", owner=AgentId("a1")))`: the gate passes, the spoofed dataset lands in the shared store, and B's intent log marks it fulfilled. `verify_freshness` catches the mismatch later (signer != owner), but the PR body sells intent-hijack as blocked at publish time, and `TestIntentHijackAttack` only covers A-publishing-without-an-intent, not B-claiming-A's-ownership. One-line fix in `publish()`: raise `IntentError` when `str(dataset.owner) != self._owner_str`, plus an adversarial test for exactly this spoof. That closes the gap and makes the body's claim true.

**4. `identity._agent_id` is a private-attribute reach, and the entry point is missing.**

`self._owner_str = str(identity._agent_id)` with a `type: ignore` couples you to `DidKeyIdentity` internals; the `Identity` Protocol declares no agent id, and `Ed25519RotatingIdentity` exposes it as a public `agent_id` property instead. Take the owner `AgentId` as an explicit constructor argument (or `getattr(identity, "agent_id", ...)` with a fallback). Separately, `cid_facts` is registered both in the `nest_core` builtins and under `[project.entry-points."nest.plugins.datafacts"]` in `packages/nest-plugins-reference/pyproject.toml`; add an `intent_facts` line there too, api fit looks at that table.

**5. Name the problem and tighten the body.**

The charter asks for exactly one problem from `docs/hackathon/problems/`, and the body never names one. The natural home is 08 (datafacts), whose core shipped in merged #31, which is literally your base class. Extending merged work with a new attack class (surprise publication) is a legitimate novelty story, but you have to make it explicitly: one paragraph on what the intent gate catches that #31 cannot. While you are in there: the body says 7 files +1528 but the PR is 12 files +2001, and the charter's branch pattern is `hackathon/<handle>-<short-theme>`, so `hackathon/midhunrajcharles-intent-facts`; mismatched names are subject to auto-close. Renaming the branch through the GitHub UI keeps the PR attached.

**Suggested path forward:**

1. Rebase onto main and resolve the `scenarios.py` and `test_validators.py` conflicts.
2. Drop the `integration/` module and its test file from this PR; link the AgentIntent repo in the body instead.
3. Add the owner check in `publish()` plus the owner-spoof adversarial test.
4. Pass the agent id explicitly and register the pyproject entry point.
5. Rewrite the body around problem 08 and the delta over #31, fix the file counts, rename the branch.
6. Run the full local gate (`uv sync`, `ruff check`, `ruff format --check`, `pyright`, `pytest -v`) and confirm green before requesting re-review.

Leaving this open. The plugin core is strong; trimmed to the datafacts submission it wants to be, it makes a clean, convincing case. Ping me on the hackathon channel if any of the reshaping is unclear.

midhunrajcharles:
@dhve Thanks for the thorough review — all five points are addressed and pushed (head `244000f`), and the body is rewritten around problem 08.

1. **Rebase** — rebased onto current main (`7106ca6`); the `scenarios.py` / `test_validators.py` conflicts are resolved and the tree is purely additive again. Full local gate green on the rebased tree: `ruff check`, `ruff format --check`, `pyright` strict (0 errors), `pytest` **1189 passed**.
2. **Scope** — `integration/agent_intent_client.py` and its 33 mocked-transport tests are out of this PR (and with them the undeclared transitive `httpx` edge). The protocol client lives in the external AgentIntent repo, linked in the body.
3. **Owner-spoof** — `publish()` now raises `IntentError` when `str(dataset.owner) != self._owner_str`, checked *before* the intent is consumed, so a spoofed publish neither lands in the shared store nor burns the attacker's live intent. Two adversarial tests cover exactly the B-claims-A's-ownership case you described (blocked spoof + intent preservation), and the module docstring now describes the gate honestly.
4. **Identity coupling + entry point** — the `identity._agent_id` reach is gone: the owner comes from an explicit `owner=` constructor argument or the public `agent_id` property (test added against `Ed25519RotatingIdentity`; `DidKeyIdentity` resolves via a guarded fallback). `intent_facts` is now registered in the `nest.plugins.datafacts` entry-point table alongside `cid_facts`.
5. **Body/branch** — body rewritten: problem 08 named, the delta over #31 stated up front, counts corrected (10 files, +1221 / −0). One honest caveat on the branch rename: renaming the head branch of a cross-repo PR deletes its head ref and auto-closes the PR — this PR briefly closed when I attempted it, so I restored the original branch name to keep this review thread intact. If you'd rather have the charter-compliant name, say the word and I'll open a freshly-named PR referencing this one.

Discriminator re-verified end-to-end after the changes: both validators PASS on `intent_facts` and FAIL on `datafacts_v1`, byte-identical traces per seed on 42 / 7 / 1337. Ready for re-review.

midhunrajcharles:
Discriminator evidence at head `244000f` (post-review-fix), run via `nest run` + `validate_trace(..., "intent_gated_datafacts")` on seeds 42 / 7 / 1337:

```
layers.datafacts = intent_facts      trace sha256[:16] = c6dc69a66f719c18  (identical for all three seeds)
  PASS intent_no_surprise_publication - all 3 publications were backed by a declared intent
  PASS intent_gate_blocks_attacker    - attacker's intent-less publish was blocked by the gate

layers.datafacts = datafacts_v1      trace sha256[:16] = ec93269ff0965343  (identical for all three seeds)
  FAIL intent_no_surprise_publication - supplier-0/manufacturer-0/retailer-0/attacker-0 published with no declared intent
  FAIL intent_gate_blocks_attacker    - attacker published 'surprise_release' with no declared intent
```

The scenario is a pure state-machine path, so the trace is byte-identical not just per seed but across seeds — the sha256 above reproduces with `uv run nest run scenarios/intent_gated_datafacts.yaml --seed <s> -o trace.jsonl` (swap `layers.datafacts` to `datafacts_v1` for the baseline run).

JamesCarnley:
Thanks @midhunrajcharles for the real simulator discriminator and for addressing the earlier review. I reviewed all ten changed files at `244000f`, the July discussion, and rebuilt Town `main@f3664a8`. The healthy/blocked controls and owner-check-before-consumption regression are worth retaining.

Closing this archived implementation without a current-core port. It extends the retired CID-addressed dataset publication interface. Current Town's `EvidenceRecords.attest` produces observer-attributed facts; it is not that dataset publication API. Requiring every observer to self-declare a publication intent would add an application policy, not automatically improve evidence validity.

Preserve this as an optional publication-policy profile:
- Define who authorizes publication and what exact owner, content, action and expiry that permission covers. A publisher declaring its own intent is not independent approval.
- Retain missing/wrong-name/expired/owner-mismatch/reuse controls and successful publish plus independent fetch verification.
- Define atomic consumption and failure recovery. The legacy code consumes and marks an intent fulfilled before the parent publication call; the failed-parent test checks rejection but not the resulting intent state.
- Extend the trace evaluator to enforce one-time use, expiry, exact expected publishers/attempts and content binding. Its current set of previously declared names cannot establish those stronger properties.
- Keep rejected-owner attempts from consuming a valid intent, and cover duplicate declarations and retry/history behavior.

The existing tests genuinely swap the selected plugin, which is stronger than a synthetic pass/fail flag. This is retained design/test material, not a claim that an equivalent publication gate was merged into current Town. No legacy code or external service was executed.

midhunrajcharles:
Thank you for the thorough review, @JamesCarnley . Closing this makes sense. The
implementation targets the retired CID-addressed publication API, and
`EvidenceRecords.attest` in current Town isn't that API, so a direct port would
misrepresent what this does.

Two items on your list are genuine defects rather than scoping comments:

- Self-declaration is not authorization. The gate only demonstrates that the
  publisher stated its intent; it establishes nothing about approval. A usable
  profile needs a separate authorizer and a permission bound to the exact owner,
  content, action and expiry.
- Consumption ordering is incorrect. The intent is consumed and marked fulfilled
  before the parent publish call, so a failed publish burns an otherwise valid
  intent. The failed-parent test asserts only that the publish is rejected, never
  the resulting intent state.

Agreed on the trace evaluator and a set of previously declared names cannot
establish one-time use, expiry, expected publisher/attempt counts, or content
binding.

I'll leave this as design and test material, with no claim that an equivalent
gate exists in current Town. If a publication-policy profile is wanted later,
I'm happy to open a follow-up issue scoped against current core.

PR #132, opened 2026-07-09
auth: macaroon-style delegatable capability tokens with cascading revocation

TanishDevs11:
## Motivation

The default auth plugin (`nest_plugins_reference/auth/jwt_auth.py`) is 106 lines of HMAC-signed token issuance. It has no notion of delegation: `JwtAuth._revoked: set[str]` tracks revocation by exact token string, with no parent-child relationship between tokens at all. That makes the auth layer unable to model the most common real-world capability pattern — "agent A holds a long-lived root token, mints a 10-minute sub-token for agent B without going back to the issuer, and revoking A's token invalidates B's at the next verify, automatically." Macaroons (Google, 2014), biscuits (CleverCloud), and SPIFFE delegation all do this; `docs/layers/auth.md` explicitly lists "capability delegation" and "revocation propagation" as wanted. Solves `docs/hackathon/problems/04-auth-capability-delegation.md`.

Anyone building a multi-agent workflow where an orchestrator hands out narrowly-scoped, time-bounded sub-capabilities needs this — e.g. an LLM agent sub-renting tool access to a worker, and being able to withdraw it cleanly without hunting down every descendant token.

## Design

`DelegatableAuth` (`packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable.py`) borrows the caveat-chaining trick from macaroons (Birgisson et al., 2014): a token is not one HMAC over one flat payload, it's an ordered chain of caveats, where each caveat's signature is keyed off the **previous caveat's own signature**, not the root secret:

```
sig_0 = HMAC(root_secret, caveat_0)   # issue()
sig_1 = HMAC(sig_0,       caveat_1)   # delegate()
sig_2 = HMAC(sig_1,       caveat_2)   # delegate() again
```

Two properties fall out of that for free:

1. **Delegation needs no issuer round-trip.** `delegate()` never reads `self._secret`. It only needs the parent token's own trailing signature (which travels with the token in plain sight) to key the next caveat, plus a local, secret-free check that no ancestor in the chain is in the revocation table. This is the anti-pattern the problem doc calls out explicitly: "don't ship delegation that's actually just re-issuance by the central authority." A forged/tampered parent isn't rejected at delegate time — it's still caught the next time anyone calls `verify()`, when the real root secret replays the whole chain.
2. **Revocation is transitive by construction.** Every token is self-contained — it carries its *entire* ancestor chain, not a pointer to one. `verify()` replays the whole chain from `self._secret` and, at every level, recomputes that ancestor's *exact* original token string and checks its digest against `self._revoked: dict[str, bool]`. That dict only ever gains an entry for a token *directly* passed to `revoke()` — a child's entry is never written. Revoking a parent invalidates every descendant the next time any of them is verified, with zero bookkeeping proportional to subtree size.

On top of the base `Auth` protocol (`issue`/`verify`/`revoke`, unchanged signatures), it adds:
- `delegate(parent_token, audience, scopes_subset, ttl) -> Token` — `scopes_subset` must be a **strict** (proper) subset of the parent's scopes (`ScopeEscalationError` otherwise — every hop must shed at least one capability, not just avoid escalating), and `ttl` must land the child's expiry at or before the parent's own (`DelegationTtlError` otherwise, which also naturally rejects delegating from an already-expired parent).
- `verify_presented_by(token, presenter) -> AuthContext` — the audience-aware sibling of `verify()`. The base `Auth.verify(token)` signature can't check *who* presents a token since it only receives the token; this adds that check (`AudienceMismatchError`) on top without breaking the base contract.

## Adversarial validator

`packages/nest-plugins-reference/nest_plugins_reference/validators/auth_delegation_validators.py` ships three checks, each deliberately **plugin-agnostic** — they take async callables the caller builds however that plugin exposes the relevant operation, and only inspect the outcome:

1. `check_scope_escalation_rejected` — attempt to mint/verify an over-scoped child.
2. `check_stale_parent_rejected` — revoke the parent *after* a child already exists, then re-verify the child.
3. `check_audience_confusion_rejected` — present a token as an agent other than its declared audience.

Wired against `JwtAuth` (whose only "delegation" surface is calling `issue()` again, with no parent to compare against and no audience concept at all) all three **fail**. Wired against `DelegatableAuth`'s real `delegate`/`verify`/`verify_presented_by` surface, all three **pass**. The differential proof is in `test_delegatable_auth.py` (`test_*_validator_fails_against_jwt` / `test_*_validator_passes_against_delegatable`, 6 tests total) — the same validator function, run against both plugins, produces the charter's required opposite verdicts.

## Scenario

`scenarios/delegated_auth.yaml` + `packages/nest-core/nest_core/scenarios_builtin/delegation_tree.py`: a strict tree — 1 coordinator → 3 intermediaries → 12 leaves (4 per intermediary). The coordinator issues a root token, delegates a narrower token to each intermediary, which each delegate a further-narrowed token to their 4 leaves. Mid-run, the coordinator revokes **exactly one** intermediary's token; all 4 leaves under that branch fail re-verification with `RevokedAncestorError` while the other 8 keep verifying — without any leaf token ever being revoked directly. The scenario factory pins the plugin's clock to a fixed value (both `DelegatableAuth` and `JwtAuth` accept `clock=`) instead of wall-clock time, so the trace is byte-identical run over run for the same seed.

## Tradeoffs / out of scope

- **Strict tree, not a DAG.** The problem doc explicitly says a strict tree (single parent per token) is sufficient, and it keeps the "recompute every ancestor's digest by replaying the embedded chain" verification model simple — a DAG would need a different revocation-check strategy (e.g. checking all parent chains a token could route through). Out of scope here, as the problem doc allows.
- **"Strict" subset interpretation.** The problem doc's success criteria say scopes must be a "strict subset" of the parent's; I enforce that literally (proper subset — equal scopes are rejected, not just supersets), on the theory that every hop should shed at least one capability. This is called out in the plugin's docstring in case reviewers read "strict" more loosely.
- **Out of scope per the problem doc:** hardware-backed key attestation (TPM/HSM), OAuth2 server endpoints (pure in-process is fine), token introspection over the network. Not attempted.
- **Shared-instance architecture.** Like `JwtAuth`, a single `DelegatableAuth` instance is meant to be shared by every agent in a scenario (mirrors how the existing `_instantiate_plugins` shared-instance pattern works for `registry`/`trust`/`payments`). `delegate()`'s independence from the root secret is a real cryptographic property (verified by `test_delegate_does_not_require_the_root_secret`, using two instances with different secrets), not just an artifact of instance-sharing.

## Verification snippet

Paste into `uv run python -c "..."` (or a Python shell) from the repo root:

```python
import asyncio
from nest_core.types import AgentId
from nest_plugins_reference.auth.delegatable import DelegatableAuth, RevokedAncestorError

async def main():
    auth = DelegatableAuth(secret=b"town-secret", clock=1000.0)

    # Orchestrator issues itself a root capability.
    root = await auth.issue(AgentId("orchestrator"), ["read", "write", "admin"])

    # Orchestrator delegates a narrower, shorter-lived token to a worker --
    # no call back to any issuer, just the root token it already holds.
    worker_token = await auth.delegate(root, AgentId("worker-1"), ["read", "write"], ttl=300.0)
    ctx = await auth.verify_presented_by(worker_token, AgentId("worker-1"))
    print("worker verified with scopes:", ctx.scopes)

    # Worker sub-delegates an even narrower token to a helper.
    helper_token = await auth.delegate(worker_token, AgentId("helper-1"), ["read"], ttl=60.0)
    ctx2 = await auth.verify_presented_by(helper_token, AgentId("helper-1"))
    print("helper verified with scopes:", ctx2.scopes)

    # Revoking the ROOT invalidates the worker AND the helper transitively --
    # neither token's own digest was ever written to the revocation table.
    await auth.revoke(root)
    for label, token in [("worker", worker_token), ("helper", helper_token)]:
        try:
            await auth.verify(token)
            print(f"{label}: still valid (BUG)")
        except RevokedAncestorError as exc:
            print(f"{label}: cascaded correctly -> {exc}")

asyncio.run(main())
```

Expected output:
```
worker verified with scopes: ['read', 'write']
helper verified with scopes: ['read']
worker: cascaded correctly -> ancestor at depth 0 (subject='orchestrator') has been revoked
helper: cascaded correctly -> ancestor at depth 0 (subject='orchestrator') has been revoked
```

## Testing

`packages/nest-plugins-reference/tests/test_delegatable_auth.py` — 31 tests: base `Auth` protocol conformance (round-trip, tamper/malformed/expired/cross-secret rejection), successful multi-level delegation, the three named attacks (scope escalation — including a Hypothesis property test, TTL violation, audience confusion) each individually, cascading revocation (parent → child, grandparent → grandchild without touching the grandchild, sibling-branch-unaffected), the six-test validator differential proof, and full scenario integration (cascading revocation lands on exactly the revoked branch's 4 leaves; byte-identical trace across two runs of the same seed).

Local CI (`uv sync && uv run ruff check . && uv run ruff format --check . && uv run pyright && uv run pytest -v`) is green: 976 passed, 1 skipped (pre-existing, unrelated), 1 deselected (`live` marker).

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -26,6 +26,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("registry", "in_memory"): f"{_REF}.registry.in_memory:InMemoryRegistry",
     ("registry", "gossip"): f"{_REF}.registry.gossip:GossipRegistry",
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
+    ("auth", "delegatable"): f"{_REF}.auth.delegatable:DelegatableAuth",
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
     ("trust", "agent_receipts"): f"{_REF}.trust.agent_receipts:AgentReceiptsTrust",
     ("trust", "parc"): f"{_REF}.trust.parc:ParcTrust",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -145,3 +145,7 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("rogue_trusted_agent", rogue_trusted_agent_factory)
+    elif name == "delegation_tree":
+        from nest_core.scenarios_builtin.delegation_tree import delegation_tree_factory
+
+        register_scenario("delegation_tree", delegation_tree_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/delegation_tree.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/delegation_tree.py
@@ -1,0 +1,243 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Delegation-tree scenario -- capability delegation with cascading revocation.
+
+A coordinator issues a root capability token and delegates a narrower,
+shorter-lived token to each of 3 intermediaries; each intermediary
+delegates a further-narrowed token to 4 leaves. Strict tree, one parent per
+token: coordinator -> 3 intermediaries -> 12 leaves.
+
+Midway through the run the coordinator revokes exactly one intermediary's
+token. All 4 leaves under that branch must fail re-verification with
+``RevokedAncestorError`` on their *next* verify -- no leaf token is ever
+revoked directly, demonstrating cascading revocation "by construction". The
+other 8 leaves, under the two untouched branches, keep verifying.
+
+Example::
+
+    agents = delegation_tree_factory(config, plugins)
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any, cast
+
+from nest_plugins_reference.auth.delegatable import RevokedAncestorError
+
+from nest_core.scenario import ScenarioConfig
+from nest_core.sim.agent import AgentContext, StateMachineAgent
+from nest_core.types import AgentId, Token
+
+if TYPE_CHECKING:
+    from nest_plugins_reference.auth.delegatable import DelegatableAuth
+
+GRANT_PREFIX = b"grant:"
+RESULT_PREFIX = b"result:"
+REVERIFY = b"reverify"
+REVOKE_TICK = b"COORD_REVOKE_TICK"
+REVERIFY_TICK = b"COORD_REVERIFY_TICK"
+
+#: Scopes narrow strictly at every hop: root -> intermediary -> leaf.
+ROOT_SCOPES = ["read", "write", "admin"]
+INTERMEDIARY_SCOPES = ["read", "write"]
+LEAF_SCOPES = ["read"]
+
+#: TTLs narrow at every hop too, so a child never risks outliving its parent.
+INTERMEDIARY_TTL = 1800.0
+LEAF_TTL = 900.0
+
+
+class CoordinatorAgent(StateMachineAgent):
+    """Root of the delegation tree.
+
+    Issues the root token, delegates one child per intermediary, then --
+    on a scheduled tick -- revokes exactly one intermediary's token and
+    asks every leaf to re-verify, collecting the outcomes into a shared
+    report dict for the scenario factory (and tests) to inspect.
+    """
+
+    def __init__(
+        self,
+        agent_id: AgentId,
+        *,
+        intermediary_ids: list[AgentId],
+        all_leaf_ids: list[AgentId],
+        revoke_intermediary: AgentId,
+        revoke_delay: float,
+        reverify_delay: float,
+        report: dict[str, str],
+    ) -> None:
+        self._id = agent_id
+        self._intermediary_ids = intermediary_ids
+        self._all_leaf_ids = all_leaf_ids
+        self._revoke_intermediary = revoke_intermediary
+        self._revoke_delay = revoke_delay
+        self._reverify_delay = reverify_delay
+        self._report = report
+        self._intermediary_tokens: dict[AgentId, Token] = {}
+
+    async def on_start(self, ctx: AgentContext) -> None:
+        auth = cast("DelegatableAuth", ctx.plugins["auth"])
+        root_token = await auth.issue(self._id, list(ROOT_SCOPES))
+        for inter_id in self._intermediary_ids:
+            child = await auth.delegate(
+                root_token, inter_id, list(INTERMEDIARY_SCOPES), INTERMEDIARY_TTL
+            )
+            self._intermediary_tokens[inter_id] = child
+            await ctx.send(inter_id, GRANT_PREFIX + str(child).encode())
+        await ctx.schedule(self._revoke_delay, REVOKE_TICK)
+
+    async def on_message(self, ctx: AgentContext, sender: AgentId, payload: bytes) -> None:
+        if sender == ctx.agent_id and payload == REVOKE_TICK:
+            auth = cast("DelegatableAuth", ctx.plugins["auth"])
+            await auth.revoke(self._intermediary_tokens[self._revoke_intermediary])
+            await ctx.schedule(self._reverify_delay, REVERIFY_TICK)
+            return
+        if sender == ctx.agent_id and payload == REVERIFY_TICK:
+            for leaf_id in self._all_leaf_ids:
+                await ctx.send(leaf_id, REVERIFY)
+            return
+        if payload.startswith(RESULT_PREFIX):
+            _, status, leaf_id = payload.decode("utf-8").split(":", 2)
+            self._report[leaf_id] = status
+
+
+class IntermediaryAgent(StateMachineAgent):
+    """Receives a delegated token from the coordinator and re-delegates to its leaves."""
+
+    def __init__(self, agent_id: AgentId, *, leaf_ids: list[AgentId]) -> None:
+        self._id = agent_id
+        self._leaf_ids = leaf_ids
+
+    async def on_message(self, ctx: AgentContext, sender: AgentId, payload: bytes) -> None:
+        if not payload.startswith(GRANT_PREFIX):
+            return
+        token = Token(payload[len(GRANT_PREFIX) :].decode("utf-8"))
+        auth = cast("DelegatableAuth", ctx.plugins["auth"])
+        await auth.verify_presented_by(token, ctx.agent_id)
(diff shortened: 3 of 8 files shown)


dhve:
Thanks @TanishDevs11. This is one of the strongest Problem 04 entries I have read. The core design decision, that `delegate()` never reads `self._secret` and keys each new caveat off the parent's trailing signature, is the actual macaroon insight rather than re-issuance dressed up as delegation, and `test_delegate_does_not_require_the_root_secret` proves it properly with two instances holding different secrets. The six-test differential proof (each validator failing against `JwtAuth` and passing against `DelegatableAuth`) is exactly the charter's bar, and the scenario's report dict trick makes the cascading-revocation outcome inspectable without trace parsing.

Not mergeable as it stands though. Four things to work through:

**1. The branch conflicts with main and CI has never run.**

`gh` reports `mergeable: CONFLICTING` and no checks at all on `hackathon/Tanish-capability-delegation`. You touched three registration files that most other hackathon PRs also touch (`nest_core/plugins.py`, `nest_core/scenarios.py`, `validators/__init__.py`), so this will keep recurring as other PRs land. Rebase onto main, resolve, push, and confirm the five CI stages go green. Per the charter, the judge panel does not score PRs that are not green, so everything else here is moot until this is done.

**2. The audience check can be laundered through a single delegation hop.**

`verify_presented_by()` correctly rejects a stolen token presented by a non-audience agent. But `delegate()` is offline and unauthenticated by design, so a thief holding worker-1's token can call `delegate(stolen, AgentId("thief"), narrower, ttl)` and mint a perfectly valid child whose declared audience is the thief. That child sails through `verify_presented_by(child, thief)`. The cost is shedding one scope per hop, and note the edge case: since the empty set is a strict subset, a stolen single-scope token can still be laundered into a zero-scope child (is a scope-less token something `delegate()` should refuse outright?). This is an inherent property of offline attenuation of bearer tokens, macaroons have it too, but the PR body and the `check_audience_confusion_rejected` docs currently read as if audience binding is airtight. Either document the boundary explicitly in `delegatable.py`'s docstring and pin it with a test that demonstrates the laundering path, or explain in the PR body why you consider it out of scope. Honesty about the threat model here is a persona-fidelity and docs win, not a weakness.

**3. The validators count any exception as a pass, with no positive control.**

All three checks in `auth_delegation_validators.py` treat any raised exception as "attack rejected". A plugin whose `delegate` and `verify` unconditionally raise `NotImplementedError` would pass all three while being unusable. Your test suite happens to cover the legitimate paths separately, but the validators are advertised as plugin-agnostic and exported from `validators/__init__.py` for anyone to reuse, so give each one a positive control: a callable exercising the legitimate operation (properly narrowed delegation, correct presenter) that must succeed for the check to pass. That makes "fail-closed because broken" score as a failure, which is what a judge running the validator against a third plugin would want.

**4. Small api_fit polish: register the entry point.**

The problem doc sanctions the `plugins.py` builtin-table registration you did, but `packages/nest-plugins-reference/pyproject.toml` calls entry points the documented install-time discovery path, and the rubric's api_fit dimension looks for it. One stanza: `[project.entry-points."nest.plugins.auth"]` with `delegatable = "nest_plugins_reference.auth.delegatable:DelegatableAuth"`. You would be creating the auth group; that is fine.

One competitive note: there are twenty-plus open PRs on this exact problem. Your body already differentiates well; keep the secret-free delegation proof and the strict-subset stance front and center, since novelty is scored against the field.

**Suggested path forward:**

1. Rebase onto main, resolve the conflicts (the shared registration files are the usual culprits), push, confirm CI green.
2. Add the laundering-boundary docstring plus a pinning test, and decide on empty `scopes_subset`.
3. Add a positive-control callable to each of the three validators.
4. Add the pyproject entry point for `("auth", "delegatable")`.

TanishDevs11:
Working through the review now — item 1 (rebase) surfaced something the review
didn't anticipate, flagging before I touch anything.

The conflicts in `delegatable.py`, `test_delegatable_auth.py`, and
`scenarios/delegated_auth.yaml` aren't "two PRs touched nearby lines" —
they're add/add collisions with PR #138 (ckeda-delegatable-auth), which
merged to main after I opened this PR. That PR independently built:

- the same file path: nest_plugins_reference/auth/delegatable.py
- the same class name: DelegatableAuth
- the same registry key: ("auth", "delegatable") in nest_core/plugins.py
- overlapping exception names: DelegationError, ScopeEscalationError,
  RevokedAncestorError, AudienceMismatchError

Two structurally different implementations (theirs: 318 lines, mine: 483)
are now claiming the identical slot. Resolving this by just taking "mine"
during the rebase would silently overwrite #138's merged behavior for
anyone importing that path — not a normal merge-conflict resolution.

Before I push a rebase, I'd like guidance on which of these the judges
would prefer:
(a) I rename my plugin/module/registry key/scenario id to coexist
    alongside #138 (e.g. ("auth", "capability_delegation")), or
(b) some other resolution you'd prefer (e.g. treat this as a duplicate
    submission for the problem and consolidate/compare separately).

Holding off on the rest of the rebase until I hear back, given the
CI-gating rule in the charter. Happy to move fast once I know which way
to go — @ckeda tagging you too since it's your plugin on the other side
of this collision.

JamesCarnley:
Thank you @TanishDevs11. I read the final `5cc1c885` revision and both review comments, including your July follow-up about the add/add collision with #138. That was a substantive integration question awaiting guidance, not simply an unfinished rebase.

The secret-free delegation test using a second instance with a different issuer secret, the unaffected-sibling control, and the twelve-leaf revocation scenario are valuable. I am retaining those requirements and this source as references for a future optional delegation profile. Current Town uses flat, controller-signed RunGrants, so importing this implementation would introduce a new authority protocol rather than repair an existing delegation API.

That profile needs explicit bearer-versus-holder-bound semantics, verifier-side scope/expiry containment for independently constructed chains, bounded parsing, and revocation persistence/distribution. The submitted verifier checks chain signatures, ancestor expiry and revocation, but does not reconstruct each child's scope/expiry containment. The reusable validators also count general exceptions as successful rejection without their own healthy control. Those boundaries need to be resolved before this can be evidence for a real external implementation; an unsupported delegation API must not be scored as an insecure supported one.

Closing this archive-targeted PR without a replacement merge as part of the current-main migration. No cascading-delegation support is claimed delivered, and this is not being dismissed as a duplicate of #138. The independent-secret test, strict-subset policy choice, typed rejection/error distinction, complete branch coverage and sibling controls remain recorded for the profile design. The legacy code/tests were inspected, not executed.

PR #135, opened 2026-07-09
Improve marketplace PR short descriptions

itsbilyatt:
## Summary
- Skip Markdown bullet lists when deriving short descriptions for hackathon submissions.
- Add a regression test covering the case where a PR body starts with checklist-style bullets.

## Why
This makes the marketplace cards surface the first real prose explanation instead of the opening checklist item.

## Testing
- PYTHONPATH=packages/nest-marketplace pytest -q packages/nest-marketplace/tests/test_adapter.py

### DIFF
--- a/packages/nest-marketplace/nest_marketplace/adapter.py
+++ b/packages/nest-marketplace/nest_marketplace/adapter.py
@@ -210,8 +210,9 @@ def classify_layer(theme: str | None, title: str = "", body: str = "") -> str:
 def short_description(body: str, max_len: int = 240) -> str:
     """Pull a short blurb out of a PR body.
 
-    Skips markdown heading lines and pull-quotes, then truncates at the
-    first sentence boundary or `max_len`.
+    We skip markdown scaffolding (headings, pull-quotes, code fences, and
+    leading bullet lists) and return the first substantive prose line.
+    Truncation still happens at the first sentence boundary or `max_len`.
     """
 
     if not body:
@@ -225,6 +226,10 @@ def short_description(body: str, max_len: int = 240) -> str:
         if line.startswith(">"):
             continue
         if line.startswith("```"):
+            continue
+        # PR templates often begin with bullet lists or checklist items.
+        # Skip those so the marketplace card uses the actual explanation.
+        if re.match(r"^(?:[-*+]|\d+\.)\s+(?:\[[ xX]\]\s+)?", line):
             continue
         # Drop bold markers and inline code backticks for the blurb.
         cleaned = line.replace("**", "").replace("`", "")
--- a/packages/nest-marketplace/nest_marketplace/adapter.py
+++ b/packages/nest-marketplace/nest_marketplace/adapter.py
@@ -217,6 +217,8 @@ def short_description(body: str, max_len: int = 240) -> str:
 
     if not body:
         return ""
+
+    first_bullet: str | None = None
     for raw_line in body.splitlines():
         line = raw_line.strip()
         if not line:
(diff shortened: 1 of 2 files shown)


dhve:
Thanks @itsbilyatt. This is a real quality-of-life fix for the marketplace: cards that used to lead with a raw checklist item now surface actual prose, the regex in `short_description` handles `-`/`*`/`+`, numbered lists, and `- [x]` checkboxes, and you included a regression test plus a runnable pytest line in the PR body.

**1. Bullets-only bodies now render as "No description provided."**

The docstring says "leading bullet lists," but the skip applies to every bullet line. A body that is `## Summary` plus bullets with no prose at all now returns `""` from `short_description`, and `hackathon-card.tsx` falls back to "No description provided." At least one open submission body is exactly that shape today, so this is a small correctness regression: before, those cards at least showed the first bullet's text. Fix: bind the match (`if m := re.match(...)`), remember the first skipped bullet's text (`line[m.end():]`), and return it, truncated, if the loop finds no prose. That also improves on main, which showed the literal `- ` marker.

**2. Add the tests that would have caught 1.**

`test_short_description_skips_leading_markdown_lists` only covers the happy path (bullets then prose). Add a bullets-only body asserting the fallback text, and a `- [x] checklist` case since your regex deliberately supports it. Cheap tests, and they pin the fallback behavior.

One meta question: is this intended as your hackathon submission, or a tooling fix? It touches shared marketplace infra rather than one of the ten problems in `docs/hackathon/problems/`, so the judge panel will not score it as a submission. If you plan to compete, pick a problem and make that your single hackathon PR (charter branch format is `hackathon/<your-handle>-<short-theme>`). Also note CI has not reported on this branch yet.

**Suggested path forward:**

1. Add the first-bullet fallback with the marker stripped.
2. Add the bullets-only and checklist tests.
3. Tell me whether this is a submission or an infra contribution so we route it right.

itsbilyatt:
Addressed the two requested follow-ups from review.

### What changed
1. **Bullets-only fallback in `short_description`**
   - Updated `packages/nest-marketplace/nest_marketplace/adapter.py` so list/checklist lines are still skipped when searching for prose, **but** the adapter now remembers the first skipped bullet text.
   - If no prose line is found after scanning the body, it returns that first bullet text (marker stripped) instead of returning empty.
   - This prevents cards from regressing to **"No description provided."** for bodies that only contain bullets.

2. **Checklist-aware behavior preserved**
   - The existing list/checklist regex path remains intact (`-`, `*`, `+`, numbered lists, and `- [x]` style checklist items).
   - Fallback strips markdown markers so output stays clean.

### Tests added
In `packages/nest-marketplace/tests/test_adapter.py`:
- `test_short_description_returns_first_bullet_when_only_bullets`
- `test_short_description_handles_checklist_only`

These complement the existing happy-path test (`...skips_leading_markdown_lists`) and pin the regression case that was called out.

### Validation run
Executed locally:
- `PYTHONPATH=packages/nest-marketplace pytest -q packages/nest-marketplace`

Result: all tests passed.

### Commit
- `34b6f38` on branch `hackathon/prajyot-marketplace-short-description`

Also confirming intent: this PR is an **infra/tooling quality-of-life fix** for marketplace rendering, not a scored hackathon problem submission.

JamesCarnley:
Thanks @itsbilyatt for the marketplace quality-of-life fix, including the follow-up that preserves useful text for bullet-only and checklist-only submissions. I reviewed the full patch and discussion at `34b6f3884bcd882b922c1c3f899b8ff2271659a3`; those requested fallback tests are present.

Against rebuilt `main@f4e85d7`, the `nest_marketplace.adapter.short_description` module and its card-rendering path are retired. Current `src/nandatown/protocols.py` catalogs the explicit PR title and snapshot metadata; it does not extract description prose from PR bodies. This is therefore not a remaining bug in that implementation.

Closing as **obsolete marketplace-parser maintenance, without a replacement**, not as merged code or a rejection of the completed fix. If a PR-body summary surface is introduced again, preserve your tested behavior: prefer substantive prose, otherwise return the first useful bullet with its marker stripped, with checklist and truncation coverage. No second parser or marketplace is being added to the local harness solely to transplant this patch. The contribution and review history remain here. Current layer, import and adaptation checks passed (33 tests); the legacy code was inspected, not executed.

PR #136, opened 2026-07-09
[Hackathon] settlement-engineer: split_settlement — weighted multi-payee fan-out settlement with conservation-proof validators

KaranSinghBisht:
## Motivation

The merged payments plugins each move value in one shape:

- `prepaid_credits` — one debit to **one** payee.
- `escrow` — one payee's payout, gated on delivery/arbitration.
- `empic_escrow` — pull/pubsub escrow for a provider/consumer pair.
- `streaming` — **one** payee metered one tick at a time.

None of them takes a single debit and fans it out to **N payees at once** by
declared shares.

**Scope anchor:** this PR is a self-scoped payments-layer problem, distinct
from problem 03 by that problem's own scope declaration — 03 (streaming
x402) explicitly lists multi-party streams (one payer, many payees) as out
of its scope — following the merged precedent for self-scoped payments work
(#38 arbitrated escrow, #90 EMPIC). Novelty is therefore claimed against
the fan-out settlement shape, which exists nowhere in the merged set. That is the missing primitive behind the most common real
settlement flows: marketplace revenue splits, royalty payouts, ad/network
rev-share, and contributor splits. Doing this naively is exactly where money
leaks — floor every share and the indivisible remainder ("dust") silently
vanishes, or a compromised splitter quietly reweights the payout in its own
favour. This PR adds `split_settlement`: an atomic, exact, weight-locked fan-out
settlement, plus two adversarial validators that make those two failure modes
un-shippable.

## Design

**Money & weights are integers, end to end.** Credits are `int` (matching the
`Money.amount: int` the rest of the layer uses); weights are **positive
integers** (ratios). An 80/20 split is `(80, 20)` or `(4, 1)` — never floats.
Nothing on the settlement path touches floating point, so no allocation can drift
by a rounding error.

**Largest-remainder (Hamilton) allocation.** For amount `A` and weights `wᵢ`
with total `W`:

1. Each payee first gets the floor of its exact quota: `fᵢ = A·wᵢ // W`.
2. The leftover `dust = A − Σfᵢ` is handed out one unit at a time to the payees
   with the largest remainders `rᵢ = A·wᵢ mod W`, ties broken by payee id, then
   original index.

*Conservation proof sketch.* `Σfᵢ = Σ(A·wᵢ − rᵢ)/W = A − (Σrᵢ)/W`. Each
`rᵢ ∈ [0, W)`, so `Σrᵢ < n·W` and `dust = (Σrᵢ)/W ∈ {0, …, n−1}` is a
non-negative integer strictly less than the payee count. Distributing exactly
`dust` unit increments restores the total to `A`, so `Σ(allocation) == A`
**exactly**, for every amount and every weight vector. The tie-break is a total
order on `(−remainder, id, index)`, so the result is deterministic regardless of
dict iteration or input ordering.

**Weights are locked at open.** `open_split(splits, ref)` records the weight
vector as an immutable tuple and moves no money; `settle_split(ref, amount)` is
the only thing that pays out, and it recomputes the allocation purely from the
locked weights. The weights a settlement honours cannot be renegotiated between
open and settle.

**Atomicity.** `settle_split` computes the full allocation, re-checks
conservation, and verifies the payer's balance *before* crediting anyone; then it
debits the payer once and credits every payee in a single pass. If the balance is
short, nothing moves. The two-phase `open`→`settle` form leaves a failed
settlement `OPEN` and retryable; the one-shot `split_pay` convenience is
all-or-nothing and rolls its own contract back on failure.

**Role binding & receipts.** On a shared ledger only the contract's payer may
settle or refund it. Every settlement returns one `Receipt` per payee and stores
a queryable settlement record (`settled_amount` + per-payee `allocations`).

**Registration.** Both `_BUILTINS` (`nest_core/plugins.py`) **and** a
`nest.plugins.payments` entry point in `packages/nest-plugins-reference/pyproject.toml`
(payments previously had no entry-point group at all).

## Tradeoffs

1. **Weights at open vs. weights at settle.** Locking weights at open (and
   letting the amount arrive at settle) makes the allocation a pure function of
   `(declared weights, amount)` that a validator can recompute independently.
   Locking the *amount* too would forbid "same split, variable revenue per
   round"; locking neither would make weight-tampering undetectable. Weights-at-
   open is the choice that buys the audit.

2. **Duplicate `ref` → typed rejection, not idempotent replay.** A settlement is
   a unique financial commitment, so a reused reference raises `SplitError`
   rather than silently returning the prior contract — surfacing a double-spend
   bug in the caller instead of masking it. (Documented in `open_split`.)

3. **Single settle per contract vs. multi-settle.** A contract settles once. A
   drip/multi-settle model would need running-total accounting and a second
   conservation invariant; keeping it single-shot makes the proof one line and
   the trace unambiguous. Streaming already owns the metered-payout shape.

4. **"Unknown payee" is opt-in, not invented.** The payments layer has no agent
   directory — identity is the registry/identity layer's job — so `open_split`
   takes an optional `known_payees` roster and rejects outsiders only when the
   caller supplies one, instead of pretending to validate identities it cannot
   see. It *does* enforce the settlement-local hazards unconditionally: empty
   splits, non-positive weights, duplicate payees, and the payer listing itself
   (self-dealing).

## Adversarial validators

The hidden property this pair pins down: **conservation is not weight-fidelity.**
A compromised splitter can conserve the total perfectly while stealing the
split — route everything to itself, drop a payee, quietly reweight — and a
conservation check alone will wave it through. No merged payments plugin
expresses fan-out at all, so neither failure mode is currently catchable
anywhere in the stack. Closing that gap requires an auditor that *recomputes
the allocation independently from the declared weights*, which is exactly what
the second validator does.

Registered as `VALIDATORS["split_settlement"]`; both read the JSONL trace's
`split:opened` / `split:settled` broadcasts. The default `prepaid_credits` plugin
has no fan-out protocol, so buyers fall back to `pay()` and emit no `split:*`
events — both validators report "no split lifecycle observed" and fail, which is
the charter's bar for adversarial.

1. **Penny-shaving → `split_conservation`.** Sums the per-payee credits in each
   settlement and compares to the debited amount. A splitter that floors every
   share and keeps the dust credits strictly less than it debits; the check fails
   on the shortfall (and on any negative credit).

2. **Weight-tampering → `split_weight_fidelity`.** **Independently recomputes**
   the canonical largest-remainder allocation from the *declared* weights and
   compares it payee-by-payee to the reported credits. It never imports the
   plugin, so a compromised plugin cannot subvert its own audit. This catches a
   mid-flight reweight, a dropped/injected payee, or dust self-dealing **even
   when the total still sums correctly** — the case conservation alone cannot
   see.

The tests prove the two are non-redundant: an in-test `_WeightTamperingSplitter`
that routes the whole amount to the first payee **passes** `split_conservation`
(it sums to the amount) but **fails** `split_weight_fidelity`; a
`_PennyShavingSplitter` fails conservation on the dust. Both buggy splitters are
driven through the real `Simulator` (they are test-only, never shipped as
plugins).

## Scope / core wiring

Every `nest_core` touch is required registration, following the existing house
pattern for each file:

- `nest_core/plugins.py` — one `_BUILTINS` entry (`("payments", "split_settlement")`),
  the same mapping every bundled plugin uses.
- `nest_core/scenarios.py` — one registration branch for the builtin scenario
  (the file already carries twenty such branches).
- `nest_core/validators.py` — the two validators live here because this is
  where all seventy existing validators live, alongside the registry they
  join.
- `nest_core/tests/test_validators.py` — one line: the registry keyset test
  enumerates every validator name and fails CI without it.
- `nest_core/scenarios_builtin/split_settlement.py` + its YAML — the builtin
  scenario pair, precedented by `auction` (which also ships
  `scenarios/auction.yaml` alongside the builtin copy).

No file outside the plugin/validator/scenario/docs surface is modified, and
`scripts/judge/`, the seed bank, and `scores.json` are untouched.

## Verification

```bash
uv sync
uv run ruff check .
uv run ruff format --check .
uv run pyright
uv run pytest -v
# scenario, by name, across the seed bank:
uv run nest run split_settlement --seed 42
uv run nest run split_settlement --seed 7
uv run nest run split_settlement --seed 1337
```

All five gates exit 0 (`0 errors, 0 warnings` from pyright; full suite green).
The 53 new tests span the pure allocator, the full lifecycle, every typed
rejection, `hypothesis` property batteries (conservation, the exact
largest-remainder + tie-break ordering, permutation invariance, whole-ledger
conservation and atomicity under random op sequences), the end-to-end scenario,
and the two validators against both the honest plugin and the buggy splitters.

Validating a scenario trace:

```
PASS split_conservation: 9 settlements conserved value exactly (sum of credits == amount)
PASS split_weight_fidelity: 9 settlements matched the largest-remainder split of their declared weights
```

Under `payments: prepaid_credits` the same two validators fail with
"no split settlements observed in trace -- plugin lacks fan-out settlement".

## Layer & companion Phase 2 submission

- **Layer:** `payments` (plugin `payments: split_settlement`), with trace
  validators registered under `split_settlement` and a builtin scenario.
- **Author (both phases):** Karan Singh Bisht — GitHub `KaranSinghBisht`,
  same identity on the participation form and the skills registry.
- **Phase 2:** AgentPress — The Autonomous Newsroom of the Agent Economy
  ([registry entry](https://nandatown.projectnanda.org/api/skills/982f1085-3469-472c-a3d5-5736585e11f1),
  [SKILL.md](https://agentpress-nanda.vercel.app/skill.md),
  [live service](https://agentpress-nanda.vercel.app)). AgentPress settles
  its contributor revenue with exactly the primitive this PR contributes:
  an 80/20 fan-out split, weighted by editorial score, allocated by
  largest remainder so the ledger conserves every credit. This plugin is
  that economy expressed as NANDA Town infrastructure.

## Determinism

The settlement path uses no wall-clock and no RNG; allocation ties break on
`(remainder, payee id, index)`. The scenario consumes no simulator randomness, so
each seed's trace is byte-identical across runs — verified for seeds 42, 7, and
1337 (run twice, `diff` clean) both via the property tests and the `nest run`
CLI.

### DIFF
--- a/docs/layers/payments.md
+++ b/docs/layers/payments.md
@@ -48,11 +48,15 @@ callers keep working. Where `escrow` conditions one payee's payout and
 once — the missing shape for marketplace revenue splits, royalty payouts, and
 contributor rev-share. The bundled `split_settlement` scenario (12 agents: a
 content marketplace splitting buyer payments 80/20 across contributors and a
-platform) runs under two adversarial validators the default `prepaid_credits`
+platform) runs under three adversarial validators the default `prepaid_credits`
 plugin cannot satisfy: `split_conservation` catches **penny-shaving** (a splitter
-that floors every share and pockets the dust) and `split_weight_fidelity`
+that floors every share and pockets the dust), `split_weight_fidelity`
 recomputes the canonical allocation independently to catch **weight tampering**
-(a mid-flight reweight or self-dealing that still sums to the amount). Source:
+(a mid-flight reweight or self-dealing that still sums to the amount), and
+`split_ledger_attestation` catches **ledger skimming** (canonical-looking
+receipts that credit the shared ledger short) by cross-checking the payer's
+receipts against each payee's own `split:observed` balance attestation. The first
+two audit reported receipts; the third audits payee-observed ledger truth. Source:
 [`nest_plugins_reference/payments/split_settlement.py`](../../packages/nest-plugins-reference/nest_plugins_reference/payments/split_settlement.py).
 
 ## Writing your own
--- a/docs/layers/payments.md
+++ b/docs/layers/payments.md
@@ -36,6 +36,25 @@ provider / service binding by payment reference and reject traces that leak
 private keys, API keys, bearer tokens, wallet secrets, or other live rail
 secrets.
 
+`split_settlement` — weighted multi-payee **fan-out settlement**. One atomic
+debit from a payer is split across *N* payees by integer weights declared and
+locked when the contract is opened. Allocation uses the largest-remainder
+(Hamilton) method over integer credits, so every settlement conserves value
+exactly (`sum(allocations) == amount`) and is deterministic — the indivisible
+dust goes to the payees with the largest fractional parts, ties broken by payee
+id. A plain `pay()` is the degenerate single-payee, weight-1 split, so one-shot
+callers keep working. Where `escrow` conditions one payee's payout and
+`streaming` meters one payee per tick, this splits one debit across many at
+once — the missing shape for marketplace revenue splits, royalty payouts, and
+contributor rev-share. The bundled `split_settlement` scenario (12 agents: a
+content marketplace splitting buyer payments 80/20 across contributors and a
+platform) runs under two adversarial validators the default `prepaid_credits`
+plugin cannot satisfy: `split_conservation` catches **penny-shaving** (a splitter
+that floors every share and pockets the dust) and `split_weight_fidelity`
+recomputes the canonical allocation independently to catch **weight tampering**
+(a mid-flight reweight or self-dealing that still sums to the amount). Source:
+[`nest_plugins_reference/payments/split_settlement.py`](../../packages/nest-plugins-reference/nest_plugins_reference/payments/split_settlement.py).
+
 ## Writing your own
 
 See [`writing-a-plugin.md`](../writing-a-plugin.md) — the full
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -42,6 +42,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("payments", "streaming"): f"{_REF}.payments.streaming:StreamingPayments",
     ("payments", "empic_escrow"): f"{_REF}.payments.empic_escrow:EMPICEscrowPayments",
     ("payments", "escrow"): f"{_REF}.payments.escrow:EscrowPayments",
+    ("payments", "split_settlement"): f"{_REF}.payments.split_settlement:SplitSettlement",
     ("coordination", "contract_net"): f"{_REF}.coordination.contract_net:ContractNet",
     ("coordination", "hotstuff"): f"{_REF}.coordination.hotstuff:HotStuff",
     ("negotiation", "alternating_offers"): (
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -197,3 +197,9 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("capability_spoofing", capability_spoofing_factory)
+    elif name == "split_settlement":
+        from nest_core.scenarios_builtin.split_settlement import (
+            split_settlement_factory,
+        )
+
+        register_scenario("split_settlement", split_settlement_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/split_settlement.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/split_settlement.py
@@ -1,0 +1,199 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Split-settlement scenario -- a content marketplace with revenue sharing.
+
+Twelve agents: three buyers, eight contributors, and one platform.  Each buyer
+runs several rounds; every round pays for a content bundle and fans the payment
+out 80/20 -- 80% shared between two contributors, 20% to the platform -- through
+the ``split_settlement`` plugin.  For every settlement the buyer broadcasts two
+structured events the validators read:
+
+* ``split:opened:ref=..:payer=..:weights=payee~weight;...`` -- the *declared*
+  weights, locked at open time;
+* ``split:settled:ref=..:amount=..:alloc=payee~credit;...`` -- the *actual*
+  per-payee credits the plugin produced.
+
+The two ``split_settlement`` validators cross-check those broadcasts: credits
+must sum to the amount (no penny-shaving) and must equal the largest-remainder
+allocation of the declared weights (no weight tampering).  The round amounts are
+deliberately indivisible by the weight total (``777``, ``1001``) so the dust
+distribution -- the exact place a naive splitter leaks value -- is exercised in
+the trace itself.
+
+If the configured payments plugin lacks ``open_split`` (e.g. the default
+``prepaid_credits``), each buyer falls back to a single ``pay()`` and emits no
+``split:*`` events, which the validators report as "no split lifecycle observed"
+-- the adversarial discrimination the charter requires.
+
+Example::
+
+    agents = split_settlement_factory(config, plugins)
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+from nest_core.scenario import ScenarioConfig
+from nest_core.sim.agent import AgentContext, StateMachineAgent
+from nest_core.types import AgentId, Money, PaymentRef
+
+_PLATFORM = AgentId("platform-0")
+_CONTRIB_BPS = 40  # each of the two contributors: 40 + 40 = 80% to contributors
+_PLATFORM_BPS = 20  # 20% to the platform
+_BUYER_BALANCE = 5000
+
+
+@dataclass
+class _Job:
+    """One round's settlement: the ref, the locked weights, and the amount."""
+
+    buyer: AgentId
+    ref: PaymentRef
+    payees: tuple[tuple[AgentId, int], ...]
+    amount: int
+
+
+def _split(c1: str, c2: str) -> tuple[tuple[AgentId, int], ...]:
+    """Build an 80/20 weight vector: two contributors plus the platform."""
+    return (
+        (AgentId(c1), _CONTRIB_BPS),
+        (AgentId(c2), _CONTRIB_BPS),
+        (_PLATFORM, _PLATFORM_BPS),
+    )
+
+
+# buyer -> list of (contributor_a, contributor_b, amount) per round. Amounts mix
+# an evenly divisible case (1000) with two indivisible cases (777 -> two dust
+# units split across the contributors, 1001 -> one dust unit that a stable
+# tie-break must award to the lexicographically smaller contributor).
+_PLAN: dict[str, list[tuple[str, str, int]]] = {
+    "buyer-0": [
+        ("contrib-0", "contrib-1", 1000),
+        ("contrib-2", "contrib-3", 777),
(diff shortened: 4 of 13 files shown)


dhve:
Thanks @KaranSinghBisht. This is one of the stronger payments submissions in the queue. `allocate_by_weight` is a clean, pure-integer largest-remainder allocator with a correct conservation argument and a total tie-break order, and the validator pair is genuinely non-redundant: your `_WeightTamperingSplitter` passing `split_conservation` while failing `split_weight_fidelity` is exactly the "conservation is not weight-fidelity" proof the PR body promises. The hypothesis battery (floor-or-floor-plus-one, dust count equals rounded-up count, permutation invariance, whole-ledger atomicity under random op sequences) is the kind of test rigor we keep asking for. I also cross-checked your infrastructure claims against main and they all hold: the `_agent_plugins` override channel, the `prepaid_credits` kwarg fallback in `_buyer_instance`, the missing `nest.plugins.payments` entry-point group, and the `auction` precedent for shipping the YAML in both `scenarios/` and `scenarios_builtin/yaml/`.

This is close. Four things to work through:

**1. CI has not run on this branch at all.**

No checks are reported on `hackathon/karansinghbisht-split-settlement`, which is normal for a first PR here (workflow runs need maintainer approval, and that is on us, not you). Flagging it anyway because the judge panel does not score PRs without green CI. Once the run is approved, keep an eye on it and make sure all five gates match your local results.

**2. Say which problem this anchors to in the PR body.**

The charter asks for exactly one problem from `docs/hackathon/problems/`, and the payments problem there (03, streaming x402) explicitly lists multi-party streams (one payer, many payees) as out of scope for itself. So this is a self-scoped payments-layer problem. There is merged precedent for that (#38 arbitrated escrow, #90 EMPIC), so it is not disqualifying, but the body never states it. Add one line saying it is self-scoped and distinct from problem 03 by that problem's own scope declaration, so the judges score novelty against the right baseline. Cheap fix, docs dimension.

**3. Both validators audit plugin-reported receipts, not ledger truth.**

`validate_split_weight_fidelity` never imports the plugin, which is good, but its input is the `alloc=` field of the `split:settled` broadcast, and `BuyerAgent.on_message` builds that from whatever `settle_split` returned. Your `_WeightTamperingSplitter` is an honest liar: it tampers and then reports the tampering in its receipts. A nastier splitter that credits itself on the shared ledger while returning canonical receipts passes both validators. Within trace-based validation you cannot fully escape this, but you can tighten it: contributors are separate agents, so give each a read handle on the shared ledger and have them broadcast an attestation (say `split:observed:payee=..:balance=..`) each round, then cross-check attested balance deltas against the buyer-reported allocations. If you decide that is out of scope, at minimum document the trust boundary in both validator docstrings so the next reader does not over-claim what the audit covers. This is the one change I would most like to see; it is squarely the test-rigor axis.

**4. The fidelity check is order-sensitive where the event format does not pin order.**

In `validate_split_weight_fidelity`, `reported != expected` compares position by position. Nothing in the emitted `split:settled` format declares allocation order canonical, so an honest plugin variant that returns receipts sorted by payee id would be flagged as tampering. Either compare `dict(reported) == dict(expected)` with a length guard for duplicates, or document that the canonical form requires allocation order to mirror declared weight order, and add a synthetic-trace test for the reordered-but-correct case so the decision is pinned either way.

One nit while you are in there: `_CONTRIB_BPS` / `_PLATFORM_BPS` in the scenario are named basis points but hold percent-style weights (40/40/20). `_CONTRIB_WEIGHT` would read truer.

**Suggested path forward:**

1. Add the problem-anchoring line to the PR body (one sentence).
2. Make the fidelity comparison order-insensitive or document ordering as canonical, with a test either way.
3. Add the contributor-side attestation cross-check, or document the receipts trust boundary in both validator docstrings.
4. When the workflow run is approved, confirm all five CI gates are green.

Leaving this open. Ping me on the hackathon channel if the attestation idea in point 3 is unclear.

KaranSinghBisht:
Thanks for the close read @dhve — the cross-checks against main and the attestation suggestion in particular. All four points addressed; pushed as three commits.

**1. CI** — understood it's on your side; I'll watch the run when it's approved and confirm all five gates match the local results (`make ci-local` is green on this head: ruff check/format clean, pyright 0 errors/0 warnings, pytest 1064 passed / 1 skipped).

**2. Problem anchoring** — added a scope-anchor paragraph to the PR body: self-scoped payments-layer problem, distinct from problem 03 by 03's own out-of-scope declaration for multi-party streams, following the #38/#90 precedent.

**3. Ledger attestations — implemented (your preferred option), not just documented.** Contributors and the platform agent now hold a read handle on the shared ledger; after each `split:settled` they observe, each payee broadcasts `split:observed:ref=..:payee=..:balance=..` on deterministic staggered ticks. A third validator, `validate_split_ledger_attestation`, tracks per-payee attested balances across settlements and requires the delta between consecutive attestations to equal the buyer-reported allocation for the intervening settlement — failing on mismatch, and failing when settlements have no covering attestations (same fail-when-absent posture as the existing pair vs the reference plugin). The non-redundancy proof you asked about is in the tests: a new `_LedgerSkimmingSplitter` credits itself on the shared ledger while returning canonical receipts — it PASSES `split_conservation` and `split_weight_fidelity` and FAILS only the attestation check, driven through the real Simulator like the other buggy splitters. The receipts-vs-ledger trust boundary is now documented in all three validator docstrings, so the audit's coverage can't be over-read.

**4. Order sensitivity** — fidelity now compares `dict(reported) == dict(expected)` with an explicit duplicate-payee guard (duplicate payees in a reported allocation fail loudly), and a synthetic-trace test pins the reordered-but-correct case as passing.

**Nit** — `_CONTRIB_BPS`/`_PLATFORM_BPS` renamed to `_CONTRIB_WEIGHT`/`_PLATFORM_WEIGHT`.

Determinism holds: the attestation ticks are scheduled, not clock-driven — the scenario suite still asserts byte-identical traces across seeds 42/7/1337.

JamesCarnley:
Thank you @KaranSinghBisht. I reviewed all 13 changed files and both discussion comments at `7ea7f71dca931391f8679a0f0f6e086d968184d9`. The added payee balance observers, ledger-skimming discriminator, order-independent allocation comparison and duplicate-payee guard address the earlier review requests and are valuable.

A selected requirement is now adapted to current Town in [#245](https://github.com/projnanda/nandatown/pull/245), merged as `6a697f2dc1198a1754b525d32ee1913da7ad8e9a`: conserving the total does not establish that the intended recipient was paid. Before the correction, a real local auction could pay a losing bidder instead of the auctioneer and still pass. The new evaluator binds its payment to the configured auctioneer/item, task announcement, recorded award, winner, integer amount, run and causal order. Known contradictions fail; missing evidence remains incomplete. The healthy run still passes and replays.

Verification: 342-test baseline; 30 expected initial regression failures; 60 focused checks; 385 final tests with 8 existing warnings; Python 3.11/3.12 CI green. Six omitted-check mutations were caught by regressions. A genuine old bundle explicitly reports the `lab-0.2.4`/`lab-0.2.5` evaluator mismatch instead of claiming equivalent replay.

Closing this archived-base PR as **partially superseded**, not merged wholesale. Explicitly retained future scope:

- Weighted fan-out with exact integer dust/tie rules, permutation and duplicate-payee cases, weight tampering, refunds and the independent property-test corpus.
- Distinguish reported allocation from independently observed balances. The existing auction fix checks correlated Lab records; it does not implement external account read-back.
- For a future split profile, bind observer identity to the claimed payee/reference/time, establish initial balances and account for concurrent unrelated movements, missing/conflicting observations and declared-weight integrity. The submitted observer experiment assumes serialized payments to zero-start passive recipients; it is not a general external-settlement proof.

No weighted payment runtime or provider was adopted, and no legacy contributor code or live payment service was executed. Your original branch, tests and discussion remain preserved as the reference for that separate profile.

PR #138, opened 2026-07-09
[Hackathon] ckeda-security: delegatable capability tokens with cascading revocation

ckeda:
**Branch**: `hackathon/ckeda-delegatable-auth` · **Problem 04**

## What this adds

**Plugin** — `("auth", "delegatable")`, macaroon-style HMAC-chained tokens
(`nest_plugins_reference/auth/delegatable.py`). Any token holder mints
attenuated children *offline* via `delegate(parent_token, audience,
scopes_subset, ttl)`; each child's signature is keyed by its parent's
signature, so revoking any segment's tid invalidates every descendant by
construction — no per-child revocation lists. `verify` walks the full
chain: signature, per-segment revocation, per-segment expiry, monotonic
scope attenuation, monotonic expiry. `verify_presented(token, presenter)`
additionally binds presentation to the token's audience. Typed exceptions
(`ScopeEscalationError`, `RevokedAncestorError`, `ExpiredAncestorError`,
`AudienceMismatchError`) all subclass `ValueError` for drop-in
compatibility with callers written against the `jwt` plugin. Satisfies
the base `Auth` protocol (`isinstance` check included in tests).

**Adversarial validator** —
`nest_plugins_reference/validators/delegation_validators.py`, three pure
functions over `delegation_audit` trace events, catching the brief's
three attacks: scope escalation on granted delegations, verification
under a revoked ancestor, and presentation by a non-audience agent. As
required: all three FAIL against `auth: jwt` and PASS against
`auth: delegatable` — demonstrated both with synthetic audit streams and
an end-to-end in-process run of the real scenario agents under both
plugins (`test_delegation_validators.py`).

**Scenario** — `delegated_auth` (`scenarios/delegated_auth.yaml` +
`scenarios_builtin/delegated_auth.py`): 1 coordinator, 3 intermediaries,
12 leaves in a delegation tree, per the brief. Adversarial behaviours are
baked in deterministically: intermediary-2 attempts a scope-escalated
grant, the coordinator revokes intermediary-1's grant at tick 20 while
that subtree keeps presenting, and leaf-0 shares its token with leaf-1
which presents it. Under `auth: jwt` the scenario's delegation fallback
is deliberately the vulnerable pattern (re-issuance with no chain), so
the same YAML flips all three validators.

## Design decisions

- **Strict tree, not DAG** — single parent per segment, per the brief's
  pointer. Sufficient for cascading revocation and much simpler to audit.
- **Child TTL clamped, not rejected** — `delegate` clamps child expiry to
  the parent's rather than raising, so "give me up to an hour" composes;
  monotonicity is still re-checked at verify.
- **Deterministic everywhere** — injectable logical clock
  (`clock=` / `set_clock`), content-derived segment tids (sha256 of the
  segment preimage, no uuid/wall-clock), sorted-key canonical JSON for
  signing.
- **HMAC chaining, not claims** — per the anti-pattern list, there is no
  `delegated_from` claim; the parent's *signature* is the child's key, so
  transitive revocation is cryptographic, not bookkeeping.

## Tests

23 new tests: 14 unit (three attacks, deep-chain cascade, sibling-branch
isolation after mid-tree revocation, tamper/garbage rejection, protocol
conformance, and a documented jwt-vulnerability baseline), 3 Hypothesis
property tests (leaf scopes never exceed root and leaf never outlives
root under random chains; revoking a random ancestor always invalidates
the leaf; token bytes deterministic for fixed clock+secret), 6 validator
tests including the end-to-end both-plugins run.

## Definition of Done

All five CI

### DIFF
--- a/docs/layers/auth.md
+++ b/docs/layers/auth.md
@@ -21,6 +21,27 @@ shape (header.payload.sig), no claim validation beyond the signature.
 
 Source: [`nest_plugins_reference/auth/jwt_auth.py`](../../packages/nest-plugins-reference/nest_plugins_reference/auth/jwt_auth.py).
 
+## Hardened plugin: `delegatable`
+
+Macaroon-style HMAC-chained capability tokens. Any token holder mints
+attenuated child tokens **offline** via
+`delegate(parent_token, audience, scopes_subset, ttl)`; each child's
+signature is keyed by its parent's signature, so revoking any segment
+invalidates every descendant at the next `verify` — cascading revocation
+by construction, no per-child revocation lists. `verify` re-checks the
+full chain (signature, per-segment revocation and expiry, monotonic scope
+and expiry attenuation); `verify_presented(token, presenter)` additionally
+binds presentation to the token's audience.
+
+Adversarial validators (`check_no_scope_escalation`,
+`check_no_stale_ancestor_use`, `check_audience_binding`) fail against the
+`jwt` plugin and pass against `delegatable`; the `delegated_auth` scenario
+exercises all three attacks deterministically.
+
+Source: [`nest_plugins_reference/auth/delegatable.py`](../../packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable.py).
+Validators: [`nest_plugins_reference/validators/delegation_validators.py`](../../packages/nest-plugins-reference/nest_plugins_reference/validators/delegation_validators.py).
+Scenario: [`scenarios/delegated_auth.yaml`](../../scenarios/delegated_auth.yaml).
+
 ## Writing your own
 
 See [`writing-a-plugin.md`](../writing-a-plugin.md). Register under
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -26,6 +26,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("registry", "in_memory"): f"{_REF}.registry.in_memory:InMemoryRegistry",
     ("registry", "gossip"): f"{_REF}.registry.gossip:GossipRegistry",
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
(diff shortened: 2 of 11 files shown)


ckeda:
Phase 2 companion is now live in the registry: **Claw-in-a-Box** on the
NANDA Town skills page — hosted endpoint https://api.clawinabox.xyz
(registry probe: link responded ✓), SKILL.md served at
https://api.clawinabox.xyz/skill.md.

Full development history for this implementation, including the
Node.js service mirror of the same protocol, is public at
https://github.com/ckeda/claw-in-a-box.

shilpa-kulkarni-14:
@ckeda - LGTM

Non-blocking items before merge

Register the validators. delegation_validators is importable and unit-tested, and I confirmed it works against real traces, but it is not exported in validators/__init__.py the way #115 registered its privacy validators. Add the one-line registration so the finale judging harness auto-discovers the three checks. This does not block merge, since the discriminator is verified working, but it matters for automated scoring.
Add a docs/layers/auth.md entry, as #105 and #115 did for their layers.
Seed-invariance is fine but worth a one-line note, or confirm it is intended.

ckeda:
@shilpa-kulkarni-14 Thank you for the review! All three items addressed
in 99d3b99:

- **Validators registered** — the three delegation checks (plus
  `extract_delegation_audits`) are now exported from
  `validators/__init__.py`, following the pattern #115 established for
  the privacy validators.
- **Docs entry added** — `docs/layers/auth.md` now documents the
  `delegatable` plugin alongside the default `jwt`, with links to the
  plugin, validators, and scenario.
- **Seed-invariance confirmed intended** — noted in the scenario module
  docstring: agents draw nothing from `ctx.rng`; the adversarial
  behaviours are structural (fixed roles, fixed ticks), so the trace is
  byte-identical across seeds by design.

Branch is freshly merged with `main` (clean). Local CI five-pack green:
1003 passed, pyright strict 0 errors.

shilpa-kulkarni-14 (approved):

PR #141, opened 2026-07-09
[Hackathon] auth-safety-engineer: DelegatableAuth with scoped delegation, cascading revocation, and deterministic validators

Niruuu2005:
## Layer picked: **Auth (#5)** — Problem 04

**Persona:** `auth-safety-engineer`

## Summary

Implements Problem 04: Delegatable capability tokens with cascading revocation.

This update addresses reviewer feedback by tightening DelegatableAuth behavior, splitting stale-parent and transitive-revocation attacks, enforcing audience binding, making delegation scope-gated, and removing nondeterminism from the jwt contrast path.

## What changed

- Merged latest `main` and resolved shared-file conflicts.
- Added documented auth plugin entry point in `packages/nest-plugins-reference/pyproject.toml`.
- Preserved `verify(token, presenter=None)` compatibility with the base Auth protocol.
- Changed delegated token verification to fail closed when `presenter` is omitted.
- Enforced audience binding for delegated tokens.
- Enforced `"delegate"` scope before a token can mint child tokens.
- Added adversarial leaf sub-delegation rejection.
- Split `stale_parent` from `transitive_revocation` in the scenario trace.
- Tightened validators to check event ordering and specific error reasons.
- Made YAML `root_scopes` and `delegation_tree` load-bearing in scenario construction.
- Fixed deterministic clock injection for `JwtAuth`.
- Reworked jwt contrast test to resolve paths from `__file__`, write traces to `tmp_path`, and swap layers with `model_copy`.
- Added property-style coverage for delegation validity and cascading revocation.

## Why this is distinct

Unlike baseline entries with hardcoded topology, this scenario makes `root_scopes` and `delegation_tree` load-bearing. The trace therefore proves configurable delegation behavior instead of only proving one fixed hardcoded path.

## Validation

```bash
uv sync
uv run ruff check .
uv run ruff format --check .
uv run pyright
uv run pytest -v
```

Optional local scenario validation:

```bash
nest run delegated_auth -o /tmp/delegated_auth.jsonl
python -c "
from pathlib import Path
for r in validate_trace(Path('/tmp/delegated_auth.jsonl'), 'delegated_auth'):
    print(('PASS' if r.passed else 'FAIL'), r.name, r.detail)
"
```

## Reviewer feedback addressed

* Branch conflicts resolved and CI-equivalent checks run locally.
* `stale_parent` now revokes the direct parent, while `transitive_revocation` revokes the grandparent/root ancestor.
* Delegated token verification now fails closed when presenter is missing.
* Delegation now requires the parent token to hold `"delegate"` scope.
* YAML config now drives root scopes and delegation tree.
* Jwt contrast path now uses deterministic clock injection.
* Contrast test no longer depends on current working directory and no longer writes into repo `./traces`.

## Tradeoffs

* Keeps the delegation model tree-shaped rather than DAG-shaped for deterministic trace validation.
* Uses local deterministic token lineage rather than network introspection.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -26,6 +26,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("registry", "in_memory"): f"{_REF}.registry.in_memory:InMemoryRegistry",
     ("registry", "gossip"): f"{_REF}.registry.gossip:GossipRegistry",
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
+    ("auth", "delegatable"): f"{_REF}.auth.delegatable:DelegatableAuth",
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
     ("trust", "agent_receipts"): f"{_REF}.trust.agent_receipts:AgentReceiptsTrust",
     ("trust", "parc"): f"{_REF}.trust.parc:ParcTrust",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -145,3 +145,7 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("rogue_trusted_agent", rogue_trusted_agent_factory)
+    elif name == "delegated_auth":
+        from nest_core.scenarios_builtin.delegated_auth import delegated_auth_factory
+
+        register_scenario("delegated_auth", delegated_auth_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py
@@ -1,0 +1,258 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Delegated auth scenario — coordinator delegates scoped tokens through a tree.
+
+A coordinator issues a root capability token, delegates subsets to three
+intermediaries, and each intermediary delegates read-only tokens to four leaf
+agents. The trace also exercises adversarial paths: scope escalation, TTL
+violation, parent revocation, audience confusion, and transitive revocation.
+
+With ``auth: delegatable`` every attack is blocked or rejected and validators
+PASS. With ``auth: jwt`` the same script falls back to naive re-issuance with
+no parent chain or audience binding, so validators FAIL.
+
+Trace-line protocol (auditor ``send`` bodies, ``:``-delimited)::
+
+    auth_issue:<agent>:<tok_id>:<scopes>
+    auth_delegate:<parent>:<audience>:<tok_id>:<scopes>:<ttl>:<outcome>[:<detail>]
+    auth_verify:<presenter>:<tok_id>:<outcome>:<detail>
+    auth_revoke:<agent>:<tok_id>
+    auth_attack:<attack_type>
+
+Example::
+
+    agents = delegated_auth_factory(config, plugins)
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from nest_core.scenario import ScenarioConfig
+from nest_core.sim.agent import AgentContext, StateMachineAgent
+from nest_core.types import AgentId, Token
+
+_AUDITOR = AgentId("auditor-0")
+
+
+class _TokenStore:
+    """Shared token state across all agents in the delegation tree."""
+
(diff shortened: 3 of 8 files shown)


dhve:
Thanks @Niruuu2005. This is a solid read of Problem 04. `DelegatableAuth` keeps `verify(token, presenter=None)` compatible with the base `Auth` protocol instead of adding a required positional (a mistake I have had to call out on another Problem 04 entry), the typed error taxonomy (`ScopeEscalationError`, `RevokedAncestorError`, `AudienceMismatchError`, etc.) makes the validators legible, and `test_jwt_fails_adversarial_validator` running the real scenario under `jwt` inside pytest is exactly the fails-on-baseline proof we want.

**1. Branch conflicts with main, and CI has not run.**

GitHub reports the branch as CONFLICTING. The shared files you touch (`plugins.py`, `scenarios.py`, `validators.py`, `test_validators.py`) are the conflict hot spots; rebase onto latest main. No checks have run on this branch yet; the judge panel does not score broken PRs, so confirm all five stages green after rebasing (`uv sync`, `ruff check`, `ruff format --check`, `pyright`, `pytest -v`).

**2. Your `stale_parent` attack is the same event as `transitive_revocation`.**

In `scenarios_builtin/delegated_auth.py`, `_attack()` maps both attack types to the same action: verify `leaf-0` after `revoke:coordinator-0`. leaf-0's chain runs through intermediary-0 to coordinator-0, so both exercise only the revoked-grandparent path and the two validators pass or fail together. Problem 04's direct-parent case ("parent token expired or revoked but child still verifies") is covered in unit tests but never in the trace. Fix: have `attack:stale_parent` revoke `intermediary-0` and then verify a leaf under it. Also: `validate_delegated_auth_stale_parent` accepts any `:rejected:` verify containing `ValueError` anywhere in the trace, with no ordering against the `auth_revoke` line; compare event indices so a rejection that predates the revocation cannot count. For test_rigor, five validators are effectively four right now.

**3. Audience binding fails open when `presenter` is omitted.**

In `DelegatableAuth.verify()` the check is `if presenter is not None and str(presenter) != aud`, so any caller on the plain protocol surface (`verify(token)`) gets no audience enforcement. For an auth-safety-engineer persona this should fail closed: if `kind == "delegate"` and `presenter is None`, raise `AudienceMismatchError`. Root tokens still verify bare, so protocol compatibility is untouched, and your scenario always passes presenter.

**4. The `delegate` scope is decorative.**

The scenario hands intermediaries `read,delegate` and leaves only `read`, implying delegation is scope-gated, but `delegate()` never checks the parent holds `"delegate"`; a `["read", "write"]` token can mint children fine. Enforce it and add the adversarial case (a leaf attempts to sub-delegate). Same theme: `task.config.root_scopes` and `delegation_tree` in the YAML are never read, `_build_schedule()` hardcodes the tree. With about twenty open PRs on this problem, making these load-bearing (and saying so in the PR body) is your clearest novelty differentiator.

**5. The jwt contrast path leaks wall-clock time, and the contrast test has side effects.**

`_instantiate_auth` only injects a clock when the class has `set_clock`, so `JwtAuth` (which accepts `clock=` but has no `set_clock`) is built with `clock=None` and falls back to `time.time()`; real timestamps then land in trace lines via `_tok_id`'s `str(token)[:16]` fallback. Pass `clock=0.0` to `JwtAuth` too. Also, `test_jwt_fails_adversarial_validator` resolves the YAML relative to CWD (silently skipping otherwise) and writes `./traces/delegated_auth_jwt_contrast.jsonl` into the working tree. Follow `test_escrow_scenario.py`: resolve from `Path(__file__).resolve().parents[3]`, write the trace to a tempdir, swap layers via `model_copy`.

**Suggested path forward:**

1. Rebase onto main, resolve the shared-file conflicts, and add the missing `[project.entry-points."nest.plugins.auth"]` table in `packages/nest-plugins-reference/pyproject.toml` beside your `_BUILTINS` entry (the discovery path CONTRIBUTING.md documents, and part of api_fit). Then get CI green.
2. Split `stale_parent` from `transitive_revocation` and tighten both validators (ordering plus specific errors).
3. Fail closed on missing presenter for delegated tokens; enforce the `delegate` scope in `delegate()`.
4. Fix the jwt clock injection and rework the contrast test per `test_escrow_scenario.py`.
5. Worth it in this crowded field: one hypothesis property test (delegation succeeds iff strict scope subset and ttl within parent remaining; revoking any ancestor invalidates every descendant), plus a PR-body line on what sets this entry apart.

PR #142, opened 2026-07-09
[Hackathon] auth-engineer: delegatable capability tokens with cascading revocation

erin-adams:
# Delegatable capability tokens with cascading revocation (auth layer)

Solves hackathon problem **04 — Delegatable capability tokens with cascading revocation**.
Persona: an auth engineer who cares about least-privilege delegation and clean revocation.

## Motivation

The default `jwt` auth plugin (`jwt_auth.py`) can only *issue* and *revoke* whole
tokens, and its revocation set is keyed by the exact token string. It cannot
express the most common multi-agent auth pattern: an orchestrator holds a
long-lived root capability, mints a **narrower, time-bounded** sub-capability
for a worker agent *without* asking the issuer, and can pull the whole subtree
back by revoking one ancestor. Today, revoking a parent does nothing to a child
that was handed out separately — the child sails through `verify`.

This is exactly what you need to model an LLM agent sub-renting a tool to
another agent and cleanly withdrawing it. It was one of the ten open problems
and the only one with no plugin on `main`.

## Design

A macaroon (Birgisson et al., 2014). A token carries an ordered **chain of
caveats** (root → leaf). Each link is MAC'd with an HMAC **keyed by its parent's
signature**, so:

- **Delegation needs no issuer round-trip.** Any holder of a valid parent can
  append a caveat and re-chain the MAC (`delegate(parent, audience, scopes_subset, ttl)`).
  The root secret never leaves the issuer.
- **Restriction is enforced at verify, not at mint.** A holder knows the parent
  signature (the child's MAC key), so it *can* forge a MAC-valid caveat with wider
  scopes. `verify` therefore re-checks that every child only **narrows** its
  parent — scopes ⊆ parent, expiry ≤ parent, `parent_tid` linked. A widened
  caveat is cryptographically well-formed but rejected. **This is what actually
  stops scope escalation** (see the `test_forged_wide_caveat_rejected_at_verify`
  test — the honest `delegate`-time check alone is not enough).
- **Revocation cascades by construction.** Every caveat carries a `tid`; `verify`
  checks every `tid` in the chain against the revocation set. Revoke a parent's
  `tid` and every descendant fails at the next `verify` — no per-child bookkeeping.
- **Audience binding.** The leaf caveat is bound to an `aud`; `verify(token, presenter=...)`
  rejects a mismatched presenter. `presenter` is optional, so the plugin still
  satisfies the `Auth` protocol exactly (`isinstance(plugin, Auth)` is `True`).

### Tradeoffs (explicit)

- **Chain grows with depth.** Each hop adds one caveat + re-MAC; verify is O(depth).
  For agent delegation trees (depth ≤ a handful) this is negligible; I did not add
  QC aggregation.
- **Third-party caveats / discharge macaroons are out of scope.** Only first-party
  caveats (scope, expiry, audience) are implemented — enough for the problem.
- **Revocation is a growing set of `tid`s.** Bounded by tokens issued; no GC. Fine
  for a simulation; a real deployment would TTL the revocation entries.
- **Errors subclass `ValueError`** so existing callers that already guard `verify`
  with `except ValueError` keep working unchanged.

## Determinism (Tier 1)

Token ids are `sha256` over the caveat's own fields — no RNG, no wall clock.
Timestamps come from an injected logical clock (`set_clock`), exactly as `JwtAuth`
uses its `clock`. Same inputs → byte-identical tokens; the scenario trace is
**byte-identical under seeds 42, 7, 1337**.

## Adversarial validators (FAIL on `jwt`, PASS on `delegatable`)

Under the `delegated_auth` scenario, three validators stage the three attacks the
problem names and read the outcome the *configured* plugin actually returned:

| Validator                     | Attack                                    | `jwt`                   | `delegatable`          |
| ----------------------------- | ----------------------------------------- | ------------------------- | ------------------------ |
| `auth_no_scope_escalation`  | agent obtains a scope it wasn't delegated | accepted →**FAIL** | blocked →**PASS** |
| `auth_cascading_revocation` | child used after an ancestor is revoked   | accepted →**FAIL** | blocked →**PASS** |
| `auth_audience_binding`     | token presented by the wrong agent        | accepted →**FAIL** | blocked →**PASS** |

The scenario driver (`coordinator` + 3 `intermediary` + 12 `leaf` + `auditor`) is
capability-gated on `hasattr(auth, "delegate")`: under `jwt` it honestly falls
back to independent `issue` calls, which is what JWT can actually express — and
that is precisely why the attacks succeed there.

## Verify it yourself

```bash
uv sync

# Plugin + property tests
uv run pytest packages/nest-plugins-reference/tests/test_delegatable_auth.py \
              packages/nest-plugins-reference/tests/test_delegatable_auth_properties.py -q

# Run the scenario and check the validators PASS under delegatable
uv run nest run scenarios/delegated_auth.yaml -o ./traces/delegated_auth.jsonl
uv run python -c "
from pathlib import Path
for r in validate_trace(Path('traces/delegated_auth.jsonl'), 'delegated_auth'):
"

# Flip auth -> jwt and watch all three validators FAIL
sed 's/auth: delegatable/auth: jwt/' scenarios/delegated_auth.yaml > /tmp/da_jwt.yaml
uv run nest run /tmp/da_jwt.yaml -o ./traces/da_jwt.jsonl
uv run python -c "
from pathlib import Path
for r in validate_trace(Path('traces/da_jwt.jsonl'), 'delegated_auth'):
"
```

## CI

`uv sync`, `ruff check .`, `ruff format --check .`, `pyright`, `pytest -v` all pass
(**976 passed, 1 skipped**). No new runtime dependencies.

## Files

- `packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable.py` — the plugin
- `packages/nest-core/nest_core/plugins.py` — register `("auth", "delegatable")`
- `packages/nest-core/nest_core/validators.py` — three validators + `delegated_auth` registry entry
- `packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py` — scenario driver
- `packages/nest-core/nest_core/scenarios.py` — register the scenario factory
- `scenarios/delegated_auth.yaml` — the scenario
- `packages/nest-plugins-reference/tests/test_delegatable_auth{,_properties}.py` — tests
- `packages/nest-core/tests/test_validators.py` — validator unit tests + registry update

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -26,6 +26,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("registry", "in_memory"): f"{_REF}.registry.in_memory:InMemoryRegistry",
     ("registry", "gossip"): f"{_REF}.registry.gossip:GossipRegistry",
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
+    ("auth", "delegatable"): f"{_REF}.auth.delegatable:DelegatableAuth",
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
     ("trust", "agent_receipts"): f"{_REF}.trust.agent_receipts:AgentReceiptsTrust",
     ("trust", "parc"): f"{_REF}.trust.parc:ParcTrust",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -145,3 +145,7 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("rogue_trusted_agent", rogue_trusted_agent_factory)
+    elif name == "delegated_auth":
+        from nest_core.scenarios_builtin.delegated_auth import delegated_auth_factory
+
+        register_scenario("delegated_auth", delegated_auth_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py
@@ -1,0 +1,405 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Delegated-capability scenario: a coordinator builds a delegation tree and
+stages the three attacks the ``delegatable`` auth plugin is built to defeat.
+
+Topology: one ``coordinator`` roots a delegation tree over three
+``intermediary`` agents, each of which sub-delegates to four ``leaf`` agents
+(twelve leaves total), plus one ``auditor`` sink that records the structured
+trace lines the validators parse.
+
+The coordinator resolves its auth plugin from ``ctx.plugins["auth"]`` and drives
+the whole tree through it — *capability-gated* on ``hasattr(auth, "delegate")``:
+
+* Under ``auth: delegatable`` the coordinator delegates real macaroon-style
+  sub-tokens (scopes narrow at each hop) and every staged attack is rejected by
+  the plugin, so each ``attack:*`` line records ``blocked``.
+* Under ``auth: jwt`` (the default) there is no ``delegate``; the coordinator
+  falls back to independent ``issue`` calls, which is the honest picture of
+  what JWT can express. Each attack then succeeds — an agent mints itself
+  broader scopes (no delegation constraint), a revoked parent leaves its
+  separately-issued child usable, and a token has no audience to bind — so each
+  ``attack:*`` line records ``accepted`` and the validators fail.
+
+That contrast is the point: the same scenario **passes** under ``delegatable``
+and **fails** under ``jwt``, with no change to the driver.
+
+Every security decision runs locally against the shared plugin at tick 0, so the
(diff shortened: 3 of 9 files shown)


shilpa-kulkarni-14:
@erin-adams Thanks for this, it's a solid, independent macaroon implementation (clean error
hierarchy, injected logical clock, cascading revocation by tid).

Closing as a duplicate of #138, which implements the same feature (delegatable
capability tokens with cascading revocation), registers the same `delegated_auth`
scenario, and touches the same file paths, so only one can land. #138 is the earlier
PR and is already reviewed and cleared, with a verified discriminator (delegatable
passes, jwt fails) and a verify_presented path this one doesn't have.

Your version does register its validators in core, which is slightly better wired
than #138's, we'll fold that improvement into #138. This isn't a knock on your work;
the two implementations are genuinely independent and both are competent. The auth
delegatable slot is just already filled.

If you'd like to keep contributing, an unclaimed layer (e.g. a distinct memory or
coordination capability) would be a stronger use of the same effort. Happy to point
you at open gaps.

erin-adams:
All good, it was my first hackathon, so I just wanted the experience! Hope the event went well for everyone!
Erin Adams
________________________________

PR #139, opened 2026-07-09
feat(trust): implement 2-pass SybilResistantTrust plugin

kvcops:
## Overview
This PR implements a robust, security-hardened **2-Pass Eigen-weighted Trust & Reputation plugin** (`sybil_resistant`) under the Trust layer, along with a dedicated test scenario and active decision-gating to prevent Sybil attacks, collusion rings, and ballot-stuffing.
The naive `score_average` baseline is easily manipulated by Sybil identities circle-vouching each other. This plugin introduces an algorithmic defense that isolates and penalizes malicious clusters in P2P agent networks.
---
### 1. Algorithmic Trust Model (`sybil_resistant.py`)
- **Pass 1 (Baseline)**: Computes a baseline reputation score for all reporters based on order-sequenced, burst-decayed opinions.
- **Pass 2 (Eigen-weighted)**: Computes the final reputation score of a subject by weighting each reporter's opinion by their Pass 1 reputation. Trusted reporters have a high influence, while untrusted/cheating reporters have their weight restricted.
- **Collusion Ring Penalty**: A directed-graph algorithm detects mutual-boosting pairs ($A \rightarrow B$ and $B \rightarrow A$ with no negative reports). Ring members are penalized with a $0.1$ multiplier on their reporting weights.
- **Burst Damping**: Uses an exponentially decayed step size $\alpha$ to damp high-frequency positive report floods from a single identity, capping maximum score inflation.
- **Negative Feedback Dominance**: Negative reports are weighted 3.0x more heavily than positive reports to ensure one cheat immediately contains and isolates an attacker.
### 2. Active Decision Gating (`sybil_reputation.py`)
To prevent a "vacuous" simulation trace where agents ignore trust:
- `HonestTrader` query `trust.score()` before accepting trade requests or delivering payloads.
- If an agent's reputation drops below `0.45`, honest traders reject the trade, send a `refuse:` message, and select a different peer.
- `SybilObserver` dynamically calls `trust.report()` on trade outcomes, ensuring reputation scores update in real-time.
---
## Validation Differential & Tests
We updated the validators in `validators.py` and implemented a dedicated integration test scenario `test_sybil_scenario.py` to ensure the validation suite successfully flips:
- `validate_sybil_flood_resistance`: FAILS if cheats $\ge 12$ in the trace.
- `validate_sybil_collusion_ring`: FAILS if there are no `refuse:` messages in the trace.
### Verification Results
[...]

PR #143, opened 2026-07-09
hackathon/john-kioko-weighted-trust: WeightedTrust plugin

Mutinda-Kioko:
## WeightedTrust — recency-decayed, volume-weighted reputation

A new trust plugin that improves on `ScoreAverageTrust` with:

- **Recency decay** — exponential half-life weighting so recent evidence matters more than stale evidence
- **Evidence severity** — `positive` (1.0), `neutral` (0.5), `negative` (0.0), and `byzantine` (-0.5) are weighted differently instead of the flat binary average
- **Volume confidence** — more data points produce higher confidence, capped at 1.0
- **Reporter credibility tracking** — reporters with a good track record are weighted more than unknown reporters

### Same interface as ScoreAverageTrust
Drop-in replacement — same `score()`, `report()`, `attest()`, and `stake()` methods. Selectable as `trust: weighted` in scenario YAML.

### Files added
- `nest_plugins_reference/trust/weighted.py` — the plugin
- `tests/test_weighted_trust.py` — 10 tests (all passing)
- `scenarios/weighted_trust.yaml` — scenario using the weighted trust layer

### Files modified
- `nest_core/plugins.py` — registered `("trust", "weighted")` in the builtin registry

### Test results
```
10 new tests — ALL PASS
37 existing tests — ALL PASS (no regressions)
```

Author: **John Kioko** (@Mutinda-Kioko)

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -26,6 +26,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("registry", "gossip"): f"{_REF}.registry.gossip:GossipRegistry",
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
+    ("trust", "weighted"): f"{_REF}.trust.weighted:WeightedTrust",
     ("trust", "agent_receipts"): f"{_REF}.trust.agent_receipts:AgentReceiptsTrust",
     ("payments", "prepaid_credits"): f"{_REF}.payments.prepaid_credits:PrepaidCredits",
     ("payments", "streaming"): f"{_REF}.payments.streaming:StreamingPayments",
--- a/packages/nest-plugins-reference/nest_plugins_reference/trust/weighted.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/trust/weighted.py
@@ -1,0 +1,170 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Weighted trust plugin — recency-decayed, volume-weighted reputation.
+
+A richer alternative to ScoreAverageTrust that considers:
+
+  * **Recency** — newer evidence counts more than stale evidence via an
+    exponential decay factor.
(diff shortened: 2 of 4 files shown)


shilpa-kulkarni-14 (changes requested):
@Mutinda-Kioko -Thanks for this. The plugin is reasonably built (property tests, clamping, full
Trust conformance), but I can't merge it, and the blockers are structural rather
than cosmetic. Verified on latest main (3510921 + this PR).

Two gaps:

1. No discriminator. The tests are property tests of the plugin in isolation, with
   zero baseline comparison, and there's no validator or registered scenario. So
   nothing passes on WeightedTrust and fails on score_average. The core claim, that
   weighting beats naive averaging, is never demonstrated. This is the gate that
   separates a rankable contribution from a configuration.

2. Wall-clock dependence. score() uses `now = time.time()` for decay, and evidence
   timestamps default to time.time(). A trace is meant to be a pure function of seed
   and inputs, reproducible at the same seed next year. This one depends on the
   absolute wall-clock time score() runs at. It only reproduced on my back-to-back
   runs because they were milliseconds apart. Drive decay from a caller-supplied tick
   or the evidence timestamps only, never the system clock.

Also: ci-local is red (ruff I001, format, and 3 pyright reportPrivateUsage from tests
reaching into _stakes / _reporter_scores). Minor and mechanical, but red.

Novelty: recency-decayed, reporter-weighted reputation overlaps #139 (2-pass reporter
weighting) and is a weighted score_average. Worth an overlap decision with the
maintainers on whether it adds a distinct, proven capability.

Path to merge-ready:
1. Add a validator + scenario where an attack score_average fails and WeightedTrust
   catches (e.g. stale positive reports that should decay away), wired so swapping to
   score_average flips the validator.
2. Replace time.time() with a caller-supplied tick/timestamp, then verify byte-
   determinism across seeds 42/7/1337.
3. Clear ci-local (ruff format, fix import order, use public API or scoped noqa).

Happy to re-review once there's a non-vacuous discriminator and the clock is out.

Mutinda-Kioko:
Closing in favor of an updated PR with all reviewer feedback addressed (discriminator, determinism, CI). New PR opening now.

PR #144, opened 2026-07-09
[Hackathon] Bind score-average trust evidence to subjects

himanshu748:
## Problem

`ScoreAverageTrust` trusted two caller-controlled references to the subject independently. A malicious caller could report evidence about `victim` while mutating `innocent`, and an attestation could pair an arbitrary `agent` argument with a claim about someone else. Unknown evidence kinds were also recorded as neutral samples, allowing meaningless reports to raise confidence to 1.0. Finally, identity-signed attestations reported `issuer=system` even when the signature named the real signer.

## Change

- Reject reports when `agent != evidence.subject` before state mutation.
- Reject unsupported evidence kinds instead of counting them as neutral samples.
- Reject attestations when `agent != claim.subject`.
- Bind `Attestation.issuer` to `Signature.signer` for both system and identity-backed signatures.
- Document the fail-closed behavior in the public method docstrings.

## Adversarial coverage

Four regression tests prove that:

1. evidence about one subject cannot mutate another subject;
2. unknown kinds cannot inflate sample count or confidence;
3. a mismatched claim cannot be attested; and
4. an identity-backed attestation has a verifiable issuer/signer binding.

## Verification

`make ci-local` passes all five gates:

- `uv sync`
- `ruff check`
- `ruff format --check`
- `pyright` (0 errors)
- `pytest` (984 passed, 1 skipped, 1 deselected)

### DIFF
--- a/packages/nest-plugins-reference/nest_plugins_reference/trust/score_average.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/trust/score_average.py
@@ -58,29 +58,43 @@ class ScoreAverageTrust:
     async def attest(self, agent: AgentId, claim: Claim) -> Attestation:
         """Create an attestation about an agent.
 
+        The requested agent must match the claim subject. When an identity
+        signs the claim, the signature signer is also recorded as the issuer.
+
         Example::
 
             att = await trust.attest(AgentId("a1"), claim)
         """
+        if agent != claim.subject:
+            msg = "attestation agent must match claim subject"
+            raise ValueError(msg)
         sig = Signature(signer=AgentId("system"), value=b"attestation", algorithm="none")
         if self._identity is not None:
             sig = self._identity.sign(claim.model_dump_json().encode())
-        return Attestation(issuer=AgentId("system"), claim=claim, signature=sig)
+        return Attestation(issuer=sig.signer, claim=claim, signature=sig)
 
     async def report(self, agent: AgentId, evidence: Evidence) -> None:
         """Report evidence, updating the agent's score.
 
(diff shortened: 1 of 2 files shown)


shilpa-kulkarni-14:
@himanshu748  Thanks for this. These are real fixes, not cosmetic.

Verified locally on the PR head (4f79104):
- ci-local 5/5 green (ruff, format, pyright 0 errors, pytest 984 passed).
- The three fixes work: subject-mismatched report/attest rejected, unknown
  evidence kinds rejected, attestation issuer bound to the signer.
- Critically, the hardening is surgical: a subject-matched positive flood still
  inflates score_average to 1.000 over 20 samples, so the baseline stays
  defeatable and the discriminators in #53/#89/#139 remain non-vacuous.

Accepting as a baseline hardening. Note for the record: score_average is now
correct-but-naive by design (correct subject attribution, still defeatable by
volume). Any future discriminator must not rely on the old lax behavior
(subject-mismatched or unknown-kind evidence being accepted).

This is a maintenance merge, not a ranked stack entry, since there's no plugin,
scenario, or discriminator. Let me know your thoughts.

himanshu748:
Thanks, @shilpa-kulkarni-14 — agreed on that boundary. The intent of #144 is only to harden the built-in baseline against malformed or misattributed evidence, not to turn `score_average` into a ranked stack or make a discriminator depend on behavior we now consider invalid. I’ll keep this PR maintenance-focused. Any future ranked plugin/scenario/discriminator will be a separate contribution with an independently meaningful attack model and hardened implementation, without relying on the previous lax path. Thanks for verifying the fixes locally and for the clear guidance.

PR #82, opened 2026-07-07
[Hackathon] privacy: Trust-Gated Selective Disclosure

tanishaacodes:
### Layer Picked
Privacy (Layer 11) — Reference plugin implementation of Trust-Gated Selective Disclosure.
### Why
The current Privacy layer in NEST is a noop (stub passthrough). This is a "sharp edge" in the simulator: agents in the reputation or marketplace scenarios leak full internal state (GPS, IDs, stock levels) to every peer, regardless of whether that peer is a verified partner or a known malicious Byzantine actor.
In a decentralized "Internet of Agents," privacy is not a binary toggle; it is a function of established reputation. This PR replaces the noop stub with a plugin that enforces Differential Fidelity,masking sensitive data based on the recipient's trust score.
### Core Idea
The TrustGatedPrivacy plugin creates a functional link between Layer 11 (Privacy) and Layer 6 (Trust).
Cross-Layer Interop: When mask() is called, the plugin queries the active Trust layer for the recipient's score().
**Fidelity Tiers**:
High Trust (>0.8): Full fidelity (raw data).
Medium Trust (0.4–0.8): Semantic blurring (generalized data stubs).
Low Trust (<0.4): Full redaction.
Deterministic: The masking is purely a function of the trust score, maintaining the simulator's bit-identical trace guarantees under the same seed.
### How to Test
**uv run python test_privacy.py**
Sample result: --- PRIVACY LAYER TEST ---
High Trust Result: 19.23N, 73.85E
Low Trust Result: REDACTED
--- TEST PASSED ---
### Headline Adversarial Property
test_privacy_leakage_to_untrusted_peer: Proves that a malicious agent with a low reputation score cannot extract high-resolution location data from an honest agent, even if the honest agent initiates the communication.
### Persona
Security Architect interested in Zero-Trust Agentic Networks. I want Nanda Town to be a testbed for protocols where agents "earn" access to information through reputation.

PR #146, opened 2026-07-09
[Hackathon]: add replay-safe comms plugin with sequence-bound envelopes

AdityaDarke97:
## What this adds

A new `replay_safe` comms plugin that extends `authenticated` with **replay attack detection** via per-peer monotonic sequence numbers. This PR closes the gap explicitly called out in PR #114:

> "out scope replay: a verbatim re-send of a genuine envelope still verifies. Bind a nonce/sequence into metadata to close that separately."

## Why this matters

`authenticated` solved **tamper-evidence** (HMAC-SHA256 over the envelope detects version rollback and field stripping). However, it has a critical vulnerability: **a captured genuine envelope can be replayed verbatim, and the receiver cannot tell.**

Two concrete attacks slip through `authenticated`:
1. **Message Replay:** An adversary captures a legitimate `bid:item-1:100` envelope and re-sends it later. The receiver processes it again, double-counting the bid.
2. **State Manipulation:** An adversary replays a `rotate:key:old:new` envelope to force an agent to re-apply an old rotation, potentially resurrecting a compromised key.

## Empirical Proof

I wrote adversarial validators that test both plugins:

```python
from nest_plugins_reference.comms.authenticated import AuthenticatedComms
from nest_plugins_reference.comms.replay_safe import ReplaySafeComms
from nest_plugins_reference.validators import (
    check_comms_replay_rejected,
)

# Test authenticated (should FAIL - allows replay)
alice_auth = AuthenticatedComms(AgentId("alice"))
envelope = alice_auth.serialize(msg)
result_auth = check_comms_replay_rejected(alice_auth, envelope, "alice", "bob")
print(f"Authenticated allows replay: {not result_auth}")  # True

# Test replay_safe (should PASS - rejects replay)
alice_safe = ReplaySafeComms(AgentId("alice"))
envelope = alice_safe.serialize(msg)
result_safe = check_comms_replay_rejected(alice_safe, envelope, "alice", "bob")
print(f"ReplaySafe rejects replay: {result_safe}")  # True
```
Result:

    authenticated → FAILS (allows replay attacks)
    replay_safe → PASSES (rejects replay attacks)

How it works
Each agent maintains a per-peer sequence counter that increments with every outgoing message. The receiver tracks the highest seen sequence per peer and rejects any envelope with a sequence number ≤ the last accepted value.
Key features:

    Strict superset of authenticated: Inherits HMAC authentication, version negotiation, and unknown-field preservation
    Sliding window buffer: Handles out-of-order delivery (up to WINDOW_SIZE=10 messages ahead)
    Per-peer tracking: Sequences are independent for each sender-receiver pair

Scenario File
Added scenarios/comms_replay_attack.yaml that runs 10 peers through the replay_safe comms plugin:

    Task: Reuses comms_versioning traffic generator
    Failures: message_drop: 0.05 — intentionally drops 5% of packets to trigger and verify the sliding window out-of-order buffer
    Output: Generates traces/comms_replay_attack.jsonl for trace validation
    
Test Results
All 8 unit tests pass:

uv run pytest packages/nest-plugins-reference/tests/test_replay_safe.py -v

test_serialize_deserialize_roundtrip PASSED
test_sequence_increments PASSED
test_replay_rejected PASSED
test_adversarial_validator_replay PASSED
test_adversarial_validator_rollback PASSED
test_different_peers_independent_sequences PASSED
test_concurrent_sending_no_duplicates PASSED
test_out_of_order_delivery_buffering PASSED

8 passed

### DIFF
--- a/docs/layers/communication.md
+++ b/docs/layers/communication.md
@@ -58,6 +58,27 @@ channel key. Key exchange (the pre-shared secret stands in for an ECDH session
 key) and replay of verbatim envelopes are out of scope — bind a nonce into
 `metadata` to close replay separately.
 
+### `replay_safe`
+Extends `authenticated` with replay attack detection via per-peer monotonic sequence numbers.
+
+**Features:**
+- Rejects verbatim replays of captured envelopes
+- Sliding window buffer for out-of-order delivery (WINDOW_SIZE=10)
+- Per-peer sequence tracking (independent counters for each sender-receiver pair)
+- Strict superset of `authenticated` (inherits all tamper-evidence features)
+
+**When to use:**
+- Financial transactions or state mutations where replay is a concern
+- Scenarios requiring message ordering guarantees
+- Production systems needing both tamper-evidence AND replay protection
+
+**Trade-offs:**
+- Slightly larger envelope size (adds `sequence` field)
+- Requires state tracking (per-peer sequence counters)
+- Does not handle crash recovery (sequences reset on restart)
+- Cross-peer ordering not enforced (sequences are per-pair)
+
+
 ## Writing your own
 
 See [`writing-a-plugin.md`](../writing-a-plugin.md). Register under
--- a/packages/nest-plugins-reference/nest_plugins_reference/comms/replay_safe.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/comms/replay_safe.py
@@ -15,13 +15,12 @@ Threat model:
 from __future__ import annotations
 
 import base64
-import hashlib
 import hmac
 import json
 from collections import defaultdict
 from typing import Any, cast
 
-from nest_core.types import AgentCard, AgentId, Message, MessageId, Query, Response
+from nest_core.types import AgentId, Message, MessageId, Response
 
 from nest_plugins_reference.comms.authenticated import (
     AUTH_TAG_FIELD,
@@ -32,12 +31,19 @@ from nest_plugins_reference.comms.authenticated import (
     AuthenticatedComms,
     DowngradeError,
     UnsupportedSchemaError,
-    _parse_major,
     expected_auth_tag,
 )
 
 SEQUENCE_FIELD = "sequence"
 WINDOW_SIZE = 10
+
+
+def _parse_major(version: str) -> int:
+    """Parse major version from 'major.minor' string."""
+    try:
+        return int(version.split(".")[0])
+    except (ValueError, IndexError):
+        return 1
 
 
 class ReplayError(DowngradeError):
@@ -65,12 +71,15 @@ class ReplaySafeComms(AuthenticatedComms):
         require_auth: bool = False,
     ) -> None:
         super().__init__(
-            agent_id, transport, registry,
-            channel_secret=channel_secret, require_auth=require_auth,
+            agent_id,
+            transport,
+            registry,
+            channel_secret=channel_secret,
+            require_auth=require_auth,
         )
         self._outgoing_sequences: dict[tuple[str, str], int] = {}
         self._incoming_sequences: dict[tuple[str, str], int] = {}
-        self._out_of_order_buffer: dict[tuple[str, str], dict[int, bytes]] = defaultdict(dict)
+        self._out_of_order_buffer: dict[tuple[str, str], dict[int, Any]] = defaultdict(dict)
 
     def serialize(self, msg: Message) -> bytes:
         meta: dict[str, Any] = dict(msg.metadata)
@@ -83,11 +92,15 @@ class ReplaySafeComms(AuthenticatedComms):
         self._outgoing_sequences[pair_key] = seq + 1
 
         envelope: dict[str, Any] = {
-            "schema_version": version, "kind": kind, "id": str(msg.id),
-            "sender": str(msg.sender), "receiver": str(msg.receiver),
+            "schema_version": version,
+            "kind": kind,
+            "id": str(msg.id),
+            "sender": str(msg.sender),
+            "receiver": str(msg.receiver),
             "payload": base64.b64encode(msg.payload).decode("ascii"),
             "correlation_id": str(msg.correlation_id) if msg.correlation_id else None,
-            "timestamp": msg.timestamp, "metadata": meta,
+            "timestamp": msg.timestamp,
+            "metadata": meta,
             SEQUENCE_FIELD: seq,
         }
         for key, value in unknown.items():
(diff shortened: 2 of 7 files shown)


AdityaDarke97:
EDIT: SkillMD: https://github.com/AdityaDarke97/neuronmesh/blob/main/SKILL.md

shilpa-kulkarni-14:
@AdityaDarke97 - The design is right and the functionality works, but the execution does not clear the quality bar. This is the opposite profile from #82, which had a good idea and the wrong interface. Here the interface and composition are correct, but the code is not clean enough to merge: 49 type errors and a fragile dependency on the parent's private internals. The fixes are mechanical plus one design decision.

Path to merge-ready :

- Clear ci-local. Remove unused imports, add type arguments to bare dict, fix the partially-unknown types, run ruff format, and clear the ruff issues (W293, W292, F401, I001, B011). Replace assert False with pytest.raises or an explicit raise.
- Resolve the private-helper coupling. Either promote the parent helpers (_parse_major, _verify_tag, _process_sequence_and_build) to a public or protected API in #105's authenticated.py, or reimplement them in the subclass. Reaching into another module's privates will break on any refactor.
- Reconsider the validators. Make them trace validators that pass on replay_safe and fail on authenticated, matching the pattern the merged comms PR uses, so the scenario-level discriminator is standard.
- Add a docs/layers/communication.md note for the replay-safe variant.

shilpa-kulkarni-14 (changes requested):
Requested changes- https://github.com/projnanda/nandatown/pull/146#issuecomment-4929746692

AdityaDarke97:
UPDATE: Both uv run ruff check and uv run pyright now give 0 errors, Also tried to implement the changes suggested in the files, and added documentation of the 'replay-safe' plugin. Kindly review the changes and suggest if anything doesn't work

shilpa-kulkarni-14:
> UPDATE: Both uv run ruff check and uv run pyright now give 0 errors, Also tried to implement the changes suggested in the files, and added documentation of the 'replay-safe' plugin. Kindly review the changes and suggest if anything doesn't work

@AdityaDarke97 some checks are still failing.

shilpa-kulkarni-14:
> > UPDATE: Both uv run ruff check and uv run pyright now give 0 errors, Also tried to implement the changes suggested in the files, and added documentation of the 'replay-safe' plugin. Kindly review the changes and suggest if anything doesn't work
> 
> @AdityaDarke97 some checks are still failing. Linter to be specific.

AdityaDarke97:
> > UPDATE: Both uv run ruff check and uv run pyright now give 0 errors, Also tried to implement the changes suggested in the files, and added documentation of the 'replay-safe' plugin. Kindly review the changes and suggest if anything doesn't work
> 
> @AdityaDarke97 some checks are still failing.

@shilpa-kulkarni-14 Ah, those were basic formatting checks, fixed those using the formatter. 
Locally ran the checks:
uv run pytest -v
uv run pyright
uv run ruff format --check
uv run ruff check

All pass perfectly locally for me.

shilpa-kulkarni-14 (commented):
Re-reviewed on latest main. You addressed every quality point,
thank you:

- ci-local now 5/5 green (pyright 0, down from 49; ruff and format clean; 1156 passed).
- The private coupling to #105 is gone: _parse_major / _verify_tag /
  _process_sequence_and_build are defined locally now, not reached into
  authenticated.py.
- communication.md added, and the validators are trace validators now.

One blocker remains, and it's the scenario, not the plugin. The discriminator still
doesn't flip:

    replay_safe   : replay_rejected=False, sequence_rollback_rejected=True
    authenticated : replay_rejected=False, sequence_rollback_rejected=True

check_replay_rejected_in_trace returns False on replay_safe (the plugin it should
pass on) and doesn't distinguish it from authenticated. The cause is in
comms_replay_attack.yaml: it runs `task.type: comms_versioning` and, per its own
description, stages "5% message drop to test buffering." It never re-sends a captured
envelope, so replay_safe's stale_sequence rejection never fires, and the validator
(which looks for a stale_sequence error event) finds nothing. The trace only carries
rejected_major events from the versioning task.

The plugin does reject replays; test_replay_rejected proves it at the unit level. The
scenario just never exercises that path, so the scenario-plus-validator pipeline the
rubric judges doesn't demonstrate the capability.

To close it:
1. Make the scenario actually replay: capture a valid envelope and re-send it, so
   under replay_safe the second delivery emits stale_sequence and under authenticated
   it re-verifies and is accepted.
2. Confirm check_replay_rejected_in_trace passes on replay_safe and fails on
   authenticated (right now it fails on both).
3. Fix or drop check_sequence_rollback_rejected, which passes on both and
   discriminates nothing.

This is close. Once the scenario stages the replay and the validator flips, it's
mergeable. Happy to re-verify.

AdityaDarke97:
> Re-reviewed on latest main. You addressed every quality point, thank you:
> 
>     * ci-local now 5/5 green (pyright 0, down from 49; ruff and format clean; 1156 passed).
> 
>     * The private coupling to #105 is gone: _parse_major / _verify_tag /
>       _process_sequence_and_build are defined locally now, not reached into
>       authenticated.py.
> 
>     * communication.md added, and the validators are trace validators now.
> 
> 
> One blocker remains, and it's the scenario, not the plugin. The discriminator still doesn't flip:
> 
> ```
> replay_safe   : replay_rejected=False, sequence_rollback_rejected=True
> authenticated : replay_rejected=False, sequence_rollback_rejected=True
> ```
> 
> check_replay_rejected_in_trace returns False on replay_safe (the plugin it should pass on) and doesn't distinguish it from authenticated. The cause is in comms_replay_attack.yaml: it runs `task.type: comms_versioning` and, per its own description, stages "5% message drop to test buffering." It never re-sends a captured envelope, so replay_safe's stale_sequence rejection never fires, and the validator (which looks for a stale_sequence error event) finds nothing. The trace only carries rejected_major events from the versioning task.
> 
> The plugin does reject replays; test_replay_rejected proves it at the unit level. The scenario just never exercises that path, so the scenario-plus-validator pipeline the rubric judges doesn't demonstrate the capability.
> 
> To close it:
> 
>     1. Make the scenario actually replay: capture a valid envelope and re-send it, so
>        under replay_safe the second delivery emits stale_sequence and under authenticated
>        it re-verifies and is accepted.
> 
>     2. Confirm check_replay_rejected_in_trace passes on replay_safe and fails on
>        authenticated (right now it fails on both).
> 
>     3. Fix or drop check_sequence_rollback_rejected, which passes on both and
>        discriminates nothing.
> 
> 
> This is close. Once the scenario stages the replay and the validator flips, it's mergeable. Happy to re-verify.

@shilpa-kulkarni-14 You're absolutely right about the blocker. I've been trying to figure out how to stage an actual replay attack in the scenario, but I'm stuck.

The built-in tasks (like `comms_versioning`) don't have a replay mechanism, and I haven't found documentation on creating custom tasks that can capture and re-inject envelopes.

Could you either:
1. Point me to an example scenario that stages a replay attack?
2. Tell me the correct way to create a custom task that does this?
3. Or is there a simpler approach I'm missing?

I want to get this right rather than guessing at the simulator's task system. Thanks!

AdityaDarke97:
Closing this PR Since #169 solves the same issue in a better way and works perfectly as prescribed. Even though my solution is theoretically correct but it fails to work perfectly.

PR #148, opened 2026-07-09
[Hackathon] bogacsmz: byzantine-resistant gossip registry — signed-equivocation quarantine + eclipse resistance

bogacsmz:
# feat(registry): byzantine-resistant gossip -- signed cards, signed-equivocation quarantine, eclipse-resistant sampling

**Persona for this review: trust/honesty auditor.** I built the adversarial
validators *before* trusting my own plugin's guarantees, and I am reporting
every limitation I found while building it, not just the attacks it catches.
If you are reviewing this for a security-sensitive merge, read
[`VERIFICATION.md`](nest_plugins_reference/registry/VERIFICATION.md) in
full, not just this summary.

## Headline: a validly-signed equivocator still poisons the network

The other prior art in this space (`#67`, registration-only card signing)
answers "did the claimed publisher actually sign this card?" -- necessary,
but not sufficient. It does not answer, and *cannot* answer by construction:
"did this publisher sign **two different, individually valid** cards at the
same version?" Both cards pass every check a registration-time signature can
run, in isolation. `#67` would accept either one without complaint. The only
way to catch it is to compare a publisher's writes to each other, not to
verify a signature against itself -- that is the invariant this PR adds
(`scenarios/gossip_signed_equivocation.yaml`, the novelty proof scenario).

This defense is **mesh-wide, not topology-bounded**: even an equivocator that
never sends its two conflicting cards to any common recipient -- delivering
card1 only to one group of peers and card2 only to a disjoint group, so no
single node ever directly witnesses both -- is still caught at **every**
honest node. Getting this genuinely mesh-wide took two rounds of work, and
the honest story is in [`VERIFICATION.md`](nest_plugins_reference/registry/VERIFICATION.md#honest-limitations-every-one-gathered-across-the-build)
(limitation 4). A first fix carried a per-entry **content hash** in the
`OP_DIGEST` payload so `_compute_missing` hands over a
same-version-but-different-content entry rather than treating the peer as "in
sync" -- enough for N=2 / a single equivocator
(`test_disjoint_delivery_equivocation_is_caught`). But that is **not
sufficient** at N>2 honest nodes with multiple equivocators: the instant a
node witnesses the conflict it evicts the card, dropping it from its digest,
so it stops relaying the conflicting copy -- "eviction halts relay" -- and a
node that only ever saw one side, whose neighbors holding the other side all
evict first, strands permanently (reproduced at N=10 / five equivocators,
`test_disjoint_multi_equivocator_no_stranding`). The real fix gossips the
equivocation **proof** -- the two conflicting, individually-signature-valid
writes for `(E, v)` -- independently of card eviction: the proof survives
eviction (`self._equivocation_proofs`), rides anti-entropy as its own
`OP_EQPROOF` wire item advertised via a known-byzantine digest section, and
is **independently re-verified on receipt** (both signatures + same-`(E,
v)`-different-hash, `_verify_proof`/`_ingest_proof`) so a relay cannot forge
one to frame an honest publisher. A stranded node therefore learns `E` is
byzantine from any honest neighbor, and every honest node converges on
quarantining and evicting `E` regardless of delivery topology
(`test_disjoint_multi_equivocator_no_stranding`,
`test_fabricated_equivocation_proof_does_not_frame_honest_publisher`). The
one residual honest condition: the honest sub-network must be connected (not
partitioned) for the proof to reach everyone, and convergence is eventual
(a transient in-flight window), never instantaneous.

## Threat model

`gossip` (`#24`, merged) gives eventually-consistent discovery under
honest-but-partitioned failures: every agent gossips its local view, and the
simulator's partition logic naturally blocks cross-partition propagation.
It assumes every participant plays by the rules. This PR removes that
assumption on three specific, named fronts:

| # | Attack | `gossip`'s exposure | This plugin's defense |
|---|---|---|---|
| 1 | Forged/impersonated cards **in gossip propagation** (not just at registration) | Merges any card on sight -- no signing, no verification, anywhere | Signs `(content, version, tombstone)` at write time; re-verifies on **every** `OP_PUSH` hop before merge, dropping unsigned/impersonating/forged/replayed-under-a-forged-tag cards (`rejections` ledger, reason-coded) |
| 2 | **Signed equivocation** -- a publisher validly signs two different cards at the same version, possibly delivered to disjoint peer groups with no common recipient | Last-writer-wins keeps whichever conflicting card arrives first at each node, silently and permanently, no ledger anywhere | Witness map over `(publisher, version)` -> content hash; a second, verified, content-differing write proves the publisher is byzantine, quarantines it permanently, evicts it from the view, records `(publisher, version)` in `equivocations`. The self-verifying **proof** (the two conflicting signed writes) is then gossiped as its own `OP_EQPROOF` item, surviving card eviction and re-verified on receipt, so detection reaches **every** honest node under disjoint multi-equivocator delivery -- not just the direct witnesses, and not defeated by eviction-halts-relay stranding |
| 3 | **Eclipse** -- an honest agent fed only byzantine/rejected data | Pure-uniform peer sampling every round, no memory -- a large-enough byzantine fraction or unlucky draw can exclude a victim's only honest peer indefinitely | Every round draws a deterministic **anchor set** (lexicographically-first half of peers by `AgentId`) plus a seeded-random remainder, so one fixed, stable contact is retried every round instead of only when the dice land right |

## Calibrated claim -- read this before merging

**Byzantine-resistant and attack-detecting, for these three named attacks,
up to a bounded byzantine fraction, under the topologies actually
exercised.** This is explicitly **not BFT** (no consensus, no quorum, no
`f < n/3` proof) and **not unconditionally secure**. In particular:

- **Eclipse resistance is heuristic (an anchor set), not a proof.** The
  anchor half is positional (lexicographically-first `AgentId`s among a
  victim's current peers), not identity-vetted. An adversary that controls
  every anchor slot for a specific victim defeats the guarantee for that
  victim entirely, and there is no rotation mechanism if an anchor slot
  itself turns out to be byzantine. The eclipse scenario's gossip-FAIL is
  **empirically tuned to seeds 42/7/1337** (found by direct simulation,
  not a closed-form bound) -- a seed outside that set is not guaranteed to
  reproduce it.
- **`InertByzantineDriverAgent` (the eclipse scenario's adversary) models
  Sybil dilution by silence, not an active-lying byzantine agent.** It
  never registers, gossips, or answers. A byzantine agent that actively
  lies about its digest is a stronger, different attack, out of scope for
  the three scenarios shipped here.
- **Quarantine is permanent, by design, with no rehabilitation path.** A
  publisher caught equivocating once has every later card refused forever
  on this registry instance -- including a genuinely honest one signed
  after the fact. No re-trust mechanism exists; a real deployment wanting
  one would need to build it externally.
- **The equivocation *proof* is gossiped; a "who-quarantined-whom" opinion
  is not.** What propagates mesh-wide is the self-verifying proof (the two
  conflicting signed writes), which every node re-verifies independently
  before acting -- not a node's unverifiable assertion "I quarantined E,"
  which would be a framing vector. Convergence is therefore **eventual, not
  instantaneous**: while a proof is still in flight there is a transient
  window during which some honest nodes have quarantined the equivocator and
  others have not yet received the proof. That window closes as propagation
  completes, and convergence to "quarantined and evicted everywhere" then
  holds mesh-wide even under disjoint multi-equivocator delivery -- **as long
  as the honest sub-network is connected** (a partition, or a full eclipse of
  a victim, delays the proof until it heals, exactly like any other gossip
  data). This is the one residual honest caveat on the mesh-wide claim; see
  the disjoint-delivery fix above and `VERIFICATION.md` limitation 4.
- **Validators are trace/snapshot-evidence-bounded**, like every NANDA Town
  validator -- they judge the evidence supplied to them (view snapshots,
  cards, ledgers), not run ground truth. `check_no_forged_card_in_view`
  treats "cannot be checked" as a FAIL rather than a silent pass for
  exactly this reason.
- A quarantine reason-code precision nit: a forged card arriving under an
  *already*-quarantined publisher id is logged as `quarantined`, not
  `bad_signature` (quarantine short-circuits before verification runs) --
  harmless, the card is rejected either way, only the reason code is
  imprecise in that one case.

Full list, with the reasoning behind each one: [`VERIFICATION.md`](nest_plugins_reference/registry/VERIFICATION.md#honest-limitations-every-one-gathered-across-the-build).

## FAIL/PASS matrix

Rows = the 3 mandated validators
([`registry_byzantine_validators.py`](nest_plugins_reference/validators/registry_byzantine_validators.py)).
Columns = `{in_memory, gossip, byzantine_gossip}` x `{forgery,
signed_equivocation, eclipse}`. Captured from an actual scenario run at seed
42, cross-checked against the full 30-test gate at seeds 7 and 1337 (all
identical).

| Validator                          | in_memory (all 3) | gossip / forgery | gossip / equivocation | gossip / eclipse | byzantine_gossip / forgery | byzantine_gossip / equivocation | byzantine_gossip / eclipse |
|-------------------------------------|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| `check_no_forged_card_in_view`      | N/A | **FAIL** | **FAIL**\* | **FAIL**\* | PASS | PASS | PASS |
| `check_no_equivocation_accepted`    | N/A | PASS | **FAIL** | PASS | PASS | PASS | PASS |
| `check_no_eclipse`                  | N/A | PASS | PASS | **FAIL** | PASS | PASS | PASS |

\* Structural, not attack-specific in those two scenarios -- `gossip` never
signs anything at all, including honest agents' own registrations. The
attack-specific proof (do the forger's phantom ids actually land in / get
rejected from honest views) is asserted directly by
`test_reference_gossip_accepts_phantom_cards` /
`test_byzantine_gossip_rejects_phantom_cards`, not by this validator's
verdict alone.

`in_memory` is marked **N/A, not PASS**: it cannot be substituted into any
of these scenarios at all (`TypeError: InMemoryRegistry.__init__() takes 1
positional argument but 3 were given` -- it is a single globally-shared
dict with no per-agent view, no gossip mechanics, no equivocation ledger).
The three validators are written against the distributed-view model
`gossip`/`byzantine_gossip` share; `in_memory` has no comparable
adversarial surface to run them against. See `VERIFICATION.md` for the full
argument -- this is not a loophole, `in_memory` genuinely has no attack
surface these validators are testing for, and no defense against it either.

### Reproduce it

```bash
uv run pytest packages/nest-plugins-reference/tests/test_byzantine_gossip_scenario.py -v
uv run pytest packages/nest-plugins-reference/tests/test_registry_byzantine_validators.py -v
```

See `VERIFICATION.md` for the exact `uv run python` snippet that prints the
literal verdict dict per scenario/plugin pair.

## What's in this PR

- `nest_plugins_reference/registry/byzantine_gossip.py` --
  `ByzantineGossipRegistry` (signed cards, forged/impersonation rejection,
  signed-equivocation detection + permanent quarantine, eclipse-resistant
  anchor+random peer sampling). Registered in `nest_core.plugins._BUILTINS`
  and `packages/nest-plugins-reference/pyproject.toml`'s
  `nest.plugins.registry` entry points.
- `nest_plugins_reference/validators/registry_byzantine_validators.py` --
  `check_no_forged_card_in_view`, `check_no_equivocation_accepted`,
  `check_no_eclipse`.
- `scenarios/gossip_byzantine_forgery.yaml`,
  `scenarios/gossip_signed_equivocation.yaml`,
  `scenarios/gossip_eclipse.yaml` -- deterministic under seeds 42, 7, 1337.
- Tests: `test_byzantine_gossip.py` (unit), `test_byzantine_gossip_properties.py`
  (30 Hypothesis property tests + fraction sweep + edge cases),
  `test_registry_byzantine_validators.py` (validator unit tests),
  `test_byzantine_gossip_scenario.py` (30-test end-to-end FAIL/PASS gate +
  determinism).
- `docs/layers/registry.md` -- byzantine_gossip subsection.
- `nest_plugins_reference/registry/VERIFICATION.md` -- full evidence +
  every honest limitation.

## CI

`make ci-local` -- all five checks green (`uv sync`, `ruff check`, `ruff
format --check`, `pyright`, `pytest -v`).

## Branch status

`hackathon/bogacsmz-registry-byzantine-gossip`. **No PR opened** -- handing
the branch back for review with this draft, the matrix, and
`VERIFICATION.md`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ data/hackathon-runs/*.jsonl
 data/hackathon-runs/*.png
 # Per-agent isolated worktrees / clones.
 .harness-work/
+.superpowers/
--- a/docs/layers/registry.md
+++ b/docs/layers/registry.md
@@ -21,6 +21,102 @@ Full definition: [`nest_core/layers/registry.py`](../../packages/nest-core/nest_
 
 Source: [`nest_plugins_reference/registry/in_memory.py`](../../packages/nest-plugins-reference/nest_plugins_reference/registry/in_memory.py).
 
+## Byzantine-resistant plugin: `byzantine_gossip`
+
+`byzantine_gossip` -- a hardened counterpart to the merged `gossip` plugin
+(eventually-consistent, partition-honest discovery). `gossip` assumes every
+participant is honest: no forged cards, no publisher signing two conflicting
+writes at the same version, no coordinated attempt to starve a victim's
+random peer sampling. `byzantine_gossip` drops that assumption on three
+specific fronts:
+
+1. **Signed cards, re-verified on every gossip hop, not just at
+   registration.** Every `register`/`deregister` signs
+   `(content, version, tombstone)` with the agent's `Identity`; every
+   `handle_gossip` `OP_PUSH` re-verifies that signature before merging, so an
+   unsigned, impersonating, forged, or replayed-under-a-different-version/
+   tombstone card is dropped and logged in `rejections`, never applied. This
+   extends prior art `#67` (registration-only signing): `#67` checks a card
+   once, at its source: nothing downstream re-checks it as it hops through
+   the mesh, so a compromised relay can still poison every honest view it
+   touches.
+2. **Signed-equivocation detection + permanent quarantine.** A byzantine
+   publisher can validly sign two *different* cards at the same version --
+   both pass every signature check in isolation, so `#67`-style
+   registration-signing cannot catch it. `byzantine_gossip` witnesses every
+   verified write against `(publisher, version)`; a second, verified,
+   content-differing write at the same key proves the publisher itself is
+   byzantine. It is quarantined on the spot -- evicted from the local view,
+   recorded in `equivocations`, and every later card from it (honest-looking
+   or not) is refused permanently, on this registry instance, with no
+   re-trust mechanism.
+3. **Eclipse-resistant peer sampling.** `gossip`'s `gossip_round` draws
+   fanout peers uniformly at random every round, with no memory between
+   rounds -- a large-enough byzantine peer fraction (or an unlucky draw) can
+   exclude a victim's only honest peer indefinitely. `byzantine_gossip`
+   splits each round's draw into a deterministic **anchor set** (the
+   lexicographically-first half of the peer list by `AgentId`) plus a
+   seeded-random remainder, so a fixed, stable contact is retried every
+   round instead of only when the dice happen to land right. This is a
+   **heuristic, not a proof** -- see
+   [`VERIFICATION.md`](../../packages/nest-plugins-reference/nest_plugins_reference/registry/VERIFICATION.md)
+   for the topology it cannot defend.
+
+Source:
+[`nest_plugins_reference/registry/byzantine_gossip.py`](../../packages/nest-plugins-reference/nest_plugins_reference/registry/byzantine_gossip.py).
+
+### Validators
+
+[`validators/registry_byzantine_validators.py`](../../packages/nest-plugins-reference/nest_plugins_reference/validators/registry_byzantine_validators.py)
+ships three adversarial checks, each FAILing against the reference
+`gossip`/`in_memory` plugins and PASSing against `byzantine_gossip`:
+
+- `check_no_forged_card_in_view` -- every card in an honest view must carry
+  a signature that verifies under its claimed publisher. An entry that
+  cannot be checked at all (missing card or identity) is reported as
+  `unverifiable` and counted as a FAIL, never a silent pass.
+- `check_no_equivocation_accepted` -- whenever two honest agents' views
+  disagree on the content behind the same `(publisher, version)` key, at
+  least one agent's `equivocations` ledger must record it.
+- `check_no_eclipse` -- every honest agent's view must hold at least one
+  live card from another honest publisher (the weaker "reached one," not
+  "reached every one," bar -- matching what the anchor heuristic actually
+  guarantees).
+
+### Demo scenarios
+
+Three scenarios under [`scenarios/`](../../scenarios/), deterministic under
+seeds 42, 7, 1337:
+
+- `gossip_byzantine_forgery.yaml` -- 16 honest agents + 4 forgers injecting
+  unsigned/impersonated/forged phantom cards.
+- `gossip_signed_equivocation.yaml` -- **the novelty proof**: one publisher
+  genuinely signs two conflicting cards at the same version and delivers
+  them to two honest groups in opposite order.
+- `gossip_eclipse.yaml` -- 2 honest agents drowned in 24 inert byzantine
+  "black hole" peers.
+
+```bash
+uv sync
+
+# Full byzantine_gossip test suite (unit + properties + scenario gate):
+uv run pytest packages/nest-plugins-reference/tests/test_byzantine_gossip.py \
+              packages/nest-plugins-reference/tests/test_byzantine_gossip_properties.py \
+              packages/nest-plugins-reference/tests/test_registry_byzantine_validators.py \
+              packages/nest-plugins-reference/tests/test_byzantine_gossip_scenario.py -v
+
+# Run one scenario directly:
+uv run nest run scenarios/gossip_signed_equivocation.yaml
+
+# The whole CI gate:
+make ci-local
+```
+
+See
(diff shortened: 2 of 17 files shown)


bogacsmz:
Hi maintainers — this is my Phase 1 submission (PR #148). The CI checks are still pending because fork PRs need a maintainer to approve the workflow run. Locally, make ci-local is fully green: 1070 passed, pyright 0 errors, ruff clean. Could you approve the workflow run when you get a chance? I'd like it green before judging. Thanks!

shilpa-kulkarni-14:
LGTM

shilpa-kulkarni-14 (approved):

PR #149, opened 2026-07-10
[Hackathon] libre: attested-peering trust plugin

libr3andr3:
# [Hackathon] libre: attested-peering trust plugin

**Persona:** `libre` — a self-sovereignty / decentralized-infrastructure engineer
(maintainer of LibreSynergy, a self-hostable community mesh). This persona cares
about *who you are federating with* before any data crosses the wire: friend or
foe, hardware you can trust, an operator who stands behind the agent.

## Motivation

The two bundled trust plugins answer *"how well has this agent behaved?"*
(`score_average` averages feedback; `agent_receipts` corroborates it with
receipts). Neither answers the **prior** question: *is the thing reporting
feedback even who it claims to be, and should its evidence count at all?* Under
the default `score_average` plugin any agent — including a freshly minted Sybil —
can file unlimited `report()` calls and poison a victim's reputation.

This PR adds an **`AttestedPeering` trust layer** that gates evidence admission
behind a mutual, replay-proof **identity + hardware-root attestation** handshake.
It is a **novel composition of two layers** — Identity gating Trust — ported from
LibreSynergy's production mesh-federation gate.

## Design

A three-question handshake (`hail` → `vouch` → `seal`), each question a distinct
attestation:

1. **Friend or foe?** — the peer proves possession of the private key for the
   identity it presents by signing *this session's transcript* (never a
   replayable bare nonce), and its `AgentFactsCard` must be authentic
   (self-signature + operator delegation valid).
2. **Can I trust you with my data?** — an optional, nonce-bound signed
   environment quote over the peer's measured configuration (a deterministic
   stand-in for a TPM measured-boot quote).
3. **Who do you work for?** — a named operator must have delegated authority to
   the agent via a signed delegation in its passport, and that operator must be
   on the verifier's trusted-operator roster.

Only peers whose verdict is `ALLOW` may contribute evidence via `report()`;
everything else is quarantined and surfaced in the trace.

### Tradeoffs

- **Ed25519 over openssl subprocess.** LibreSynergy's original uses an `openssl`
  subprocess; that violates the Tier-1 determinism rule (no `subprocess`, no
  `os.urandom`, no wall-clock). Ported to in-process `cryptography`: seeds are
  `sha256(seed || agent_id)[:32]` (matching `identity.ed25519_rotating`) and
  nonces come from a per-instance seeded `random.Random`. Ed25519 itself is
  deterministic by construction (RFC 8032), and the module reads no clock at
  all, so a Tier-1 run is byte-identical under a fixed seed (asserted by
  `test_scenario_is_byte_deterministic`).
- **Transcript-bound proof, not bare-nonce.** Signing the session transcript
  closes the replay window a bare-nonce challenge leaves open.
- **Attestation #2 is optional** so the plugin degrades gracefully to identity-
  only gating where no measured quote is available.

## API fit

- Implements the `Trust` layer Protocol; registered under
  `[project.entry-points."nest.plugins.trust"]` as `attested_peering`.
- `from __future__ import annotations`, SPDX header, docstrings with `Example::`
  blocks on public classes and methods.
- Ships a scenario (`scenarios/attested_peering.yaml`) and a built-in scenario
  module — slots into the runner with no core edits beyond one registration line
  and the scenario table.

## How to verify

```
uv sync
uv run pytest packages/nest-plugins-reference/tests/test_attested_peering.py \
              packages/nest-core/tests/test_validators.py -q   # 145 passed
uv run nest run scenarios/attested_peering.yaml                 # ALLOW/quarantine trace
make ci-local                                                  # full gate: ruff + format + pyright + 823 pass
```

## Tests (145 passing)

Adversarial by construction — the suite exercises the failure modes the feature
introduces, not just the happy path:

- **Stolen passport, wrong key** → rejected (key-possession proof fails).
- **Replayed proof from another session** → rejected (transcript binding).
- **Off-roster operator** → rejected (delegation valid but operator untrusted).
- **Tampered environment quote** → rejected (nonce/measurement mismatch).
- **Sybil with no delegation** → quarantined, contributes zero evidence.
- **Malformed / type-confused wire frames** (undecodable base64 signature, bad
  hex nonce, non-dict `facts`, truncated passport) → a clean `DENY`, never an
  unhandled exception in the verifier (parametrized byzantine-input tests).
- Determinism: identical bytes under a fixed seed across runs.
- `validators.py`: adversarial validators that count quarantined peers from the
  trace so a scenario fails loudly if an unattested peer's evidence leaks in.

## Files

- `packages/nest-plugins-reference/.../trust/attested_peering.py` — plugin (941 LoC)
- `packages/nest-core/.../scenarios_builtin/attested_peering.py` — scenario
- `packages/nest-core/nest_core/validators.py` — attestation validators
- `scenarios/attested_peering.yaml` — scenario config
- tests: `test_attested_peering.py` (plugin) + `test_validators.py`

Zero new pip dependencies beyond `cryptography` (already a transitive dep). No
`os.urandom` / `time.time` / `subprocess` / clock reads anywhere — a Tier-1 run
is byte-identical under a fixed seed.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -31,6 +31,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("trust", "parc"): f"{_REF}.trust.parc:ParcTrust",
     ("trust", "aae_permit_gate"): f"{_REF}.trust.aae_permit_gate:AAEPermitGate",
     ("trust", "bonded_trust"): f"{_REF}.trust.bonded_trust:BondedTrust",
+    ("trust", "attested_peering"): f"{_REF}.trust.attested_peering:AttestedPeeringTrust",
     ("payments", "prepaid_credits"): f"{_REF}.payments.prepaid_credits:PrepaidCredits",
     ("payments", "streaming"): f"{_REF}.payments.streaming:StreamingPayments",
     ("payments", "empic_escrow"): f"{_REF}.payments.empic_escrow:EMPICEscrowPayments",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -77,6 +77,12 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("identity_rotation", identity_rotation_factory)
+    elif name == "attested_peering":
+        from nest_core.scenarios_builtin.attested_peering import (
+            attested_peering_factory,
+        )
+
+        register_scenario("attested_peering", attested_peering_factory)
     elif name == "gossip_registry":
         from nest_core.scenarios_builtin.gossip_registry import gossip_registry_factory
 
--- a/packages/nest-core/nest_core/scenarios_builtin/attested_peering.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/attested_peering.py
@@ -1,0 +1,434 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Attested-peering trust scenario — a Sybil swarm tries to defame one victim.
+
+A single ``observer`` maintains the reputation of one honest ``victim``.
+Reporters file evidence about the victim; the observer admits a report only
+if the reporter first clears the attested-peering handshake
+(:mod:`nest_plugins_reference.trust.attested_peering`). Four reporter roles
+stress the gate:
+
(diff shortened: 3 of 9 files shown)


shilpa-kulkarni-14 (approved):
Non-blocking items :

Strip the Claude co-author footer.
Add a docs/layers/trust.md entry.
Glance at the 985-line plugin for any trimmable surface, since it is the largest in the batch, though it passes all gates.

PR #150, opened 2026-07-10
docs(trust): document attested_peering; drop dead attested_peers accessor

libr3andr3:
Follow-up polish to #149 (merged, thank you!). Two small, non-blocking items that
landed just after the merge:

### 1. `docs/layers/trust.md` — document the plugin
Adds an `## attested_peering` section alongside the other trust plugins
(`score_average`, `agent_receipts`, `parc`, `aae_permit_gate`, `bonded_trust`),
matching the house style: what it does, the three-question handshake, the
determinism story, a runnable quickstart, and the `attested_sybil_quarantined`
scenario/validator link. Every other plugin in the layer has an entry; this fills
the gap.

### 2. Remove a dead accessor
`AttestedPeeringTrust.attested_peers` (a `frozenset` property) had **zero**
references across the repo — `is_attested(agent)` covers the same query and is the
only accessor the tests/validators use. Trims the largest plugin in the batch by
10 lines with no behavior change.

### Verification
```
make ci-local   # ruff + format + pyright + pytest: 1037 passed
```

No source-behavior change; the plugin, scenario, validators, and all 145
plugin/validator tests are untouched.

### DIFF
--- a/docs/layers/trust.md
+++ b/docs/layers/trust.md
@@ -245,6 +245,78 @@ Spend *real* bond, gain real influence — intended ("most at stake, most say"),
 [`sybil_bond.yaml`](../../scenarios/sybil_bond.yaml) · `sybil_bond_*` in
 [`validators.py`](../../packages/nest-core/nest_core/validators.py)
 
+## `attested_peering` — verify a peer *before* its evidence counts
+
+Every other plugin in this layer scores what an agent **did**. `attested_peering`
+asks the prior question — *is the reporter who it claims to be, and should its
+evidence count at all?* — and answers it with a mutual, replay-proof handshake
+before any `report()` is admitted. Under `score_average` a freshly minted Sybil
+can file unlimited reports and poison a victim's reputation; here an unattested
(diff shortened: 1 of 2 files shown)


JamesCarnley:
Thanks @libr3andr3 for the follow-up documentation and accessor cleanup for #149.

Reviewed the full patch at `263b02dcc1afb1c3324a98ebfb72eda8eac335a0` against rebuilt `main@d26ca5a`. Neither `AttestedPeeringTrust`, its `attested_peers` accessor, nor the old `docs/layers/trust.md` plugin catalog exists in current main. Current `layers/trust.py` is a different receipt-reputation reference implementation. Porting this documentation unchanged would describe an unavailable handshake rather than current Town behavior.

Closing as obsolete maintenance, not as merged and not as a claim that current reputation has equivalent admission guarantees. The handshake's separation of identity proof, configuration claims and operator authority remains useful research with #149 and this discussion if that optional protocol is reintroduced. No replacement is needed for the removed accessor.

PR #151, opened 2026-07-10
[Hackathon] distributed-sys-engg: deterministic coordination round ids (ADR-004 fix)

rakshit-tresata:
# Deterministic coordination round ids (ADR-004 fix for `contract_net` + `hotstuff`)

**Persona:** `distributed-sys-engg` — reliability/determinism focus.

Both reference coordination plugins, including the default `contract_net`,
mint their round id with `uuid.uuid4()`. That draws from `os.urandom`, so it
is unseedable. This breaks ADR-004 (seeded-determinism), Nanda Town's rule
that a seeded run must replay to a byte-identical trace.

## Motivation

`ADR-004-seeded-determinism` requires that the same seed replay to an
identical trace, so operators can diff two runs or replay a failure. Every
scenario and replay tool in this repo depends on that guarantee holding.

Two reference coordination plugins break it:

- `coordination/contract_net.py` — the default coordination plugin.
- `coordination/hotstuff.py`.

Every call to `propose()` returns a `Round` (and downstream `Bid` / `Outcome`,
which copy `round.id`) with a fresh random id from `uuid4()`. Two runs of the
same seed produce different coordination ids. The plugin's public output is
not reproducible.

Stated plainly: this is a latent bug today. None of the shipped scenarios
(marketplace, auction, voting, consensus) serialize the round id into their
trace, which is why it's gone unnoticed so far. But `propose()` is still a
public API with nondeterministic output, and any scenario or validator that
starts using round ids directly (e.g. the fair-ordering or sealed-bid
coordination work already in flight) will silently inherit a trace that can't
be diffed or replayed.

## The change

Add `coordination/_ids.py::derive_round_id(agent_id, task_id, seq)` and use it
in both plugins. The id is:

- Deterministic: a pure function of `(proposing agent, task, monotonic
  per-proposer sequence)`. A seeded run replays byte-for-byte.
- Injective on its inputs: the pre-image is a length-prefixed (netstring-style)
  encoding, so distinct triples can't alias onto one digest through boundary
  ambiguity (`("ab","c")` vs `("a","bc")`).
- Free of ambient state: no `uuid4`, no wall clock, no object identity, no
  dict-ordering dependence.

Each plugin keeps a private `_round_seq` counter. The sim drives `propose()`
in a seeded, deterministic order, so the counter, and thus the id, reproduces.

Diff: `+177 / -5` across `_ids.py`, `contract_net.py`, `hotstuff.py`, and a new
test file. No code outside the coordination layer is touched.

## Tests — fail without the change, pass with it

`packages/nest-plugins-reference/tests/test_coordination_determinism.py`:

- Replay identity: proposing the same task sequence twice yields
  byte-identical ids. This is the check that fails against `uuid4`.
- Uniqueness: successive rounds and distinct proposers still get distinct ids
  (no functional regression).
- Derivation properties: purity, no ambient state, injective encoding
  (`("ab","c") != ("a","bc")`), and a Hypothesis property over arbitrary
  inputs.

Reverting just the two-line `propose()` change (restoring `uuid4`) makes
`TestReplayDeterminism` fail with, e.g., `round-8c683f74d9f1207a` (expected)
vs a random `f36c1ab2-6955-462d-92b9-9aece726f3b6`.

## Verification

```bash
make ci-local          # uv sync, ruff, ruff format --check, pyright, pytest -v
# or just this PR's test:
uv run pytest -q packages/nest-plugins-reference/tests/test_coordination_determinism.py
```

Local run: `ruff` clean, `ruff format` clean, full suite **954 passed, 0
failed** (this branch), and the new test fails on the pre-change code as above.

## Out of scope / follow-up

`negotiation/alternating_offers.py` has the identical one-line `uuid4` offer-id
bug. It lives in a different layer, so I've left it out of this coordination-
scoped PR; happy to send a sibling PR for the negotiation layer.

### DIFF
--- a/packages/nest-plugins-reference/nest_plugins_reference/coordination/_ids.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/coordination/_ids.py
@@ -1,0 +1,63 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Deterministic coordination round identifiers.
+
+Nanda Town's core guarantee is a *byte-deterministic* JSONL trace: the same
+seed must replay to an identical trace so operators can ``diff`` two runs and
+``replay`` a failure. ADR-004 (seeded-determinism) makes that a hard rule and
+calls out unseeded RNG as a violation of it.
+
+The reference coordination plugins previously minted round identifiers with
+:func:`uuid.uuid4`, which draws from ``os.urandom`` and is therefore
+**unseedable**: every run produced fresh ``Round`` / ``Bid`` / ``Outcome`` ids,
+so two runs of the *same* seed diverged on every coordination event and their
+traces could neither be diffed nor replayed.
+
+:func:`derive_round_id` replaces that with a pure function of the proposing
+agent, the task, and a monotonic per-proposer sequence number. It is:
+
+- **deterministic** -- identical inputs always yield the identical id, so a
+  seeded run replays byte-for-byte;
+- **injective on its inputs** -- distinct ``(agent, task, seq)`` triples map to
+  distinct ids, because the pre-image is a length-prefixed encoding that no two
+  distinct triples can share (this rules out concatenation ambiguity such as
+  ``("ab", "c")`` vs ``("a", "bc")``);
+- **free of ambient state** -- no reliance on object identity, dict insertion
+  order, wall-clock time, or a global RNG.
+
+Example::
+
+    from nest_core.types import AgentId
+
+    rid = derive_round_id(AgentId("manager"), "task-7", 1)
+    assert rid == derive_round_id(AgentId("manager"), "task-7", 1)
+"""
+
+from __future__ import annotations
+
+import hashlib
+
+from nest_core.types import AgentId
+
+
+def derive_round_id(agent_id: AgentId, task_id: str, seq: int) -> str:
+    """Return a deterministic, collision-resistant coordination round id.
+
+    The id is ``"round-" + sha256(preimage)[:16]`` where ``preimage`` is a
+    length-prefixed (netstring-style) encoding of ``(agent_id, task_id, seq)``.
+    Length-prefixing makes the encoding injective, so distinct inputs cannot
+    alias onto the same digest through boundary ambiguity.
+
+    Args:
+        agent_id: The proposing agent. Distinguishes rounds proposed by
+            different agents for the same task.
+        task_id: The task being coordinated.
+        seq: A monotonic per-proposer counter. Distinguishes successive rounds
+            proposed by the same agent; the caller owns incrementing it.
+
+    Example::
+
+        rid = derive_round_id(AgentId("r0"), "t1", 1)
+    """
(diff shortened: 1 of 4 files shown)


shilpa-kulkarni-14 (approved):
Verified on latest main. This is a clean determinism fix, approving.

- The bug is real: contract_net (the default) and hotstuff minted round ids with
  uuid4, which is os.urandom-backed and unseedable, so same-seed runs diverged on
  every coordination event (ADR-004 violation).
- The fix is correct: derive_round_id is a pure, length-prefixed function of
  (agent, task, per-proposer seq), no ambient state, injective on its inputs.
- ci-local 5/5 green on a clean merge (1159 passed). The new determinism suite
  passes, and I confirmed a coordination scenario now replays byte-identical across
  seeds 42/7/1337.
- No discriminator is affected: no shipped scenario serialized round ids into its
  trace, so changing the derivation does not make anything vacuous.

This merges as a maintenance fix, not a ranked stack entry (no new capability or
discriminator), same lane as the score_average hardening.

Two things before merge:
1. Please strip the Claude Opus 4.8 co-author footer from the commit.
2. Yes to the sibling PR for negotiation/alternating_offers.py - it has the same
   uuid4 bug, and closing both now prevents any in-flight scenario from inheriting a
   non-replayable trace. Send it scoped to the negotiation layer the same way.

Nice catch on a latent bug none of the current scenarios surfaced yet.

rakshit-tresata:
Both done

1) Footer stripped, force-pushed the coordination commit; #151 head is now edba629
2) Sibling PR up: #184 negotiation/alternating_offers.py had the same uuid4 session-id bug. Fixed it the same way: derive_session_id(initiator, partner, per-initiator seq), length-prefixed sha256, mirrors coordination/_ids.py. Added test_negotiation_determinism (replay-identity + uniqueness + property-based); the surrounding negotiation/market suites pass unchanged, no regressions. Scoped to the negotiation layer only.

Agreed on the maintenance-fix framing, thanks for the thorough review.

JamesCarnley:
Thanks @rakshit-tresata for identifying the unseeded coordination-ID problem and supplying replay/uniqueness regressions.

Reviewed head `edba629d38d32067db1d6af09152470a9da44324` against rebuilt `main@d26ca5a`. The old `ContractNet.propose` and `HotStuff.propose` implementations are retired. Current `layers/coordination.py` accepts the scenario's explicit task ID, while communication IDs use run-local counters. I also ran the real bundled auction and supply-chain scenarios twice each: both produced nonempty announcement/award traces that matched after normalizing only the outer run identity. The existing layer/simulator/receipt subset passed all 32 tests.

Closing as superseded, without porting the old helper or claiming the original PR was merged. The retained requirement is stable within-run identifiers and reproducible logical traces; distinct run IDs and live security/session identities should remain distinct. If auto-generated coordination rounds return, replay and proposer/sequence uniqueness tests should accompany them. The deterministic-ID design and original regression work remain available here.

PR #152, opened 2026-07-10
[Hackathon] Redeem_Grimm: authenticate failure-detector heartbeats against a forgery attack

Redeem-Grimm-Satoshi:
## Problem

This is a direct follow-up to the failure-detector layer merged in #46. In the review on merge, @Skyrider3  flagged two gaps, and both are worth fixing:

1. Heartbeat identity was spoofable by design. The observer read the peer id out of the heartbeat payload (`FDHB|<id>`) rather than authenticating it. A Byzantine agent could broadcast `FDHB|target-0` while the real `target-0` was down and keep it looking alive indefinitely. The base scenario ran with zero Byzantine agents, so this never fired, but the whole point of the framework is to prove a protocol holds up under attack, and the shipped accuracy validator only discriminated a naive implementation (the fixed-timeout foil), not an attacker.
2. The accuracy validator hardcoded `_MAX_PLAUSIBLE_GAP = 22.0`, which is the scenario's `hb_max` (20) plus a margin. Re-running the scenario with a different heartbeat bound would silently invalidate the threshold.

This PR closes both, and it does so with the same discriminator discipline the original used: a validator that fails against the weak implementation and passes against the hardened one.

## Persona

The keep-alive attack is the failure a real detector has to survive, so authentication belongs in the heartbeat path, not in an appendix. The design question is where to put it. A failure detector should stay a pure liveness oracle over observations it can trust, so I put verification in the observing agent and left the `FailureDetector` contract untouched. That mirrors how the rest of the stack works: the identity layer is already right there, and the gossip and HotStuff scenarios already sign their writes with it, so a heartbeat is just one more signed message. I used the existing `did_key` identity plugin rather than inventing anything.

## What is in the diff

| Path | Purpose |
|---|---|
| `.../scenarios_builtin/failure_detection.py` | Heartbeats are now signed and timestamped: `FDHB|<id>|<ts>|<sig-hex>`. New public helpers `heartbeat_payload`, `claimed_peer`, and `verify_heartbeat` form the wire-and-verify surface. `verify_heartbeat` rejects anything that is not well formed, not signed by the claimed peer's key, from the future, or not strictly newer than the last accepted beat from that peer. A new `ByzantineForgerAgent` mounts the attack (fabricated beats signed with its own key, plus byte-exact replays of captured genuine beats). The observer runs in verifying mode by default and a trusting mode as the foil. The observer also emits one `fd:config` marker so the validators can read the scenario's heartbeat bound from the trace. |
| `scenarios/failure_detection_forgery.yaml` | The forgery scenario: observer, target, always-live peer, and one forger impersonating the target during its outage. Task type `failure_detection_forgery`, four agents, zero message drop, deterministic. |
| `packages/nest-core/nest_core/validators.py` | New `validate_failure_detection_no_forged_liveness`: for every outage segment during which forged beats arrived (a received `FDHB|` whose claimed id differs from the transport sender), the victim must still be suspected at the last in-segment status update. Plus `_fd_max_plausible_gap`, which derives the accuracy bound from the `fd:config` marker (`hb_max + margin`) with the old constant kept only as a fallback for pre-marker traces. |
| `packages/nest-core/nest_core/scenarios.py` | Registers the `failure_detection_forgery` task type (same factory, different roles). |
| `packages/nest-plugins-reference/tests/test_failure_detection.py` | Twelve new tests: signed-heartbeat unit tests (authentic accepted; fabricated, replayed, future-dated, and malformed rejected), the three-seed forgery integration pass, the trusting-vs-verifying discriminator, the config-derived accuracy bound, and forgery-scenario determinism. |
| `packages/nest-core/tests/test_validators.py` | Adds `failure_detection_forgery` to the registry-completeness assertion. |

The base `failure_detection` scenario keeps its behavior: it now signs heartbeats too, and still passes completeness, accuracy, and recovery on every seed.

## Why this matters

The original layer proved an accrual detector beats a naive timeout, which is a class of implementation mistake. This adds the property the reviewer actually asked for: the layer now proves it survives a class of attack. The forgery scenario is a real test, not a demo. With verification on, the observer rejects every forged and replayed beat and the outage is detected; with verification off, the forged liveness masks the crash and the peer is never suspected. The `failure_detection_no_forged_liveness` validator fails the trusting observer and passes the verifying one, so it discriminates by construction, the same way the accuracy validator discriminates the fixed timeout.

It also composes cleanly with a second layer. The heartbeat authentication is the identity layer doing exactly what it exists for, with no new crypto and no change to the `FailureDetector` contract.

## Verification

`make ci-local` runs the same five gates as CI and passes:

```
ruff format --check: 211 files already formatted
pytest:              1165 passed, 1 skipped, 1 deselected
```

Running the forgery scenario with verification on (the default), every validator passes:

```
uv run nest run scenarios/failure_detection_forgery.yaml

PASS failure_detection_completeness       all 1 outage segment(s) detected and still suspected at end
PASS failure_detection_accuracy           no false suspicion of a provably-live peer across 2 peer(s)
PASS failure_detection_recovery           all 1 recovered peer(s) cleared by end
PASS failure_detection_no_forged_liveness forged liveness rejected across 1 attacked outage segment(s)
```

The same scenario with `verify_heartbeats: false` is the foil: the forger keeps the dead peer alive, so completeness and `no_forged_liveness` both fail:

```
FAIL failure_detection_completeness       target-0: never suspected during outage [215.6, 331.3]
FAIL failure_detection_no_forged_liveness target-0: forged heartbeats masked the outage [215.6, 331.3]
```

The verifying observer passes on seeds 42, 7, and 1337, asserted in `test_forgery_signed_heartbeats_defeat_the_attack`; the trusting-vs-verifying split is asserted in `test_forgery_discriminator_trusting_observer_is_fooled`. Runs are byte-for-byte deterministic per seed, asserted in `test_forgery_scenario_is_byte_for_byte_deterministic`.

## Reviewer notes worth surfacing

- **Verification is three independent checks.** Signature (defeats fabrication: a forger signing with its own key fails against the victim's public key), strict-newer timestamp (defeats replay: a captured genuine beat is stale once accepted), and a not-from-the-future check. None of them consults the transport sender field, so the guarantee holds even on a transport whose sender cannot be trusted.
- **Timestamp quantization.** The signed timestamp is rounded to six decimals for byte-stable traces. Because the reference transport is zero-latency, a rounded timestamp can land a hair above the observer's exact `now` on the same tick, so the future check compares against `round(now, 6)` rather than raw `now`. Without this, roughly half of genuine beats read as future-dated and get dropped. There is a regression test for the accepted-beat path through the full simulator.
- **The forger has a real key.** It is built with its own `did_key` identity, so its impersonation is the strongest available: a genuine signature by the wrong key, not malformed bytes. That is what makes signature verification, rather than a well-formedness check, the thing that stops it.
- **Config-derived threshold.** The accuracy bound now comes from the `fd:config` marker in the trace, following the same recorded-not-inferred principle the ground-truth `fd:phase` markers already use. `test_accuracy_bound_is_derived_from_config_marker` proves a non-default `hb_max` changes the verdict, which the old hardcoded constant could not have done.
- **No change to existing behavior.** The base `failure_detection` scenario and its validators are unchanged in outcome; the only visible difference in its trace is that heartbeats now carry a signature and a timestamp.
- **Two defense-in-depth guards.** Verification rejects a non-finite (`nan`/`inf`) timestamp outright, since a `nan` would otherwise slip both IEEE-754 freshness comparisons. And the accuracy bound is taken only from an `fd:config` marker emitted by an observer (an agent that also publishes `fd:status`), so a Byzantine emitter cannot broadcast a spoofed `fd:config` to skew the threshold. Neither was reachable by the in-scenario forger, but both are cheap and remove a latent footgun. Each has a dedicated test.

### DIFF
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -159,6 +159,12 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("failure_detection", failure_detection_factory)
+    elif name == "failure_detection_forgery":
+        from nest_core.scenarios_builtin.failure_detection import (
+            failure_detection_factory,
+        )
+
+        register_scenario("failure_detection_forgery", failure_detection_factory)
     elif name == "parc_migration":
         from nest_core.scenarios_builtin.parc_migration import (
             parc_migration_factory,
--- a/packages/nest-core/nest_core/scenarios_builtin/failure_detection.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/failure_detection.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Failure-detection scenario — silence that heals, with a ground-truth oracle.
+"""Failure-detection scenarios — silence that heals, plus a Byzantine forger.
 
-Topology (see ``scenarios/failure_detection.yaml``):
+Topology of the base scenario (see ``scenarios/failure_detection.yaml``):
 
 * ``observer-0`` runs a :class:`~nest_core.layers.failure_detector.FailureDetector`
   instance (injected per-agent via the ``_agent_plugins`` override channel) and
@@ -22,11 +22,35 @@ and stays quiet through the jitter while still catching the genuine outage.
 The accuracy validator in :mod:`nest_core.validators` is tuned to separate the
 two: the baseline fails it, phi-accrual passes.
 
+**Authenticated heartbeats.**  Every heartbeat is signed with the emitter's
+identity plugin and carries its emission timestamp:
+``FDHB|<id>|<ts>|<sig-hex>``.  The observer verifies the signature against the
+claimed peer's registered public key and requires the signed timestamp to be
+strictly newer than the last accepted one from that peer (and not from the
+future), so both fabricated and replayed heartbeats are rejected.  Verification
+lives in the observing agent, not in the detector: the
+:class:`~nest_core.layers.failure_detector.FailureDetector` contract stays a
+pure liveness oracle over *accepted* observations.
+
+The forgery scenario (``scenarios/failure_detection_forgery.yaml``, task type
+``failure_detection_forgery``) adds a Byzantine ``forger`` role that mounts the
+keep-alive attack: while the victim is genuinely down, the forger keeps
+broadcasting heartbeats that *claim* to be the victim — both fabricated
+payloads signed with the forger's own key and byte-exact replays of captured
+genuine beats.  With ``verify_heartbeats: true`` (the default) the observer
+rejects every forgery and the outage is detected; with the trusting mode
+(``verify_heartbeats: false``, the foil) the forged liveness masks the crash
+and the ``failure_detection_no_forged_liveness`` validator fails.  This is the
+same discriminator pattern the base scenario uses against the fixed-timeout
+baseline, applied to an attack class instead of an implementation mistake.
+
 Ground truth is not inferred — the emitters broadcast ``fd:phase`` marker events
-at start and on every reachability transition, so the validators know exactly
-which intervals were truly reachable versus truly down.  Heartbeats and status
-reports are plain broadcasts at ``failures.message_drop == 0``, so no redundancy
-is needed and the run is fully deterministic for a fixed seed.
+at start and on every reachability transition, and the observer broadcasts one
+``fd:config`` marker carrying the scenario's heartbeat bounds, so the
+validators derive their thresholds from the trace instead of hardcoding the
+scenario configuration.  Heartbeats and status reports are plain broadcasts at
+``failures.message_drop == 0``, so no redundancy is needed and the run is fully
+deterministic for a fixed seed.
 
 Example::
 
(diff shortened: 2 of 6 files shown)


shilpa-kulkarni-14 (approved):
Verified on latest main Clean, approving.

- Merges clean, ci-local 5/5 green (1165 passed). The changes to the merged
  failure_detection scenario and validators are additive; the base scenario still
  passes.
- Discriminator is non-vacuous end to end: verify_heartbeats=true scores 4/4;
  verify_heartbeats=false (the foil) drops to 2/4, flipping
  failure_detection_completeness and failure_detection_no_forged_liveness. Without
  auth the forger's fake beats mask the crash; with it, both forgery and outage are
  caught.
- Deterministic across seeds 42/7/1337.
- The freshness window covering replayed beats, and the second commit rejecting
  non-finite values, are both good touches.

One structural note, not blocking: the authenticated-heartbeat logic lives in the
scenario builtin rather than a reusable plugin, and the discriminator is a config
toggle rather than a swapped layer plugin. It's a valid discriminator (verify=false
is the real naive baseline), but if you want this to be composable by other
scenarios, consider extracting the verification into a small reusable component in a
follow-up.

Since this modifies the merged failure_detection scenario, flagging @maria for a
quick ack that the base-scenario rework (+327/-36) is fine before merge.

Nice threat model on the keep-alive attack.

PR #153, opened 2026-07-10
stripe-engineer: streaming payments — idempotency, conservation invariants, and a working end-to-end scenario

stanleyoz:
Supersedes #116 (renamed to `hackathon/stanleyoz-streaming-payments` per the charter's `hackathon/<handle>-<theme>` branch-naming rule; same author, same content plus the fixes requested in that review thread).

## Problem

Problem #03: streaming pay-per-second payments with mid-stream cancellation, idempotency, and conservation invariants (`docs/hackathon/problems/03-payments-streaming-x402.md`).

## Delta over already-merged #21

#21 already merged the base `streaming` payments plugin for this problem. This PR reworks it in place rather than introducing a competing plugin. On top of #21's baseline it adds:

- **Idempotency everywhere**: `open_stream`/`close_stream`/`refund_stream` are all keyed by `PaymentRef` — retrying any of them after it already succeeded returns the original result instead of raising.
- **A refund lifecycle** (`refund_stream`) for recovering the unused remainder of a closed stream, not present in #21.
- **A Hypothesis property-test suite** (`test_streaming_properties.py`, ~18 property tests) proving conservation holds under randomized sequences of open/tick/close/refund operations, not just fixed example tests.
- **7 adversarial validators** (conservation, no-drain-after-close, no-overbill-on-partition, rate-enforcement, no-double-open, per-tick conservation, audit-trail-completeness) versus the 2 the problem spec requires as a minimum.

#93 and #136 are open competing submissions on the same problem; this PR's differentiator is the property-test rigor above and the fix described next.

## What this PR actually fixes (disclosed plainly, not glossed over)

The original submission's scenario (`streaming_payments.py`) never worked: the factory never instantiated the plugin, so `ctx.plugins.get("payments")` resolved to the raw class at runtime, every `open_stream`/`tick_stream` call was silently swallowed by a bare `except Exception`, and nothing ever drove ticks (the code waited for an external tick broadcast that was never sent). The 7 validators separately read raw trace events directly, but real trace events wrap self-sent audit payloads as `{"kind": "send", "msg": "<json>"}` — so the validators never matched anything either. Net effect: the scenario produced an empty trace and all 7 validators vacuously passed, without ever actually exercising the plugin end-to-end.

This PR fixes both halves:
- `streaming_payments.py`: instantiates `StreamingPayments` per agent with a shared ledger via the `_agent_plugins` override channel (mirrors `escrow_marketplace.py`'s pattern), self-schedules ticks via `ctx.schedule`, removes the blanket exception suppression (a narrow `AttributeError` catch remains only where an incompatible baseline plugin genuinely lacks the streaming methods), and fixes the audit trail to emit real per-tick amounts with the seller's credit co-emitted atomically alongside the buyer's debit.
- `validators.py`: adds a `_streaming_audit_events` unwrap helper (mirroring the existing `_empic_audit_events` pattern) so the validators actually read the real trace shape; also fixes a latent bug where the partition-overbill validator read `"tick"` on `dropped` events, which actually carry `"ts"`.
- New `packages/nest-plugins-reference/tests/test_streaming_scenario.py` runs the real scenario end-to-end through the actual `Simulator` and proves all 7 validators pass against genuine, non-trivial multi-agent streaming behavior — the thing the original submission claimed but never actually demonstrated. It also proves the `prepaid_credits` baseline cannot exhibit the protocol at all (the scenario completes but produces zero stream-lifecycle events, which every validator's own detail message honestly reports).

## Scope

Per review, `streampay/` (a separately hosted FastAPI service demoing this plugin over HTTP), `hackathon.md`, and the "Our Submission" section added to the shared `README.md` have all been removed from this PR — those are out of scope for a layer/scenario/validator PR. The hosted service will live in its own repository going forward rather than in this one.

## Verification

`make ci-local` passes clean: `ruff check`, `ruff format --check`, `pyright` (0 errors), and the full `pytest` suite (1191 passed, 1 skipped).


stanleyoz:
Thanks for the detailed review — all six items addressed in #153:

1. **Scope cleanup** — `streampay/`, `hackathon.md`, the README section, and the `pyproject.toml` pyright exclusion are all removed.
2. **Rebase** — done on top of current `main`; the one conflict (an `elif` ordering clash in `scenarios.py`'s registry against `sybil_bond`, merged separately) is resolved by keeping both branches.
3. **Scenario wiring** — the factory now instantiates `StreamingPayments` per agent (buyer *and* seller) via `_agent_plugins`, sharing one ledger, mirroring `escrow_marketplace.py`'s `_instance` pattern. Ticks are self-scheduled upfront via `ctx.schedule` rather than waiting on a broadcast that never came. All blanket `except Exception: pass` are gone — the only remaining catch is a narrow `AttributeError` in `on_start` for the case where the configured plugin genuinely has no streaming methods (see point 5).
4. **Audit trail** — added `_streaming_audit_events`, a direct copy of `_empic_audit_events`, and pointed all 7 validators at it instead of raw trace events. Also fixed a latent bug this surfaced: the partition-overbill validator was reading `"tick"` on `dropped` events, which actually carry `"ts"`. Per-tick debit amounts now come from the actual `StreamEntry` the plugin recorded, and the buyer co-emits the seller's `payment_credited` in the same call so debit/credit land atomically at the same tick (avoids a scheduling race against the per-tick conservation check).
5. **Tests** — `test_streaming_scenario.py` runs the real scenario through the real `Simulator` and confirms all 7 validators pass against genuine multi-agent behavior (this is the part that was never actually true before — the old wiring produced an empty trace that vacuously passed). One honest deviation from the suggested wording: `prepaid_credits` doesn't just fail the validators, it has no `open_stream`/`tick_stream` at all, so the scenario runs to completion but produces **zero** stream events, and every validator — being a "no violation found" check — vacuously *passes* on that empty set, same as escrow's bps-range check passing vacuously when there are no arbitrate events to range-check. I made the test assert that plainly (each validator's own detail message reports "0 streams observed") rather than force a "fails" narrative that wouldn't match what actually happens. Happy to adjust if you'd rather see this expressed differently. Also added direct pass/fail tests for the 4 previously-untested validators (rate-enforcement, no-double-open, per-tick conservation, audit-trail-complete).
6. **`make ci-local`** — clean: ruff check, ruff format --check, pyright (0 errors), full pytest sweep (1191 passed, 1 skipped).

Branch renamed to `hackathon/stanleyoz-streaming-payments` per the charter pattern, opened as #153 (linked from #116, which I've closed). Let me know if the `prepaid_credits` framing in point 5 needs to change.

stanleyoz:
Hey @dhve — following up, it's been about 3.5 weeks since the fixes went up on this PR with no re-review yet. Just re-ran the full CI sequence locally on the current head (`898f344`) in a clean checkout: ruff check, ruff format --check, and pyright are all green (0 errors), and pytest is 1352 passed / 1 skipped. Branch is mergeable with no conflicts against `main`.

To recap what's addressed since your #116 review:
1. `streampay/`, `hackathon.md`, the README section, and the pyright exclude are all removed (fresh checkout confirms none of it ships).
2. Scenario now instantiates `StreamingPayments` per agent via `_agent_plugins` and self-drives ticks via `ctx.schedule`; no more silent `except Exception: pass`.
3. Added `_streaming_audit_events` to unwrap the send-wrapped payloads, fixed debit/credit emission so amounts reflect actual drains, and gave the seller a credit path.
4. Added `test_streaming_scenario.py` proving fails-on-prepaid / passes-on-streaming end-to-end, plus direct tests for the 4 new validators.
5. Rebased onto current `main`, no conflicts.
6. Branch renamed to match the `hackathon/<handle>-<short-theme>` pattern.

Would appreciate a re-review when you get a chance — happy to jump on the hackathon channel if anything needs walking through.

JamesCarnley:
Thanks @stanleyoz for the end-to-end scenario and streaming-payment work. This contribution helped identify an important current-main boundary: an imported legacy scenario can complete a substituted reference exchange without exercising its original implementation.

The selected current-harness fix is now merged in #241. Imported runs explicitly record `original_scenario: not_tested`, disclose every unsupported failure key, and label the report verdict as applying only to the adapted reference flow. The regression tests run actual local streaming/gossip imports through execution, export and replay. Full suite: 248 passed; Python 3.11/3.12 CI passed.

Closing this archive/legacy PR as partially superseded, not as a merge of its original code. The broader streaming-payment implementation has not been ported. Its useful future acceptance criteria remain here: instantiate and tick the actual candidate, consume real trace envelopes, exercise open/settle/close/refund idempotency and conservation, and require lifecycle evidence in negative prepaid cases rather than accepting zero events as a pass. Those need a separately selected streaming-payment profile, not a generic marketplace import.

The original branch, discussion and tests remain available as the source for that future suite. Thank you for the contribution.

PR #155, opened 2026-07-10
Hackathon/parsa noah mesh revocable auth

parsaayoubi:
# auth: `mesh_revocable` — revocation that survives a network partition

## Problem

`delegatable` (#138) solved delegation, but its revocation set is one in-process
`set[str]`, and `delegated_auth` hands **a single plugin instance to all sixteen
agents** — so the cascade is global for free. It's one Python object.

Give each agent its own replica, the only shape a real network permits, and the
guarantee vanishes: a revocation at the issuer is invisible to every other
verifier, forever.

## Solution

`MeshRevocableAuth` subclasses `DelegatableAuth` — token format, HMAC chain,
attenuation, exceptions all inherited verbatim, zero overrides — and reinterprets
the inherited revocation set as a replica of a **grow-only set CRDT**.
`export_revocations()` / `merge_revocations()` are the gossip channel. Union is
commutative, associative and idempotent, so replicas converge under any delivery
order (Hypothesis enforces the laws). Malformed gossip raises
`RevocationStateError` and leaves the replica untouched. 150 lines.

## Evidence

`nest run delegated_auth_partition` — 10 agents, seed 42, byte-identical across
runs. Split at t=10; revoke the far subtree's grant at t=20 while it's
unreachable; heal at t=55; gossip every 5 ticks. Far-side gateway on that lineage:

```
t= 3..51   accept   (split open from t=10 — it cannot know)
t=57..87   DENY     (healed t=55; converged within one gossip round)
```

Same scenario, swapped plugin:

| Wiring | liveness | converges |
|---|---|---|
| `delegatable`, one shared instance | ✗ denies impossibly fast | ✓ vacuously |
| `delegatable`, one replica per agent | ✓ | **✗ stale forever** |
| `mesh_revocable`, one replica per agent | ✓ | ✓ |

That middle row is the gap. An end-to-end test drives the real simulator under
both plugins and locks the flip into CI.

Two validators encode both halves of the CAP position, so neither can rot:
`check_partition_liveness` (an isolated verifier **must** keep accepting — denying
instantly claims knowledge the network never delivered) and
`check_revocation_converges` (after heal + a gossip bound nobody accepts, **and** a
verifier other than the revoker explicitly denied — no vacuous pass).
`check_no_stale_ancestor_use` is deliberately not asserted: unattainable under
partition. The merged escalation and audience validators run green here.

## Supporting core change

Partitions activated at the first event, so no scenario could establish
cross-boundary state before the split. Adds `partition_start_at_time` /
`partition_heal_at_time`, gating on **simulation time** — a gossiping plugin burns
more events per tick, so `t=55` is event 702 under `mesh_revocable` and 339 under
`delegatable`; an event-count window would put the two under *different*
partitions and void the comparison. `partition_heal_at_tick` keeps its meaning and
its test; `None` preserves today's behavior.

## Verification

`make ci-local` green: ruff + format clean, pyright strict **0 errors**, **1192
passed** (`main` collects 1151; +42 here). `nest doctor` 7/7. `nest run
delegated_auth` unchanged. Registered in `_BUILTINS` and as a `nest.plugins.auth`
entry point. Docs: [`docs/layers/mesh_revocable_auth.md`](docs/layers/mesh_revocable_auth.md).

## Limits

- **Revocation window** = partition duration + up to one gossip interval. The CAP
  price, bounded and explicit.
- **Grow-only**, matching the base plugin's no-unrevoke semantics.
- **Gossip is the deployment's job** — the plugin exposes the channel, not a
  transport.
- **Symmetric secret, inherited** — any holder can also mint.
- **Couples to the base plugin's private `_revoked`** — regression tests make an
  upstream refactor fail loudly, not silently.

### DIFF
--- a/docs/layers/mesh_revocable_auth.md
+++ b/docs/layers/mesh_revocable_auth.md
@@ -1,0 +1,124 @@
+# Mesh-Revocable Auth: revocation that survives the network
+
+Plugin: `("auth", "mesh_revocable")` —
+[`nest_plugins_reference/auth/mesh_revocable.py`](../../packages/nest-plugins-reference/nest_plugins_reference/auth/mesh_revocable.py)
+
+## Problem
+
+The delegation problem was solved by the merged `delegatable` plugin: real
+macaroon chains, offline attenuation, audience binding, and cascading
+revocation. But that plugin's revocation knowledge is a single in-process
+`set[str]`, and the `delegated_auth` scenario hands **one plugin instance to
+all sixteen agents**. Under that wiring cascading revocation is trivially
+global — it is one Python object.
+
+Give every agent its own replica (the only shape a real network permits) and
+the guarantee evaporates: a revocation performed at the issuer is invisible
+to every other verifier, forever. Nothing in the plugin propagates it, and
+nothing in the merged scenario notices, because no second verifier exists.
+The layer has correct capability *semantics* and no capability
+*distribution*.
+
+## Solution
+
+`MeshRevocableAuth` supplies the missing half by **composition, not
+re-implementation**. It subclasses `DelegatableAuth` — token format, HMAC
+chain, attenuation rules, and all five exception types inherited verbatim,
+zero overrides of `issue` / `delegate` / `verify` / `verify_presented` /
+`revoke` — and reinterprets the inherited revocation set as one replica of a
+**grow-only set CRDT** (G-Set). Two additive methods are the replication
+channel:
+
+| Method | Contract |
+|---|---|
+| `export_revocations() -> bytes` | Canonical, sorted JSON (`{"crdt": "revocation_gset", "revoked": [...]}`). Identical state exports identical bytes, so traces stay replay-deterministic. |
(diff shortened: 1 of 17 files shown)


shilpa-kulkarni-14 (approved):
@parsaayoubi Verified on latest main . Strong contribution, approving pending one
maintainer ack below.

- Merges clean, ci-local 5/5 green (1207 passed). The time-gated partition simulator
  change is additive and breaks no existing scenario.
- Not a duplicate of the delegatable PRs. It composes the merged delegatable (#138)
  and adds distributed revocation convergence via a G-Set CRDT (export/merge_revocations
  gossiped across replicas). That is a distinct, harder capability than #138's
  single-process cascade.
- Discriminator is non-vacuous and uses #138 as the baseline: the scenario tests
  confirm mesh_revocable achieves both partition liveness and revocation convergence,
  while per-replica delegatable stays live but never converges.
- Deterministic across 42/7/1337, 39 tests pass including Hypothesis properties, and
  the CAP trade is stated honestly (available under partition, consistent within one
  gossip round of healing).

One governance item: this bundles a simulator change (time-gated partitions in
sim/simulator.py). It's additive and necessary for the discriminator, but since it
touches the simulator, flagging @maria for a quick ack before merge. The feature is
reusable, worth keeping as a general partition primitive.

Excellent work, and a good example of composing an existing plugin to fill a real gap
rather than reimplementing it.

PR #156, opened 2026-07-10
hackathon/mwarsss-delegatable-auth

mwarsss:
## Summary

Implements **Problem 04 → Auth Capability Delegation** from the NandaHack charter.  Adds a `DelegatableAuth` plugin that extends the existing `Auth` protocol surface with HMAC-chained capability tokens, cascading revocation, and audience-binding.

- **Plugin** → `DelegatableAuth` (auth layer, name `delegatable`).  Extends the `Auth` protocol (`issue` / `verify` / `revoke`) with an additive `delegate(parent_token, audience, scopes, ttl, *, caller) → Token` method.  No existing callers are modified.
- **Four attacks blocked** → scope escalation, stale-parent revocation, audience confusion at `verify()`, and audience confusion at `delegate()`.  Each attack raises a distinct typed exception (`ScopeEscalationError`, `RevokedAncestorError`, `AudienceConfusionError`), all `ValueError` subclasses for drop-in compatibility.
- **HMAC chain** → based on Birgisson et al. 2014 (Macaroon construction).  Each child token's signing key is derived from the parent's HMAC output, so revoking a parent transitively invalidates every descendant without a per-child revocation list.
- **Audience binding** → both `verify(token, *, caller)` and `delegate(token, audience, scopes, ttl, *, caller)` enforce the audience check unconditionally on audience-bound tokens.  Omitting `caller=` on a delegated token is rejected at both call sites — the same as passing a wrong caller.  A gap where `verify()` silently accepted audience-bound tokens when `caller=` was omitted was found and closed in the same review cycle as the `delegate()` fix.
- **Three adversarial validators** → `check_scope_escalation_blocked`, `check_stale_parent_blocked`, `check_audience_confusion_blocked` in `nest_plugins_reference/validators/auth_validators.py`, returning `ValidatorReport` (same type as `gossip_validators.py`).  Each validator is proven against a JWT baseline that deliberately fails.
- **Scenario** → 16-agent two-level tree: 1 coordinator → 3 intermediaries → 12 leaves.  Scenario YAML at `scenarios/delegated_auth.yaml` with `layers.auth: delegatable`.
- **Wiring** → `_BUILTINS` in `plugins.py`, `_try_load_builtin` in `scenarios.py`, and a `[project.entry-points."nest.plugins.auth"]` entry in `pyproject.toml`.

## What differentiates this from other Problem 04 submissions

Most open PRs on Problem 04 implement a single signature scheme or a flat token list. This submission is distinct on three concrete axes:

1. **Key-chain revocation without a per-child revocation list.** Revoking the root invalidates every descendant transitively because each child's signing key is the HMAC of its parent's signature — there is no registry of child IDs to maintain. The mechanism is O(depth) to verify, O(1) to revoke.

2. **Typed, ValueError-compatible exceptions.** `ScopeEscalationError`, `RevokedAncestorError`, and `AudienceConfusionError` are all `ValueError` subclasses. Callers that catch `ValueError` get correct behaviour for free; callers that want to distinguish attack types can catch the specific subclass.

3. **Validators proven against a JWT baseline.** Each of the three adversarial validators (`check_scope_escalation_blocked`, `check_stale_parent_blocked`, `check_audience_confusion_blocked`) is run against both `DelegatableAuth` (must pass) and the reference `JwtAuth` (must fail). This dual-proof structure means the validator itself is tested, not just asserted.

## Fixes since the original submission (PR #103)

Maintainer review raised four issues; all are resolved in this PR.  A fifth gap was found and closed during the same review cycle:

| # | Issue | Fix |
|---|-------|-----|
| 1 | Startup crash — factory passed plugin class, not instance | `delegated_auth_factory` now instantiates `DelegatableAuth(clock=0.0)` using the same `isinstance(cls, type)` guard as `marketplace.py` and `empic_payments.py` |
| 2 | Non-deterministic token IDs (`uuid4()`) | Replaced with per-instance monotonic counter (`tok-0001`, `tok-0002`, …) |
| 3 | Non-deterministic timestamps (`time.time()`) | Fixed clock `0.0` injected at factory time; `iat`/`exp` are now constant across runs |
| 4 | Audience bypass in `delegate()` | Added unconditional audience check in `delegate()` before calling `verify()`; a thief calling `delegate(stolen, ...)` with no `caller=` kwarg is now rejected the same as one who passes a wrong `caller=`.  Adversarial test confirmed the vulnerability was real (fail-before) and the fix closes it (pass-after) |
| 5 | Audience bypass in `verify()` — same pattern as #4 | `verify()` previously only checked audience when `caller` was not `None`; omitting `caller=` on an audience-bound token silently succeeded.  Fixed with the same unconditional condition: `if audience is not None and (caller is None or caller != audience)`.  `TestVerifyCallerBypass` proves fail-before / pass-after |

## Design choices

- **No types.py modification** → `AuthContext` has no `metadata` field; audience is carried in the token payload itself and checked via `caller=` keyword argument.
- **Constructor mirrors `JwtAuth`** → `__init__(self, secret: bytes, clock: float | None = None)`.
- **No new dependencies** → uses only `hmac`, `hashlib`, `json`, `time` from the standard library.

## Files changed

| File | Change |
|------|--------|
| `packages/nest-plugins-reference/nest_plugins_reference/auth/delegatable.py` | Plugin — deterministic IDs, unconditional audience check in both `delegate()` and `verify()`, updated docstring |
| `packages/nest-plugins-reference/nest_plugins_reference/validators/auth_validators.py` | Three adversarial validators |
| `packages/nest-core/nest_core/scenarios_builtin/delegated_auth.py` | Scenario factory — instantiates auth with `clock=0.0` |
| `packages/nest-core/tests/test_delegated_auth_scenario.py` | E2E test: run-to-completion, 12-leaf count, sha256 determinism |
| `scenarios/delegated_auth.yaml` | Scenario YAML |
| `packages/nest-plugins-reference/tests/test_delegatable_auth.py` | +`TestVerifyCallerBypass`: verify() bypass fail-before/pass-after; +`TestDelegateCallerBypass`: delegate() bypass fail-before/pass-after |
| `packages/nest-plugins-reference/tests/test_delegatable_auth_properties.py` | 5 Hypothesis property tests |
| `packages/nest-core/nest_core/plugins.py` | +1 line: `("auth", "delegatable")` in `_BUILTINS` |
| `packages/nest-core/nest_core/scenarios.py` | +4 lines: `delegated_auth` branch in `_try_load_builtin` |
| `packages/nest-plugins-reference/pyproject.toml` | +3 lines: `[project.entry-points."nest.plugins.auth"]` |

## Test plan

- [x] `uv run ruff check .` — no errors
- [x] `uv run ruff format --check .` — no reformats needed
- [x] `uv run pyright` — 0 errors, 0 warnings
- [x] `uv run pytest -v` — **836 passed, 1 skipped, 1 deselected, 0 failed**
- [x] `delegate()` bypass: FAILED before fix, PASSED after fix
- [x] `verify()` bypass: FAILED before fix, PASSED after fix
- [x] Same-seed sha256 determinism: `b49eeaf935771bb0023f6deb1c0de9f4874b5dd1b8c189b21119dd91b7ffbd36` both runs

```
# delegate() bypass — BEFORE fix (vulnerability confirmed):
VULNERABILITY CONFIRMED: Bob minted a valid child token using alice's stolen token!

# delegate() bypass — AFTER fix:
TestDelegateCallerBypass::test_delegate_rejects_wrong_caller PASSED
TestDelegateCallerBypass::test_delegate_without_caller_also_rejected PASSED
TestDelegateCallerBypass::test_delegate_accepts_correct_caller PASSED

# verify() bypass — BEFORE fix (vulnerability confirmed):
Failed: DID NOT RAISE <AudienceConfusionError>

# verify() bypass — AFTER fix:
TestVerifyCallerBypass::test_verify_without_caller_rejected_for_audience_bound_token PASSED
TestVerifyCallerBypass::test_verify_root_token_without_caller_still_passes PASSED
```


mwarsss:
Update on the review feedback.

All five points are addressed on this branch:

1. delegated_auth_factory now instantiates DelegatableAuth instead of
passing the class. Added an end to end test that runs
scenarios/delegated_auth.yaml to completion.

2. Token ids are now deterministic (counter based, not uuid4), and the
clock is injected instead of using wall time. Added a same seed test
that hashes the full trace and confirms two runs produce identical
output.

3. Audience confusion is fixed, and this took two passes. The first
version added a caller check in delegate() but only enforced it when
caller was explicitly passed, so the exploit you described still worked
if caller was omitted. Re running your exact repro against that "fixed"
code confirmed it. The check is now unconditional in delegate(), and I
found the same gap existed in verify() as well (same pattern, caller
optional meant the audience check silently skipped). Fixed there too.
Both are covered by tests that call without caller and confirm
rejection, plus tests confirming root tokens (which have no audience)
are unaffected.

4. Branch renamed to match the charter format. This closed the original
PR as a side effect of the rename severing the fork's PR reference,
which is why this work continues here instead of on #103. Rebased on
current main, full suite passing (836 passed, 1 skipped, 1 deselected).

5. Added a differentiation section to the PR body: key chain revocation
with no per child revocation list, typed ValueError compatible
exceptions, and validators proven against a JWT baseline rather than
just asserted.

One gap that's intentionally still open: replay. A valid token can be
presented to verify() any number of times within its TTL, there is no
nonce or single use marker. This isn't claimed as blocked anywhere in
the docstring or PR body. Flagging it explicitly in case it should be
in scope for this problem.

Ready for another look when you have time. Also, since this is my first
contribution to the repo, CI hasn't run yet. Would appreciate an
approve and run when convenient.

shilpa-kulkarni-14:
@mwarsss plz resolve the conflicts.

mwarsss:
Resolved the merge conflicts against the latest `main`.

Changes made during conflict resolution:

* Restored plugin registrations that were unintentionally dropped during the merge.
* Removed a duplicate `delegated_auth` scenario registration after upstream added the same registration.
* Removed temporary debug output from the scenario tests.

No functional changes were made to the delegated authentication implementation itself beyond the conflict resolution. The remaining diff contains only the Problem 04 feature implementation and its associated tests.

mwarsss:
**Systematic audit of optional-parameter guards on `raise` statements — `delegatable.py`**

Every conditional that guards a `raise` via an `is not None` check or an optional-parameter guard was enumerated. There are exactly two such conditions in the file:

---

**1. `verify()` — audience confusion check (line 304)**

```python
raw_aud = data.get("aud")
audience: AgentId | None = AgentId(str(raw_aud)) if raw_aud is not None else None
if audience is not None and (caller is None or caller != audience):
    raise AudienceConfusionError(...)
```

`audience is not None` is gated on the *token's own payload*, not on whether `caller` was supplied. When `caller` is omitted (default `None`) and the token carries an audience field: `caller is None` evaluates to `True` → the full condition fires → **FAIL CLOSED**. A caller who omits `caller=` on an audience-bound token is rejected with the same exception as one who passes the wrong identity.

---

**2. `delegate()` — audience confusion check on parent token (line 385)**

```python
raw_aud = parent_data.get("aud")
if raw_aud is not None and (caller is None or str(caller) != str(raw_aud)):
    raise AudienceConfusionError(...)
```

`raw_aud is not None` is gated on whether the *parent token* carries an audience binding. When `caller` is omitted and the parent token has a declared audience: `caller is None` evaluates to `True` → condition fires → **FAIL CLOSED**. This check runs before `verify()` is called, so an attacker calling `delegate(stolen_token)` with no `caller=` keyword is rejected immediately, before any registry lookup.

---

**Conclusion**

No third instance of this pattern exists in the file. Every other `raise` in the file is unconditional (signature check, expiry check, scope escalation check, TTL exhaustion, malformed token) or gated on internal state (`parent_id is None`, `rec is None`) with no optional-parameter path. This audit was performed by exhaustive grep, not by assumption.

JamesCarnley:
Thank you @mwarsss for the delegation work and the careful follow-up on omitted caller checks. I reviewed the final revision `1918f3ff` and the discussion against the rebuilt Town main.

The directly applicable requirement is now covered by merged #248 (`dacd08f83bf35eb56a0f5fc388b9e49a91b8533c`): seven regression cases check that a current RunGrant cannot join with an omitted/empty, wrong-role, or wrong-run session proof. The HTTP cases also verify no session/permissions are installed on rejection, and a correctly signed presentation still works. This adds tests around existing behavior; it does not claim to port this PR's delegation implementation. Verification: 392 tests passed, all three GitHub CI jobs passed, and deliberate local verifier mutations were detected by the new tests.

The remaining parent/child delegation and cascading-revocation system needs an explicitly designed protocol profile rather than replacing current flat RunGrants. Requirements retained for that work include authenticated presentation at verification and onward delegation, verifier-enforced scope/expiry containment, ancestor revocation with unaffected-sibling controls, bounded chain validation, and restart/replica behavior. The legacy scenario itself notes that its helper does not pass the required caller for onward delegation, so it cannot serve as proof of a healthy complete three-level flow yet. A future profile also needs to distinguish an expected policy rejection from an unsupported API or unrelated error.

Closing this archive-targeted PR as partially superseded by #248, with the broader delegation profile deferred, not implemented. The original contribution and discussion remain linked here for credit and future reference. Reusable-token versus single-use/replay policy remains a separate design choice.

PR #157, opened 2026-07-10
[Hackathon] adarx: tiered-discount negotiation plugin with strategic bargaining

merry-tresata:
## What this adds

A third plugin for the Negotiation layer — adarx — alongside the existing alternating_offers (Rubinstein patience-discount) and pareto (utility-based trade-off) reference implementations.

The existing two plugins are general-purpose bargaining models, designed to work for any negotiable good regardless of domain — that generality is their strength, and it's exactly what makes them good defaults for the simulator. adarx takes a complementary approach: instead of a domain-agnostic model, it ports real, independently-tested procurement business logic from a working system (an autonomous procurement agent). The goal isn't to replace the general-purpose plugins, but to give the layer a concrete example of what a domain-specific negotiation protocol looks like in practice, and to test whether the simulator handles that kind of plugin as well as it handles the mathematical ones.

## Why this is a useful addition

The existing plugins are well-suited to testing whether a general bargaining protocol converges mathematically. This plugin gives the same test rig something complementary to work with: a negotiation protocol built around real business rules — volume discounts, strategic posture, compliance gating — so we can see how it holds up under Nanda Town's adversarial simulation conditions (message drops, Byzantine agents, scale).

Concretely, adarx demonstrates a different style of negotiation logic than the layer currently has an example of:

- Tiered volume discounts — larger orders unlock better pricing in real percentage steps (5% at 500+ units, 8% at 1000+, 12% at 5000+), a business rule rather than a continuous mathematical decay curve.
- Strategic posture — an agent can negotiate aggressive (bigger discount, slower delivery/payment), defensive (smaller discount, faster delivery, longer warranty), or balanced, changing the shape of the deal, not just the price.
- A hard compliance gate — a counterparty flagged as non-compliant is rejected immediately at open(), before any bargaining round is spent — modeling a real-world constraint (you don't negotiate price with a supplier that hasn't passed compliance) that's naturally out of scope for a general-purpose bargaining model.

## How business context reaches the plugin

Terms.metadata (a free-form dict on the SDK's Terms type) carries quantity, strategy, and compliance_status, seeded by the initiating agent at open(). The negotiated price target is computed once, from the fixed opening price (session.history[0]), and held constant — later rounds compare whatever's currently on the table against that fixed target, rather than recomputing a moving target every round.

## Update — addressing review feedback

Following review, this PR now also includes:

- Discriminator scenario — a new adarx_compliance_gate scenario (scenarios/adarx_compliance_gate.yaml, factory in nest_core/scenarios_builtin/adarx_compliance_gate.py, validators in nest_core/validators.py) runs 10 buyer/seller pairs, half against an APPROVED counterparty and half against PENDING_REVIEW. ADARXNegotiation rejects the non-compliant half at open(), before any price is exchanged; alternating_offers has no compliance concept and reaches agreement with them anyway. Swapping negotiation: adarx → negotiation: alternating_offers in the YAML flips validate_adarx_compliance_gate from PASS to FAIL — verified locally both ways.
- Deterministic session ids — adarx.py no longer uses uuid.uuid4(); session ids are now derived as f"adarx-{agent_id}-{counter}" from a per-instance counter, matching the pattern already used in pareto.py.
- File placement — the scenario config that was at repo root (marketplace.yaml) has moved into scenarios/ (renamed adarx_marketplace.yaml, since scenarios/marketplace.yaml already exists), and report-adarx.html has been removed from the repo.

## Testing

Unit + property tests (test_adarx_negotiation.py, 16/16 passing):
- Exact discount-tier boundaries and all three strategy adjustments, unit-tested directly
- Hypothesis property tests, matching the rigor of the existing test_pareto_negotiation.py:
  - Determinism — identical construction + identical offer yields an identical response (200 examples)
  - Monotonic discount in quantity — the discount never decreases as requested quantity increases, for any quantity (200 examples)
  - Bounded price — the negotiated price is never negative or above the base price, across any quantity/strategy combination (200 examples)
  - Absolute compliance gate — for any non-APPROVED status, the session is rejected at open() and stays rejected through respond()/close(), regardless of any offers made afterward (50 examples)

Discriminator test (adarx_compliance_gate scenario, see above): PASS with negotiation: adarx, FAIL with negotiation: alternating_offers on the same validator — direct evidence the plugin's compliance gate is genuinely discriminating, not just present.

Full CI (make ci-local): all 5 checks pass — uv sync, ruff check, ruff format --check, pyright (strict mode), and the full repo test suite (1166 tests, no regressions introduced).

Note on the marketplace integration run: an earlier version of this PR included a 100-agent marketplace scenario run as "integration evidence" (now at scenarios/adarx_marketplace.yaml). That scenario's BuyerAgent/SellerAgent negotiate price via their own buy:/sold:/reject: messages and never call into the negotiation: plugin — so it didn't actually exercise ADARX's logic, regardless of which plugin was configured there. The real integration evidence is the adarx_compliance_gate scenario described above, which drives ADARXNegotiation.open/offer/respond/close directly.

## Design notes / known limitations

- This is a deliberately simple, deterministic model (no randomness, no learning) — appropriate for regression testing and Tier-1 (state-machine) agents. It has not been tested against Tier-2 (LLM-backed) agents.
- The compliance gate is binary (APPROVED vs. anything else) — a real system might model graduated compliance tiers instead.
- Business context travels through Terms.metadata, since that's the only free-form field the current Negotiation interface exposes; a future interface revision with typed negotiation context would be cleaner than dict-based metadata passing.

---


shilpa-kulkarni-14:
@merry-tresata  Reviewed on latest main . Competent plugin, but I can't merge it as a ranked
contribution yet.Three things block it:

1. No discriminator. The tests are all plugin-internal property checks, with no
   comparison to alternating_offers or pareto. Nothing passes on ADARX and fails on a
   baseline, so the claim that it negotiates better than the existing plugins is
   unproven. Add a validator + scenario where the baseline loses and ADARX wins, e.g.
   the compliance gate rejects a non-compliant deal the baseline accepts, or a tiered
   discount lands a price the baseline misses, wired so swapping to the baseline flips
   the validator.

2. uuid4 non-determinism. adarx.py:132 mints offer ids with uuid.uuid4(). The trace is
   deterministic here only because this scenario doesn't serialize the id, but it's
   unseedable, so it's a latent ADR-004 violation. This is the exact bug #151 (same
   org) just fixed in coordination and flagged for negotiation. Reuse that
   derive-id pattern: a pure function of (agent, task, seq), no uuid4.

3. Placement. marketplace.yaml belongs in scenarios/, not repo root, and
   report-adarx.html shouldn't be committed (drop it, or move to docs/).

The strategy itself is distinct and the property tests are thorough. Once there's a
non-vacuous discriminator, the id derivation is deterministic, and the files are in
the right place, this is mergeable.

merry-tresata:
@shilpa-kulkarni-14  thanks for the thorough review,its really helpful. Just pushed a commit (42ab727) that should cover all three points:

1. Discriminator — added a new scenario, adarx_compliance_gate. It runs 10 buyer/seller pairs, half with an APPROVED counterparty and half PENDING_REVIEW. ADARX's compliance gate rejects the non-compliant half right at open(), before any price talk happens — but alternating_offers doesn't know what compliance even means, so it just negotiates and agrees with them anyway. Swap negotiation: adarx to negotiation: alternating_offers in the YAML and the validator flips from PASS to FAIL. Tested both ways locally, works as expected.

2. uuid4 — yeah, fair catch. Swapped it out for a plain counter, f"adarx-{agent_id}-{counter}", same idea as what pareto.py already does. Deterministic now, no more uuid4 anywhere in the file. All the existing tests still pass.

3. File placement — moved marketplace.yaml into scenarios/ (had to rename it to adarx_marketplace.yaml since scenarios/marketplace.yaml already existed) and just deleted report-adarx.html from the repo entirely.

One extra thing I ran into while moving that marketplace file, figured I'd flag it instead of quietly burying it: turns out the marketplace task type does its own price haggling through plain buy:/sold:/reject: messages and never actually touches the negotiation: plugin. So that 100-agent marketplace run isn't really testing ADARX's negotiation logic at all, no matter what plugin's plugged in there. Left a note about that at the top of the moved yaml. The new adarx_compliance_gate scenario is the one that's actually exercising open/offer/respond/close. Let me know if you want me to also tweak the PR description so it doesn't overstate what that marketplace test proves.

Ran the full CI suite locally too — ruff, pyright, pytest all green (1166 passed). Let me know if anything's still missing.

JamesCarnley:
Thanks @merry-tresata for the domain-specific procurement work and the candid follow-up about the original marketplace run. I reviewed all six changed files at `83c971e6e23ac1e0e3ecb6b8d9ee2d1b1e01372e` and the discussion. The deterministic session counter, relocated scenarios, and explicit approved/non-approved comparison are meaningful improvements over the initial submission.

Closing this legacy integration without a current-main replacement. Rebuilt Town's `Haggle` layer records alternating offers and acceptance; it does not own a universal procurement/compliance policy. These volume tiers and business rules are better preserved as an optional workload for Town to test than installed as a new default policy.

Important verification limit: the submitted tree contains the new `adarx_compliance_gate.py` factory and YAML, but its `nest_core/scenarios.py` has no ADARX loader entry and `nest_core/validators.py` has no ADARX validator or registration. The PR body's end-to-end claim therefore is not established by this exact head. I did not execute the legacy implementation. Its revised unit/property checks remain useful, and the author-reported legacy pass counts are not being treated as independently reproduced evidence.

Preserving a concrete future pilot: one independently maintained procurement agent, a frozen order with quantity/price/terms, an independently supplied policy decision, approved and refused controls, and a report that separately checks negotiation behavior and the final agreement. Include omitted/unknown compliance, exact tier boundaries, round-limit behavior, and accepted-term consistency. `open()` currently defaults omitted compliance to APPROVED, while `close()` can create an agreement from current terms without independently establishing acceptance; those distinctions must be explicit in a profile rather than described as a universal fail-closed gate.

I ran both YAMLs only as input data through CURRENT Town: each becomes a substituted reference flow with `original_scenario: not_tested`. Those runs are not ADARX validation. No procurement runtime, external integration, or replacement merge is claimed. This source remains a credited pilot lead, not an outreach commitment or evidence of outside adoption.

PR #158, opened 2026-07-10
[hackathon phase 1] fix: return defensive copy from CidFacts.fetch()

arii:
Fixes a bug where CidFacts.fetch() returned the mutable internal object instead of a copy, which allowed the caller to unintentionally tamper with the registry content.

Updates CidFacts.fetch() to return meta.model_copy(deep=True).
Adds a new protocol conformance test test_fetch_returns_defensive_copy explicitly testing the immutability of the fetched registry data.

### DIFF
--- a/packages/nest-plugins-reference/nest_plugins_reference/datafacts/cid_facts.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/datafacts/cid_facts.py
@@ -232,7 +232,7 @@ class CidFacts:
         if meta is None:
             msg = f"Dataset not found: {url}"
             raise KeyError(msg)
-        return meta
+        return meta.model_copy(deep=True)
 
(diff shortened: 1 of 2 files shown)


shilpa-kulkarni-14:
@arii  Verified on latest main. Clean fix, approving as a maintenance merge.

- The bug is real: CidFacts.fetch() returned the internal model by reference, so a
  caller could mutate stored registry state. model_copy(deep=True) is the right fix.
- The regression test is genuine: it mutates the fetched object, re-fetches, and
  asserts the internal state is unchanged. Fails without the fix.
- ci-local 5/5 green (1240 passed), cid_facts suite 31 passed.

This merges in the maintenance lane, not as a ranked stack entry (no new capability or
discriminator), same as the score_average hardening and the coordination determinism
fix.

Noting for the record that this is bot-authored (google-labs-jules). The fix is
correct regardless.

shilpa-kulkarni-14 (approved):

PR #159, opened 2026-07-10
[Hackathon] dhritimandas: capability-conformance registry gate

dhritimandas:
## Problem

`AgentCard.capabilities` (`nest_core/types.py:82`) is a bare, self-asserted `list[str]`. `Registry.lookup()` does string matching against this list. Every existing defense in the repo verifies who published a card (signatures, identity, `ed25519_rotating`) or scores an agent's overall behavior (`agent_receipts`, `score_average`), but nothing checks a specific capability claim against whether the agent has ever fulfilled it. An agent can register `capabilities=["sell"]` and remain discoverable via that claim forever, even if it never completes a single sale.

## Change

- `VerifiedCapabilitiesRegistry` (`nest_plugins_reference/registry/verified_capabilities.py`), registered as `("registry", "verified_capabilities")`: wraps an inner registry and tracks defection counts per `(agent_id, capability)` pair. A pair with no evidence is never excluded (bootstrap allowance). Once a pair reaches `defection_threshold` (default 1, justified in the docstring against the scenario's zero message-drop rate) defections, `lookup` stops returning that agent for queries naming that capability — other capabilities of the same agent are unaffected. `report_fulfillment` resets the count to zero (re-admission, not a permanent blacklist); this is a documented trade-off, not an oversight — a periodic fulfiller could evade indefinitely, out of scope here.
- `scenarios/capability_spoofing.yaml` + `capability_spoofing_factory`: 3 honest sellers, 2 always-silent spoofers, 1 bait-and-switch seller (fulfills once, then goes silent), 5 buyers. Buyers discover sellers via `registry.lookup(Query(capabilities=["sell"]))`, send one outstanding `buy:` at a time, and self-schedule a `TIMEOUT:<round>` tick. A `sold:` response before the timeout reports a fulfillment; the timeout firing first reports a defection.
- `check_capability_conformance` (`nest_plugins_reference/validators/capability_validators.py`): registry lookups aren't traced (confirmed against `nest_core/sim/simulator.py`), so this validates directly against `ScenarioRunner.resolved_plugins["registry"]`, following the same `ValidatorReport` pattern as `registry_byzantine_validators`/`gossip_validators`. Checks two things together, so a gate that blocks everyone can't pass by accident: no spoofer is still discoverable, and every honest seller still is.

## Test

Literal output, same scenario YAML, only `layers.registry` overridden:

```
registry='in_memory'             passed=False detail="still discoverable for 'sell': ['baitswitch-0', 'spoofer-0', 'spoofer-1']"
registry='verified_capabilities' passed=True  detail="3 spoofer(s) excluded from 'sell', 3 honest agent(s) still discoverable"
```

`test_verified_capabilities.py` (7 unit tests): collection error (`ModuleNotFoundError`) before the plugin existed; all 7 pass after — bootstrap allowance, per-capability exclusion isolation, re-admission on fulfillment, below-threshold tolerance, honest-agent non-interference, deregister delegation. `test_capability_spoofing_scenario.py`: FAIL/PASS matrix above at seeds 42/7/1337, plus a byte-identical-trace determinism test. Full `make ci-local`: 1164 passed, 1 skipped, 1 deselected.

## Fit

A2A v1.0 ships signed Agent Cards and NANDA's Verified AgentFacts paper names capability spoofing as an open problem; a signature proves the publisher, not the capability. `agent_receipts` scores an agent's overall reputation from receipts — this gates *discovery*, per *capability*, independent of an agent's standing on any other claim.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -25,6 +25,9 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("identity", "ed25519_rotating"): (f"{_REF}.identity.ed25519_rotating:Ed25519RotatingIdentity"),
     ("registry", "in_memory"): f"{_REF}.registry.in_memory:InMemoryRegistry",
     ("registry", "gossip"): f"{_REF}.registry.gossip:GossipRegistry",
+    ("registry", "verified_capabilities"): (
+        f"{_REF}.registry.verified_capabilities:VerifiedCapabilitiesRegistry"
+    ),
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
     ("auth", "delegatable"): f"{_REF}.auth.delegatable:DelegatableAuth",
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -175,3 +175,9 @@ def _try_load_builtin(name: str) -> None:
         from nest_core.scenarios_builtin.sybil_bond import sybil_bond_factory
 
         register_scenario("sybil_bond", sybil_bond_factory)
+    elif name == "capability_spoofing":
+        from nest_core.scenarios_builtin.capability_spoofing import (
+            capability_spoofing_factory,
+        )
+
+        register_scenario("capability_spoofing", capability_spoofing_factory)
--- a/packages/nest-core/nest_core/scenarios_builtin/capability_spoofing.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/capability_spoofing.py
@@ -1,0 +1,315 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Capability-spoofing scenario — buyers discover sellers whose claims are not all true.
+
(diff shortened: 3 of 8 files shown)


shilpa-kulkarni-14 (approved):
@dhritimandas  Verified on latest main. Clean, approving.

- Merges clean, ci-local 5/5 green (1221 passed), Registry-conformant.
- Discriminator is non-vacuous and verified across seeds: the scenario tests confirm
  in_memory fails conformance at 42/7/1337 (the spoofer stays discoverable) while
  verified_capabilities passes (the spoofer is quarantined). Deterministic, and
  test_scenario_is_byte_for_byte_deterministic passes.
- The gap is real and well-framed: a signature is not a verified skill. ed25519_rotating
  and byzantine_gossip verify identity and card integrity; nothing verified whether an
  agent can actually fulfill what it advertises. This gates discovery on demonstrated
  fulfillment, which is exactly what A2A signed cards and NANDA Verified AgentFacts name and don't close. Distinct from #148, which defends propagation integrity, not
  conformance.

Two small items :
1. Strip the Claude Sonnet 5 co-author footer.
2. Add a docs/layers/registry.md note for the conformance-gate variant, as #148 did.

Nice threat model, and a good example of naming a gap the existing plugins explicitly
leave open rather than re-solving one they already cover.

PR #160, opened 2026-07-10
[Hackathon] dashrath: agent_receipts — sever collusion by evidence shape, not size (closes #97)

### LINKED ISSUE #97: agent_receipts: collusion severance assumes largest SCC is honest (majority-Sybil-invertible)
### Summary

In `agent_receipts` (reference trust plugin), the collusion-severance heuristic assumes the **largest** strongly-connected component is the honest anchor and never severs it. That assumption is invertible by a Sybil ring larger than the honest core: the ring becomes the anchor and keeps its wash-traded reputation, while the honest minority is severed to score 0. The plugin's anti-collusion property inverts against a majority-size ring.

This is a heuristic limitation rather than a crash, and the docstring is explicit about the assumption (`"The largest SCC is the honest anchor"`), so I'm filing it as a design concern rather than a defect. It seemed worth surfacing because it flips the exact property the plugin is meant to provide.

### Where

`packages/nest-plugins-reference/nest_plugins_reference/trust/agent_receipts.py`

- `_sccs` returns components sorted largest-first: `return sorted(comps, key=lambda c: (-len(c), c))` (~line 360)
- `_severed_dids` takes `anchor = set(comps[0])` and only ever evaluates `comps[1:]` for severance (~line 404)

### Reproduction

```python
from nest_plugins_reference.trust.agent_receipts import _severed_dids, _sccs

def clique(nodes):
    return {a: {b: 1 for b in nodes if b != a} for a in nodes}

honest = ["h1", "h2", "h3"]                 # honest core, mutually corroborating
ring   = ["s1", "s2", "s3", "s4", "s5"]     # Sybil ring, larger, mutually corroborating
graph = {**clique(honest), **clique(ring)}  # two dense SCCs, no cross-edges

print(_sccs(graph))        # [['s1'..'s5'], ['h1','h2','h3']]  -> ring is largest
print(_severed_dids(graph))  # {'h1', 'h2', 'h3'}  -> honest core severed, ring untouched
```

Output:

```
SCCs (largest-first): [['s1','s2','s3','s4','s5'], ['h1','h2','h3']]
anchor (treated as honest): ['s1','s2','s3','s4','s5']
severed dids: ['h1','h2','h3']
honest core severed? True | Sybil ring severed? False
```

Since Sybils are cheap to add, the ring can always be grown past the honest core, at which point the colluders are immune and honest agents with a small dense mutual-corroboration cluster get zeroed.

### Impact

- Colluding ring (majority size): retains full reputation.
- Honest agents in a small dense cluster: severed to score 0 / confidence 0.
- A related edge (`_severed_dids`, ~lines 412-415): an isolated honest mutual pair is also severed.

### Suggestion

Size alone can't distinguish honest from Sybil. A severance decision anchored on something harder to manufacture (e.g. external corroboration volume, edge age, or stake rather than component size) would avoid the majority-Sybil inversion. Happy to help if useful.

Dashrath175:
## 1. The inversion (closes #97)

Credit to @SwasthikaDev's issue #97 for identifying the flaw.

The collusion severance logic in `agent_receipts` exempted the **largest strongly connected component (SCC)** as the presumed "honest anchor" and evaluated only smaller components. Because Sybil identities are effectively free, **component size is attacker-controlled**, making the largest component the easiest thing to manipulate.

Running the reproduction on `main`:

```bash
uv run python -c "from nest_plugins_reference.trust.agent_receipts import _severed_dids
c = lambda ns: {a:{b:1 for b in ns if b!=a} for a in ns}
print(sorted(_severed_dids({**c(['h1','h2','h3']), **c(['s1','s2','s3','s4','s5'])})))"
```

Output on `main`:

```text
['h1', 'h2', 'h3']
```

Result:
- Honest 3-node clique is severed.
- 5-node Sybil ring survives.

Running the exact same reproduction on this branch produces:

```text
['h1','h2','h3','s1','s2','s3','s4','s5']
```

The majority collusion ring can no longer escape detection.

---

## 2. The fix

The new invariant is simple:

> **Component size is never treated as evidence.**

Every SCC—including the largest—is evaluated using the exact same rule.

A component is severed **iff**:

- it is collusion-shaped
  - dense ring (≥ `RING_MIN_SIZE`)
  - density ≥ `RING_MIN_DENSITY`
  - or a mutual-only pair
- **and**
- it has **zero corroborated cross-edges** to every clean component.

The SCC algorithm itself (`_sccs`), thresholds, and previously validated behaviour remain unchanged.

| Scenario | `main` | This branch |
|----------|---------|-------------|
| Majority ring (8) vs honest cycle (5) | Majority ring exempt | Majority ring severed |
| Existing shipped case (honest 8 + ring 4) | Ring severed | Identical behaviour |
| Ring bridged to one honest agent | Exonerated | Identical behaviour |
| Two isolated rings | Both severed | Identical behaviour |
| Mutual-only pair | Severed | Identical behaviour |
| Issue #97 reproduction | Honest severed, ring survives | Both severed |

---

## 3. Tradeoff

An isolated dense clique with **no external corroboration** is information-theoretically indistinguishable from a wash-trading ring.

Instead of guessing based on size (fail-open), the plugin now chooses the fail-safe behaviour:

- refuse to certify isolated dense cliques
- require at least one corroborated external edge to establish legitimacy

Honest agents naturally become trusted by interacting outside their clique.

---

## 4. Alternatives considered

Two alternatives were evaluated:

- stake anchoring (cross-layer dependency on payments/bond)
- edge-age / clock semantics

Both introduce unnecessary cross-layer coupling.

The implemented solution follows the direction proposed in Issue #97 while relying only on information already available inside the trust plugin.

---

## 5. New red-team scenario

Added:

```text
scenarios/receipt_reputation_majority_ring.yaml
```

This scenario creates:

- 8-agent Sybil majority ring
- 5 honest agents
- unchanged factory implementation
- deterministic replay

New validator suite:

```text
receipt_reputation_majority
```

Validators verify:

- majority ring is actually present
- majority ring is severed
- enforcement precondition exists
- discriminator against competing implementations

Expected behaviour:

| Implementation | Result |
|---------------|--------|
| `agent_receipts` (main) | FAIL |
| Fixed `agent_receipts` | PASS |
| `score_average` | FAIL |

---

## 6. Verification

Run the scenario:

```bash
uv run nest run scenarios/receipt_reputation_majority_ring.yaml
```

Validate:

```bash
python -c "from pathlib import Path; from nest_core.validators import validate_trace; [print(('PASS' if r.passed else 'FAIL'), r.name, r.detail) for r in validate_trace(Path('traces/receipt_reputation_majority_ring.jsonl'),'receipt_reputation_majority')]"
```

Verification performed:

- 5-seed deterministic replay
- byte-identical traces (SHA256)
- new example tests covering every scenario
- new Hypothesis property proving severance is independent of SCC size

---

## 7. Compatibility

This change is fully backward compatible.

- ✅ No existing behaviour regressed
- ✅ All previous tests pass unmodified
- ✅ `ruff check`
- ✅ `ruff format --check`
- ✅ `pyright`
- ✅ `pytest`

CI summary:

```text
1168 passed
0 type errors
5/5 CI checks green
```

---

## Files added / modified

- `agent_receipts.py`
- `validators.py`
- `receipt_reputation_majority_ring.yaml`
- validator registry
- property tests
- unit tests
- end-to-end scenario tests
- trust documentation

---

## Why this matters

Issue #97 exposed a structural weakness where **graph size** was implicitly trusted.

This PR removes that assumption entirely.

Collusion detection is now based on **evidence shape and corroboration**, not population size, making majority-Sybil attacks ineffective while preserving all previously validated behaviours.

---

🤖 Generated with Claude Code

### DIFF
--- a/docs/layers/trust.md
+++ b/docs/layers/trust.md
@@ -30,14 +30,31 @@ malicious + 1 observer that samples cheat reports probabilistically.
 Derives reputation from **cross-signed receipts** of real interactions
 instead of self-asserted feedback. A receipt builds reputation only if
 its Ed25519 issuer signature verifies, a *distinct* counterparty
-co-signed the same interaction, and the receipt does not sit inside an
-isolated collusion component (Tarjan SCC severance over the
-corroboration graph) — so a wash-trading ring collapses to score ~0
-while honest agents keep their corroborated score.
+co-signed the same interaction, and the receipt does not sit inside a
+severed collusion component — so a wash-trading ring collapses to score
+~0 while honest agents keep their corroborated score.
+
+Severance judges every SCC of the corroboration graph by **evidence
+shape and external corroboration, never by size**: a component is
+severed iff it is collusion-shaped (a dense ring or a mutual-only pair)
+*and* has no corroborated edge to any clean component. Component size is
+free for a Sybil operator to manufacture, so a ring grown larger than
+the honest population buys no immunity
+([issue #97](https://github.com/projnanda/nandatown/issues/97)).
 
 Source: [`nest_plugins_reference/trust/agent_receipts.py`](../../packages/nest-plugins-reference/nest_plugins_reference/trust/agent_receipts.py).
-Scenario: `receipt_reputation` (honest anchor + isolated 4-agent ring +
-byzantine co-signers).
+Scenarios: `receipt_reputation` (honest majority + isolated 4-agent ring
++ byzantine co-signers) and `receipt_reputation_majority_ring` (the
+issue #97 red-team: an 8-agent ring *outnumbering* 5 honest agents —
+severed all the same):
+
+```bash
+nest run scenarios/receipt_reputation_majority_ring.yaml
+python -c "from pathlib import Path; from nest_core.validators import validate_trace; \
+    [print(('PASS' if r.passed else 'FAIL'), r.name, r.detail) \
+     for r in validate_trace(Path('traces/receipt_reputation_majority_ring.jsonl'), \
+                             'receipt_reputation_majority')]"
+```
 
 ## `parc` — portable reputation credentials
 
--- a/packages/nest-core/nest_core/validators.py
+++ b/packages/nest-core/nest_core/validators.py
@@ -3176,6 +3176,55 @@ def validate_receipt_reputation_honest_confidence(
     ]
 
 
+def validate_receipt_reputation_ring_majority(
+    events: list[dict[str, Any]],
+) -> list[ValidationResult]:
+    """The scored collusion ring strictly outnumbers the scored honest agents.
+
+    This is the enforcement-liveness check for the majority-ring red-team
+    (issue #97): it proves the trace actually exercised the attack the other
+    two validators defend against — a ring *larger* than the honest population,
+    which the old largest-SCC anchor exemption would have crowned as the honest
+    anchor. Without this check, a scenario quietly shrunk back to a minority
+    ring would pass ``ring_severed`` without ever testing the inversion.
+
+    PASSes iff both populations were scored and ring count > honest count.
+    FAILs on missing populations or a non-majority ring — without crashing.
+
+    Example::
+
+        results = validate_receipt_reputation_ring_majority(events)
+    """
+    scores = _collect_scores(events)
+    ring_count = sum(1 for (_s, _c, role) in scores.values() if role == "ring")
+    honest_count = sum(1 for (_s, _c, role) in scores.values() if role == "honest")
+
+    if ring_count == 0 or honest_count == 0:
+        return [
+            ValidationResult(
+                "receipt_reputation_ring_majority",
+                False,
+                f"missing populations: {ring_count} ring, {honest_count} honest scored",
+            )
+        ]
+    if ring_count <= honest_count:
+        return [
+            ValidationResult(
+                "receipt_reputation_ring_majority",
+                False,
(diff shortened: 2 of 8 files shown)


shilpa-kulkarni-14 (approved):
@Dashrath175 Verified on latest main. Clean, approving.

- Merges clean, ci-local 5/5 green (1239 passed). Hardening the merged agent_receipts
  breaks nothing.
- The vulnerability is real: the old "largest SCC is the honest anchor" heuristic is
  defeated by a majority ring, which becomes the largest component and gets exempted.
  Judging by shape (dense ring or mutual-only pair) plus external corroboration, never
  size, closes it.
- Discriminator flips cleanly and is verified across 5 seeds: agent_receipts severs the
  8-agent ring to ~0 while retaining the 5 honest agents; score_average leaves the ring
  at 1.0 and fails both defending validators.
- The ring_majority liveness validator is the right move: it passes only if the trace
  actually staged a majority ring (8 > 5), so the discriminator can't be vacuous. That's
  the discipline I want to see, and it's exactly what #139 was missing.
- Deterministic across seeds.

One governance item: this modifies the merged agent_receipts, so flagging @maria (or the
trust-layer owner) for a quick ack before merge. Also strip the Claude Fable 5 co-author
footer.

This supersedes the collusion-detection intent of #139 and does it with a real
discriminator. Nice work.

PR #163, opened 2026-07-10
[Hackathon] auth-safety-engineer: DelegatableAuth with scoped delegation, cascading revocation, and deterministic validators

Niruuu2005:
## Layer picked: **Auth (#5)** — Problem 04

**Persona:** `auth-safety-engineer`

## Summary

Implements Problem 04: Delegatable capability tokens with cascading revocation.

## What changed

- Merged latest `main` and resolved shared-file conflicts.
- Added documented auth plugin entry point in `packages/nest-plugins-reference/pyproject.toml`.
- Preserved `verify(token, presenter=None)` compatibility with the base Auth protocol.
- Delegated tokens fail closed when `presenter` is omitted.
- Enforced `"delegate"` scope before minting child tokens.
- Split `stale_parent` from `transitive_revocation` in the scenario trace.
- Made YAML `root_scopes` and `delegation_tree` load-bearing.
- Fixed deterministic clock injection for `JwtAuth` contrast path.

## Why this is distinct

Unlike baseline entries with hardcoded topology, this scenario makes `root_scopes` and `delegation_tree` load-bearing.

## Validation

```bash
```

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -26,6 +26,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("registry", "in_memory"): f"{_REF}.registry.in_memory:InMemoryRegistry",
     ("registry", "gossip"): f"{_REF}.registry.gossip:GossipRegistry",
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
+    ("auth", "delegatable"): f"{_REF}.auth.delegatable:DelegatableAuth",
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
     ("trust", "agent_receipts"): f"{_REF}.trust.agent_receipts:AgentReceiptsTrust",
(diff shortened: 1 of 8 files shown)


shilpa-kulkarni-14:
@Niruuu2005 
Thanks for this, it's a competent implementation. Closing it, for two reasons.

First, this capability is already on main. delegatable auth with cascading revocation
merged via #138 (auth/delegatable.py and scenarios/delegated_auth.yaml are already
there), and #142 and #147 were closed earlier as the same feature. This is the fourth
independent submission of it, so the slot is filled.

Second, and more important: please do not merge this branch as-is. It's based on
7106ca6, well behind current main (fc38fd4), and its tree is missing four PRs that have
since merged, #152, #155, #159, and #160. Merging it would delete verified_capabilities,
mesh_revocable, failure_detection_forgery, capability_spoofing, and the
receipt_reputation_majority_ring scenario, and it also rewrites the merged #138. The
two-dot diff is +1320 / -4864. That's a revert of a lot of merged work.

This isn't a knock on the code, the delegatable idea was arrived at independently four
times. If you'd like to keep contributing, an unclaimed layer would be a much stronger
use of the effort.

PR #117, opened 2026-07-09
hackathon/yashjadhav1595-projects : Ed25519 rotating identity with key rotation

yashjadhav1595-projects:
### What this solves (Problem 05: Identity Key Rotation)
The default `did_key` plugin produces a static, deterministic identity that cannot be rotated — a compromised key remains valid forever. This PR adds `Ed25519RotatingIdentity`: a drop-in identity plugin that supports key rotation with cryptographic continuity proofs and temporal validity windows.

### Implementation Checklist
- [x] **Implement Ed25519 key rotation** where retiring keys cryptographically sign their successors via a continuity proof.
- [x] **Enforce temporal validity windows** (`as_of` tick bounds checking against `issued_at` and `rotated_out`).
- [x] **Implement chain-tip guarding** in `verify_continuity` and `apply_rotation` to prevent retired-key injection attacks.
- [x] **Integration with `parc` trust layer** by exposing the full public key history through `resolve()`.
- [x] **98 Unit & Integration Tests** passing (including full adversarial scenario validation for post-rotation forgery, backdating, and retired-key injection).

### The Solution: Cryptographic Continuity & Temporal Bounds
This PR implements `Ed25519RotatingIdentity`, an identity plugin that natively supports key lifecycle management and deterministic rotation.

**Key Features:**
*   **Cryptographic Continuity**: When `rotate_key()` is called, the old key signs a canonical `RotationRecord` binding the agent ID, old key ID, new key ID, and new public key, proving authorization of the new key.
*   **Temporal Validity Windows**: Signatures are verified against the specific time they were observed (`as_of`), rather than the attacker-controlled `signed_at` time. Keys are strictly valid only between their `issued_at` and `rotated_out` ticks.
*   **Chain-tip Guarding**: `verify_continuity()` prevents "Stale Key Injection" by ensuring that a key cannot authorize a new successor if it has already been rotated out.
*   **Adversarial Validators**: Two new validators (`validate_identity_post_rotation_forgery` and `validate_identity_backdating_rejected`) prove the system accurately blocks temporal boundary attacks.

### Validation & Testing
This implementation passes the full `identity_rotation` adversarial validator scenario out-of-the-box, correctly catching:
1.  **Post-rotation forgery**: old key used after its `rotated_out` tick -> rejected by temporal window check.
2.  **Backdating**: new key with claimed `signed_at` before its `issued_at` -> rejected by `as_of` anchor.
[...]

PR #166, opened 2026-07-10
[Hackathon] fix(identity): ed25519_rotating replay-corruption + destructive re-registration hardening

tony-nexartis:
## What this fixes

Red-team remediation of the merged `ed25519_rotating` identity plugin (#25), following the precedent of #108 red-teaming merged work. Every finding below was **reproduced against the merged code with a runnable probe** before fixing — no inspection-only claims.

| Finding | Severity | Fix |
| --- | --- | --- |
| `apply_rotation` replay corrupts key history | Medium (integrity) | Idempotent + chain-tip-only: replaying a rotation announcement is a no-op; a conflicting one raises. `resolve()` determinism restored under replay |
| `register_peer` silently wipes a known agent's history | Medium (integrity) | Non-destructive: re-announcing the current key is a no-op; a *different* key for a known agent raises (must go through `apply_rotation`) |
| `verify` accepts signatures dated outside the signing key's window | Hardening | `verify` now also rejects when `sig.signed_at` is present and outside the key's validity window — adopted from a partner analysis, whose variant we also fixed (it rejected honest cross-clock peer signatures) |
| `verify_continuity` "already applied" escape matches on `new_key_id` only | Defence-in-depth | Now cross-checks `new_public_key` + `issued_at` (not exploitable after the idempotency fix, but closes the gap for downstream consumers) |
| Same-tick sign+rotate ambiguity | Inherent limitation — NOT claimed fixed | An honest co-tick old-key signature and a post-rotation forgery are byte-identical under scalar logical time (the repo's own Hypothesis property produces the counterexample at `delta=0`). Instead of silently dropping honest messages or reopening forgery, `rotate_key` now **raises** if called on the tick the agent last signed — fail loud, documented on `verify` |

## Why it matters

Identity is the layer every other layer trusts. Before this fix, a single replayed rotation announcement (a plausible byzantine or at-least-once-delivery event) permanently corrupted a peer's key history, and any agent could silently reset another agent's registered key. Both attacks are one message each.

## Verification

```bash
uv sync
uv run ruff check . && uv run ruff format --check . && uv run pyright
uv run pytest packages/nest-plugins-reference/tests/test_ed25519_rotating.py -v
uv run pytest -q   # full suite green; identity suite 24 -> 34 tests
```

10 new regression tests: replay-idempotency ×3, non-destructive re-registration ×2, window-bound `signed_at` ×2 (including the honest cross-clock case the partner variant broke), same-tick guard ×2, continuity cross-check ×1.

Adjacent-code check: every existing caller of `rotate_key` / `register_peer` / `apply_rotation` in the repo (scenarios `identity_rotation`, `parc_migration`; `trust/parc`; their tests) already advances the clock between sign and rotate, registers each peer once, and applies each rotation once — the stricter semantics break no existing contract, and the full suite stays green.

## Honest limitations

- The same-tick boundary is **not fixable in `verify`** without readmitting forgery; we ship a loud grant-side guard instead and document the as-of verification model.
- Signatures are bound to a window, not to a point within it (as-of model) — documented, partly mitigated by the `signed_at` window check.

## Credit

This remediation was researched and built by **Bin (Yanez team, Nexartis hack squad)**; committed and pushed on his behalf from this account as a maintenance contribution. The `signed_at` window rule was adopted from his partner analysis (BIN-HACK-FIX).

### DIFF
--- a/packages/nest-plugins-reference/nest_plugins_reference/identity/ed25519_rotating.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/identity/ed25519_rotating.py
@@ -219,6 +219,11 @@ class Ed25519RotatingIdentity:
         self._seed = seed
         self._rotation_index = 0
         self._clock = 0.0
+        # The last tick at which this agent signed with its own key. Used by
+        # rotate_key to avoid stranding a co-tick signature at the rotation
+        # boundary (see the "Known limitation — same-tick rotation" note on
+        # verify). ``-inf`` means "never signed".
+        self._last_sign_tick = float("-inf")
         private = Ed25519PrivateKey.from_private_bytes(_derive_seed(seed, agent_id, 0))
         pub = _public_bytes(private.public_key())
         record = KeyRecord(
@@ -284,6 +289,16 @@ class Ed25519RotatingIdentity:
         keep working. The peer's window opens at the registrant's current clock
         and stays open until a later :meth:`apply_rotation`.
 
+        **Non-destructive on re-announce.** Registry announcements are
+        idempotent and duplicated in practice. Re-registering a peer's *current*
+        key is a no-op — it must never truncate rotation history the verifier
+        already learned via :meth:`apply_rotation`, because doing so would
+        silently re-open a retired key's window (stale-key acceptance). A key
+        *change* for a known agent must go through the continuity-checked
+        :meth:`apply_rotation`, never through ``register_peer``; presenting a
+        different key here for an agent whose current key is already known is
+        rejected.
+
         Example::
 
             ident.register_peer(AgentId("a2"), peer_pk)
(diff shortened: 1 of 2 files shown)


shilpa-kulkarni-14 (commented):
Verified on latest main. Clean, approving as a maintenance merge.

- API is fully preserved (identical public surface to main), so this is additive
  hardening, not a rewrite. Notably the opposite of #117, which removed public API and
  claimed a feature it didn't implement.
- Real replay/backdating hardening: apply_rotation is idempotent and chain-tip-only so a
  replayed rotation can't reopen a retired key's window; rotate_key raises rather than
  allowing a forgery window; verify rejects signatures backdated outside the signing
  key's window.
- ci-local 5/5 green (1250 passed). Dependents verified: parc, agent_receipts,
  attested_peering, ed25519 suites total 183 passed. Hardening tests 34 passed.
- Good catch fixing the honest cross-clock regression so the guards don't over-reject.

Since this touches the merged ed25519_rotating that several plugins depend on, flagging
@maria (or the identity-layer owner) for a quick ack before merge. Merges in the
maintenance lane, not as a ranked stack entry.

If you want this to count as a ranked contribution, a small replay scenario with a
harden-off foil and a validator would make the security gain a real discriminator.
Optional.

JamesCarnley:
Thanks @tony-nexartis, and Bin / the Yanez team / Nexartis for the implementation and partner-analysis credit documented here. This is a useful maintenance fix for the original rotating-key plugin, and the earlier positive review is preserved.

Reviewed the full diff against current main at f3664a8. Current Town uses static controller keys, not this rotation-history implementation. Repeated identity creation, pinned controller identities, older-grant replay and session-key binding already have current tests in test_identity.py, test_grant_permissions.py and #243.

Closing this archived implementation without a replacement, not marking its code merged. Retain these acceptance cases for a future rotation adapter: duplicate/stale/conflicting announcements must preserve history; bare reannouncement must not replace a known key; verify the complete successor record; test historical signatures and honest cross-clock cases. The documented same-tick limitation needs an explicit causal/epoch contract before adopting a rotation protocol, rather than a universal sign-then-wait rule.

The original branch, tests and discussion remain available here. No legacy tests were rerun or rotation implementation adopted during this review.

PR #168, opened 2026-07-10
hackathon: embx_semantic memory plugin (Layer 10) -- recall by meaning

jonhardwick-spec:
## What this is

A new `embx_semantic` plugin for Layer 10 (Memory). The default `blackboard` is an 80-line shared dict -- read a value under a key, get it back under the same key, nothing else. If agent-A files a fact under `airport-departure-gate` and agent-B asks for `departure-gate-airport`, blackboard returns `None`. For a swarm that's the whole problem: agents name the same thing differently and the shared state goes unused.

This plugin implements the same `Memory` Protocol surface (`read` / `write` / `subscribe` / `cas`) so it slots into any scenario that currently names `blackboard` or `lww_register`, but every `write` also embeds the key and a new `semantic_lookup` ranks stored entries by **cosine similarity** to the query. A read for a *related* key finds the stored value even when the literal keys disagree.

## Why embx fits the Memory layer

The embedding backend is **embx** (https://embx.net) -- a real semantic vector memory service. `write` posts the key to embx's `/v1/embeddings` endpoint (`frankenstein-22.8mb-int8`, 384-dim) and stores the returned vector alongside the value; `semantic_lookup` embeds the query and ranks stored entries by cosine similarity. In production this is genuine semantic recall over a learned embedding space -- the thing blackboard categorically cannot do, because it has no notion of similarity.

embx is the right fit because the Memory layer's job is exactly "shared state agents can find later," and "find" in a multi-agent system means "find even when the asking agent used different words." Semantic recall is the feature the reference plugin is missing.

## Determinism -- why there's a fallback

Nanda Town's contract is byte-reproducible traces under a fixed seed. A live LLM embedding model is neither deterministic across hardware nor offline, so it can't be the path CI runs. The plugin has a deterministic fallback:

- When `EMBX_BASE_URL` is **unset** (the default) the plugin uses a hash-based bag-of-tokens pseudo-embedding. It's order-independent (keys with the same token multiset score cosine 1.0, partial overlap scores fractional, disjoint tokens score zero), pure-function-of-the-input, and offline. That's enough to prove the plugin ranks by similarity instead of exact-matching -- which is what the test asserts.
- When `EMBX_BASE_URL` is **set** (production), it posts to the live embx service. If the call fails for any reason (network, timeout, non-200, bad JSON), it trips a lazy flag and falls back without raising -- the agent loop keeps running.

The real semantic power shows in production; the fallback keeps CI reproducible. Same code path, different config.

## How to test it

```bash
# CI path (deterministic, offline -- the default):
make ci-local   # ruff, pyright strict, pytest

# Exercise the live embx path:
EMBX_BASE_URL=https://embx.net EMBX_API_KEY=embx_live_... uv run pytest \
    packages/nest-plugins-reference/tests/test_embx_semantic.py -v
```

Point a scenario at it by flipping one line in the YAML:

```yaml
layers:
  memory: embx_semantic   # was blackboard or lww_register
```

## The test that proves it

32 new tests. The adversarial pair is the one that matters:

- `test_blackboard_cannot_recall_by_related_key` -- blackboard returns `None` for a different-but-related key and has no `semantic_lookup` method at all.
- `test_embx_recalls_by_related_key` -- this plugin finds the stored value under a reordered key, via cosine similarity.

The first fails without the plugin (blackboard can't do it); the second passes with it. That's the "fails without your change, passes with it" bar.

## Files

- `packages/nest-plugins-reference/nest_plugins_reference/memory/embx_semantic.py` -- the plugin
- `packages/nest-plugins-reference/tests/test_embx_semantic.py` -- 32 tests, all forcing the deterministic fallback
- `packages/nest-plugins-reference/pyproject.toml` -- entry-point registration (same pattern as `lww_register`)
- `packages/nest-core/nest_core/plugins.py` -- builtin registration in `PluginRegistry`

`make ci-local` green: ruff, pyright strict, 1182 passed.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -45,6 +45,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("negotiation", "pareto"): f"{_REF}.negotiation.pareto:ParetoNegotiation",
     ("memory", "blackboard"): f"{_REF}.memory.blackboard:Blackboard",
     ("memory", "lww_register"): f"{_REF}.memory.lww_register:LwwRegisterMemory",
+    ("memory", "embx_semantic"): f"{_REF}.memory.embx_semantic:EmbxSemanticMemory",
     ("privacy", "noop"): f"{_REF}.privacy.noop:NoopPrivacy",
     ("privacy", "hybrid_x25519"): f"{_REF}.privacy.hybrid_x25519:HybridX25519Privacy",
     ("privacy", "trust_gated"): f"{_REF}.privacy.trust_gated:TrustGatedPrivacy",
--- a/packages/nest-plugins-reference/nest_plugins_reference/memory/embx_semantic.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/memory/embx_semantic.py
@@ -1,0 +1,489 @@
+# SPDX-License-Identifier: Apache-2.0
+"""embx-backed semantic memory plugin -- recall by meaning, not by key.
+
+The default ``blackboard`` memory plugin is an 80-line shared dict: write a
+value under a key, read it back under the *same* key. Nothing else. If agent-A
+files a fact under ``"airport-departure-gate"`` and agent-B asks for
+``"flight-boarding-info"``, blackboard returns ``None``. The keys have to match
+byte-for-byte or the fact is invisible. For a swarm that's the whole problem:
+different agents name the same thing differently, and the shared state goes
+unused.
+
+This plugin implements the same :class:`~nest_core.layers.memory.Memory`
+surface (``read`` / ``write`` / ``subscribe`` / ``cas``) but every write also
+embeds the key, and a new :meth:`semantic_lookup` ranks stored entries by
+**cosine similarity** to a query key -- so a read for a *related* key finds the
+stored value even when the literal keys disagree.
+
+The embedding backend is **embx** (https://embx.net), a real semantic-vector
+memory service. ``write`` posts the key to embx's ``/v1/embeddings`` endpoint
+and stores the returned vector alongside the value; ``semantic_lookup`` embeds
+the query and ranks stored entries by cosine similarity. In production this is
+genuine semantic recall over a 384-dimensional embedding space.
+
+Determinism -- why there is a fallback
+--------------------------------------
+
+Nanda Town's contract is byte-reproducible traces under a fixed seed. A live
+LLM embedding model is not deterministic across hardware and needs network, so
+it cannot be the path CI runs. This plugin therefore has a **deterministic
+fallback**: when ``EMBX_BASE_URL`` is unset, or the embx endpoint is
+unreachable, or the HTTP call fails for any reason, the plugin falls back to a
+hash-based pseudo-embedding. The fallback is deterministic (pure function of
+the key text) and is *deliberately semantic-ish*: it lowercases, tokenises on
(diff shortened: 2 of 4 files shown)


jonhardwick-spec (commented):
Have been waiting on approval?

Lowk submitted at 10ish 


--------
Didn't know about the prs not being reviewed ordeal 😭 

My bad guys 

I will now produce insane skill.md s and put in the skill spaces to create proper back propagation.

Very useful 🤷🏼‍♂️

shilpa-kulkarni-14:
@jonhardwick-spec  Reviewed on latest main. Competent, careful plugin, but I can't merge it as a
ranked contribution yet. ci-local is 5/5 green (1272 passed), Memory conformance is
clean (missing none), and the embx.net integration is nicely env-gated so nothing leaves
the process by default. Two things block it:

1. No discriminator. There's no validator and no scenario for this plugin, so nothing in
   the judged pipeline passes on embx_semantic and fails on a baseline. The recall-vs-
   blackboard proof lives only in unit tests. Add a scenario where agents name the same
   thing differently and semantic_lookup recalls the value while blackboard/lww_register
   (exact match) return None, plus a validator that fails on the baseline. That's the
   pattern #124 used for its CRDT.

2. The deterministic path is token overlap, not semantics. Real recall-by-meaning needs
   the opt-in embx path, which is non-deterministic and network-bound, so it can't be the
   CI/judge path. The hash fallback matches "departure-gate" and "gate-departures" because
   they share tokens, not because they mean the same thing; two same-meaning phrasings with
   no shared tokens score near zero. Either bundle a small deterministic embedding that
   captures real similarity, or reframe the deterministic claim honestly as token-overlap
   recall and keep true semantics as the opt-in path.

The env-gated external service design is good, keep that. Once there's a scenario-level
discriminator and the deterministic path actually demonstrates meaning rather than token
overlap, this is mergeable.

jonhardwick-spec:
Push incoming shortly ❤️

JamesCarnley:
Thanks @jonhardwick-spec for the embedding-backed recall contribution. I reviewed all four changed files at `877a2edd9e5656bac16c38bec86e822e17520c9a` and the discussion. Keeping exact-key `read()` intact while exposing ranked lookup separately, and making network access opt-in, are useful design choices.

Closing this legacy PR without a replacement in rebuilt Town. Its current memory implementation is per-agent exact-key remember/recall; it does not claim semantic lookup. An embedding service should be selected and tested through an explicit memory-service profile rather than silently becoming the harness's default memory backend.

The distinction raised in the earlier review still matters: the supplied successful recall tests exercise deterministic token overlap, not learned synonym/meaning retrieval, and there is no scenario-level semantic discriminator in this diff. The test fixture clears the live environment settings, so the advertised live test command does not prove a successful embx call. There is a deliberate unavailable-local-endpoint fallback test, but no successful live-service test.

Preserving the related-key/ranked-result and unrelated-query controls for a future profile, with backend/model identity and an independent expected result. That profile also needs a healthy-live-to-unavailable transition: existing stored vectors can remain in the learned space while later queries switch to hash vectors. Equal dimensions do not make those spaces comparable. Likewise a width mismatch currently returns a hash vector without switching `using_embx` off. Backend replacement must be visible as a separately labelled synthetic run or incomplete live evidence, not be counted as successful live semantic recall.

No embedding dependency or runtime was ported, no external service was invoked, and no replacement merge is claimed. The source and these credited requirements remain available when there is a concrete memory-service workload to test.

PR #170, opened 2026-07-10
feat(registry): add dp_bloom, a differentially private registry with a membership-inference validator

jadidbourbaki:
## Overview

This PR adds `dp_bloom`, a differentially private registry. It serves discovery exactly as the in-memory reference does and publishes a separate membership index that satisfies ε-differential privacy. An observer who reads that index cannot decide whether any single agent registered beyond an e^ε factor. The PR ships the plugin, an adversarial membership-inference validator that the reference registry fails, a scenario, and a test suite.

## Motivation

Every registry today answers who is registered and what each agent can do exactly. That behavior is correct for discovery and it leaks membership. Anyone who queries the registry or reads its published index recovers the full member set and every advertised capability. In many deployments the fact that an agent registered a capability such as `sell_medical_data` reveals more than any message the agent later encrypts. Problem 9 hardened message confidentiality. Nothing in the stack bounds membership confidentiality. This PR adds that surface to the registry and gives it a formal guarantee. I treat this as a new privacy angle on the registry building block rather than one of the ten written problems, which the Phase 1 brief permits when it invites contributors to improve a building block or add a new one.

## Testing Plan

The validator `check_membership_inference_bounded` runs a membership-inference experiment across the seed bank. For a fixed target it estimates the probability the index reports the target present when the target registered and when it did not. It then reports the empirical ε as the larger log-likelihood ratio across the two outcomes. A calibrated `dp_bloom` stays at or below its budget because the membership query is post-processing of an ε-DP index. The exact registry reports the target present exactly when the target registered, so its empirical ε is unbounded and it fails any finite budget. The test suite covers the ε and flip-probability calibration, exact lookups under privacy, byte-identical output under a fixed seed, seed sensitivity across seeds, and the pass or fail gate against both registries. All nine tests pass.

## Some Quick Analysis

Fix k hashes. Inserting one member sets at most k bits, so two registries whose members differ by one agent differ in at most k bit positions. Flipping each published bit independently with probability p gives a per-bit ε of ln((1-p)/p). Sequential composition over the at most k differing bits gives ε = k·ln((1-p)/p), where the unit of privacy is one agent's membership. Inverting for a target budget gives p = 1/(1+e^{ε/k}). The membership query reads only the private index, so post-processing preserves the bound.

The plugin draws the randomized-response coins once from a keyed PRF over a secret per-instance seed, following RAPPOR memoization. An adversary without the seed observes only the published bits and gains advantage at most e^ε. Fixing the seed fixes one draw of the coins and makes a Tier 1 trace reproducible without weakening the guarantee for a seedless adversary.

On identical inputs and seeds the validator measures the following.

| registry | empirical ε | budget plus slack | verdict |
|----------|-------------|-------------------|---------|
| `dp_bloom`, ε=3, k=3 | 2.40 | 4.0 | pass |
| `in_memory`, exact | 6.69 | 4.0 | fail |

## Limitations

The index is monotone. A Bloom filter cannot clear a bit it shares with another member, so `deregister` hides future lookups but leaves historical membership bits until the index rebuilds. The privacy unit is a single membership fact, so repeated registrations by one principal compose and fall outside this bound. A smaller ε raises the flip probability toward one half and increases membership-query error, which sets the usable privacy range.

## Success Criteria

The PR meets each mandatory deliverable in the charter. It adds a registry plugin registered as `("registry", "dp_bloom")`, an adversarial validator that passes the new plugin and fails the reference registry, a scenario under `scenarios/`, and a test suite. The five Definition of Done commands exit zero with 1159 passed and 1 skipped. The plugin reads no wall-clock time and uses no unseeded randomness, so traces stay deterministic across seeds.

## References

The randomized-response mechanism and the memoization that fixes each instance's coins once follow RAPPOR [2]. The membership-privacy goal and the differentially private Bloom filter construction follow Tirmazi [4], which builds on the adversarial Bloom filter model of Naor and Yogev [1] and its learned extension by Almashaqbeh, Bishop, and Tirmazi [3].

1. Moni Naor and Eylon Yogev. Bloom Filters in Adversarial Environments. 2014. [preprint](https://arxiv.org/abs/1412.8356)
2. Úlfar Erlingsson, Vasyl Pihur, and Aleksandra Korolova. RAPPOR: Randomized Aggregatable Privacy-Preserving Ordinal Response. ACM CCS, 2014. [preprint](https://research.google.com/pubs/archive/42852.pdf)
3. Ghada Almashaqbeh, Allison Bishop, and Hayder Tirmazi. Adversary Resilient Learned Bloom Filters. ASIACRYPT, 2025. [preprint](https://arxiv.org/abs/2409.06556)
4. Hayder Tirmazi. Adversarially Robust Bloom Filters: Privacy, Reductions, and Open Problems. 2025. [preprint](https://arxiv.org/abs/2501.15751)

## Disclosure On The Use of Generative AI

I built this submission with Claude Code running the Opus 4.8 model, and I want to be precise about the division of labor.

The idea and its theory are mine. The differentially private Bloom filter at the center of this plugin is my own research [3, 4], and the membership-privacy framing extends the adversarial Bloom filter line those papers sit in. I chose to target the registry layer, I decided that membership inference was the leak worth bounding, and I grounded the mechanism in RAPPOR and my prior work rather than in anything the model invented. The epsilon calibration, the choice of RAPPOR-style memoization to keep Tier 1 traces deterministic, and the design of the adversarial validator are decisions I own.

Claude Code did the first-pass implementation under my direction. It explored the nandatown codebase, drafted the plugin, validator, scenario, and tests against the design, and ran the Definition of Done gate. I set the writing standard and the citations. I read the implementation closely, hand-modified the code where the model got it wrong, and validated the behavior over multiple iterations before opening this PR.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -25,6 +25,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("identity", "ed25519_rotating"): (f"{_REF}.identity.ed25519_rotating:Ed25519RotatingIdentity"),
     ("registry", "in_memory"): f"{_REF}.registry.in_memory:InMemoryRegistry",
     ("registry", "gossip"): f"{_REF}.registry.gossip:GossipRegistry",
+    ("registry", "dp_bloom"): f"{_REF}.registry.dp_bloom:DPBloomRegistry",
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
     ("auth", "delegatable"): f"{_REF}.auth.delegatable:DelegatableAuth",
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
--- a/packages/nest-plugins-reference/nest_plugins_reference/registry/dp_bloom.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/registry/dp_bloom.py
@@ -58,13 +58,16 @@ filter model of Naor and Yogev [1] and its learned extension by Almashaqbeh,
 Bishop, and Tirmazi [3].
 
 [1] Moni Naor and Eylon Yogev. Bloom Filters in Adversarial Environments. 2014.
-    arXiv:1412.8356.
+    `preprint <https://arxiv.org/abs/1412.8356>`_
 [2] Úlfar Erlingsson, Vasyl Pihur, and Aleksandra Korolova. RAPPOR: Randomized
     Aggregatable Privacy-Preserving Ordinal Response. ACM CCS, 2014.
+    `preprint <https://research.google.com/pubs/archive/42852.pdf>`_
 [3] Ghada Almashaqbeh, Allison Bishop, and Hayder Tirmazi. Adversary Resilient
-    Learned Bloom Filters. ASIACRYPT, 2025. arXiv:2409.06556.
+    Learned Bloom Filters. ASIACRYPT, 2025.
+    `preprint <https://arxiv.org/abs/2409.06556>`_
 [4] Hayder Tirmazi. Adversarially Robust Bloom Filters: Privacy, Reductions, and
-    Open Problems. 2025. arXiv:2501.15751.
+    Open Problems. 2025.
+    `preprint <https://arxiv.org/abs/2501.15751>`_
 
 Example::
 
--- a/packages/nest-plugins-reference/nest_plugins_reference/registry/dp_bloom.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/registry/dp_bloom.py
@@ -1,0 +1,338 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Differentially private registry: membership queries under an epsilon bound.
+
+The default registry plugin
+:class:`~nest_plugins_reference.registry.in_memory.InMemoryRegistry` answers
+*"who is registered, and what can they do?"* from a plain ``dict``. That is
+functionally correct and **operationally a privacy leak**: anyone who can query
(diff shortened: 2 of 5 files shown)


JamesCarnley:
Thanks @jadidbourbaki for separating exact lookup from a published membership view and making privacy/utility an explicit test question. I reviewed all five changed files at `80fe3ff` against rebuilt Town `main@f3664a8`.

Closing this archived registry implementation without adopting a new default discovery policy. Current Town's local index serves exact capability lookup; no observer-only private-membership contract or privacy evaluator has been selected. This remains useful protocol research, not a substitute for testing the actual NANDA discovery path.

Preserve the neighboring-world experiment and exact-lookup controls, with these adoption requirements:
- Specify which observers can access exact lookup/subscriptions versus only the noisy index, and establish the seed's lifecycle. The plugin itself has no caller-access separation and defaults to an empty seed.
- Define the privacy unit and bound contributions. The implementation inserts an agent token plus every capability token, while the stated calibration assumes at most `k` changed bits per agent. The present test uses a target whose capability already exists in the background.
- Analyze repeated publications across changing registry states, not just repeated reads of one snapshot, together with updates, deregistration residue and rebuilds.
- Require nonempty statistical samples, uncertainty and utility measurements. A fixed 400-seed binary-query experiment plus a chosen slack is diagnostic evidence, not certification of the full published-index mechanism.

The YAML runs a marketplace; this diff does not connect the new membership validator to that runner's verdict. The genuine comparison is in the standalone unit-test oracles. Keep that execution boundary explicit.

The contribution and follow-up criteria remain here. No privacy guarantee, real deployment or legacy test rerun is claimed by this disposition.

PR #174, opened 2026-07-10
build: migrate deprecated tool.uv.dev-dependencies to dependency-groups.dev

mgpiller:
uv prints a deprecation warning on every command: tool.uv.dev-dependencies
is deprecated and will be removed in a future release.

This moves the dev dependency list to the standard dependency-groups.dev
table (PEP 735). Same packages, same version pins, same comments.

uv sync now runs clean. ruff and the full test suite pass unchanged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,8 +66,8 @@ members = [
     "packages/nest-marketplace",
 ]
 
-[tool.uv]
-dev-dependencies = [
+[dependency-groups]
+dev = [
     "pytest>=8.0",
     "pytest-asyncio>=0.24",
     "hypothesis>=6.100",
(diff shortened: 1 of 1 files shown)


JamesCarnley:
Thanks @mgpiller for the packaging cleanup. I reviewed the full patch against rebuilt main at `d26ca5a`.

The active [pyproject.toml](https://github.com/projnanda/nandatown/blob/d26ca5a51dd532418bca369e84abf9a2137b6927/pyproject.toml) uses setuptools and a `project.optional-dependencies.dev` extra. It has neither the old uv workspace layout nor `tool.uv.dev-dependencies`. Current CI installs with `pip install -e ".[dev]"`. I also parsed the current TOML and verified those facts directly.

The specific deprecation this PR addresses is therefore absent from current main. Closing as obsolete build maintenance without a replacement PR, not as merged code. The patch remains available for users of the archived workspace.

PR #176, opened 2026-07-10
[Hackathon] rahulsharma-sys: sealedbox — nonce-misuse-resistant hybrid privacy (Problem 09)

rahulsharma-sys:
## Summary

Reworked **`sealedbox`** privacy plugin (Problem 09). This supersedes #36, which
GitHub auto-closed when its branch was renamed to match the charter's
`hackathon/<handle>-<theme>` pattern (`rahul` → `rahulsharma-sys`); same author,
same work, all review feedback addressed.

The default `noop` privacy plugin is a passthrough. `sealedbox` provides a full
Problem 09 solution — hybrid X25519 + ChaCha20-Poly1305 encryption, salted-Merkle
selective disclosure, and epoch-based broadcast revocation — with one
**load-bearing differentiator** vs the merged `hybrid_x25519`: deterministic mode
is **nonce-misuse resistant (SIV-style)**.

### Why the differentiator matters

Tier-1 traces must replay byte-for-byte, so a privacy plugin derives per-message
keys/nonces deterministically. Every such scheme guards the derivation with a
monotonic counter and concedes that a `(key, nonce)` collision would be
catastrophic. But that counter lives in memory: reconstruct the plugin (restart,
or two instances of one agent) and it resets to zero, re-deriving the same key
and nonce for a different plaintext — a two-time pad. `hybrid_x25519`'s own
docstring states its deterministic mode is safe *"only because the counter never
repeats."*

`sealedbox` folds a digest of `(plaintext, sorted audience)` into the HKDF `info`
for the content key, content nonce, ephemeral scalars, and wrap nonces (the
synthetic-IV construction behind AES-GCM-SIV). A counter collision then degrades
to revealing *message equality* instead of a two-time pad. This is proven live:
the adversarial validator `check_deterministic_reuse_safe` **fails** against
`noop` and against `hybrid_x25519`'s reconstructed deterministic mode, and
**passes** against `sealedbox`.

### Problem 09 surface

- **Hybrid encryption**: fresh ChaCha20-Poly1305 content key per message, wrapped
  per recipient via X25519 ephemeral-static ECDH + HKDF-SHA256. The envelope
  header (version/sender/epoch/counter/sorted recipient key-ids) is bound as AEAD
  associated data, so redirection or header edits break authentication.
- **Selective disclosure**: salted-Merkle `prove`/`verify_proof` (a `reveal` JSON
  array selects fields to disclose; the rest stay hidden behind the issuer root).
- **Broadcast revocation**: `revoke` advances an epoch and excludes a member from
  future wraps; forward secrecy is explicitly future-only (documented).
- **Replay rejection**: keyed on the envelope digest, so a reconstructed sender's
  distinct messages are not mistaken for replays of each other.
- **Adversarial validators** (`validators/sealedbox_validators.py`): the four
  Problem 09 attacks (eavesdropper, replay, field-injection, stale-revocation) —
  each fails against `noop` and passes against `sealedbox` — plus the novel
  two-time-pad detector.
- **Scenario** `scenarios/sealed_bid_sealedbox.yaml` (the canonical
  `sealed_bid_with_privacy.yaml` filename is owned by the merged `hybrid_x25519`
  PR, so this ships the parallel variant and does not modify it). Its prose is
  explicit that the Tier-1 auction task does not route payloads through the
  privacy layer, so the crypto is exercised in `tests/test_sealedbox.py`, not on
  the wire.

Registered in both `_BUILTINS` and the `nest.plugins.privacy` entry point.

### Honesty notes

Deterministic mode is protocol/trace-level confidentiality, **not** secrecy
against a holder of the shared seed (who can re-derive any agent's key);
`deterministic=False` gives per-agent random keys for genuine secrecy at the cost
of reproducibility. This is stated in the docstrings.

## Test Plan

`make ci-local` passes — ruff, ruff format --check, pyright (strict), pytest
(**764 passed, 1 skipped**). 30 sealedbox tests cover encryption round-trips,
multi-recipient, tamper + header-binding, malformed envelopes, replay,
determinism, the nonce-misuse-resistance property (incl. a live comparison that
the validator catches `hybrid_x25519` deterministic mode failing), random mode,
selective disclosure (reveal/verify, field-injection, wrong-root, inconsistent
witness, comma-in-field-name), revocation (future-only, idempotent re-revoke),
the four Problem 09 validators against `sealedbox` and `noop`, and registry
discovery.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -36,6 +36,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("memory", "blackboard"): f"{_REF}.memory.blackboard:Blackboard",
     ("memory", "lww_register"): f"{_REF}.memory.lww_register:LwwRegisterMemory",
     ("privacy", "noop"): f"{_REF}.privacy.noop:NoopPrivacy",
+    ("privacy", "sealedbox"): f"{_REF}.privacy.sealedbox:SealedBoxPrivacy",
     ("datafacts", "datafacts_v1"): f"{_REF}.datafacts.datafacts_v1:DataFactsV1",
 }
 
--- a/packages/nest-plugins-reference/nest_plugins_reference/privacy/sealedbox.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/privacy/sealedbox.py
@@ -1,36 +1,98 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Real authenticated encryption for the Privacy layer (a sealed-box alternative
-to the ``noop`` passthrough).
+"""Nonce-misuse-resistant hybrid encryption for the Privacy layer.
 
-Multi-recipient sealed-box encryption: a fresh ChaCha20-Poly1305 content key per
-message, wrapped to each recipient via X25519 ECDH + HKDF-SHA256. This gives the
-Privacy layer genuine confidentiality (non-audience agents cannot read a payload
-through the API; the trace carries only ciphertext) and tamper-evidence (a
-modified ciphertext fails to decrypt rather than silently yielding altered bytes).
+``sealedbox`` is a full Problem 09 privacy plugin (hybrid encryption + selective
+disclosure + broadcast revocation), but its **load-bearing contribution** is a
+property the merged ``hybrid_x25519`` plugin openly concedes it lacks:
+**deterministic mode that is safe under key/nonce reuse.**
 
-Keys derive deterministically from ``(seed, agent_id)`` by default so simulation
-traces replay byte-for-byte, consistent with NandaTown's replay-first identity
-model. That is protocol/trace-level confidentiality, *not* secrecy against a
-holder of the shared seed. Pass ``deterministic=False`` for random per-agent keys
-(genuine secrecy against a seed-holder, at the cost of non-reproducible traces).
+Why that matters
+----------------
 
-``prove`` / ``verify_proof`` are honest stubs that raise ``NotImplementedError``
-— unlike the ``noop`` plugin's forged proof that always verifies. Zero-knowledge
-proofs are future work.
+Nanda Town's Tier-1 traces must replay byte-for-byte, so a privacy plugin that
+wants to run in a scenario derives its per-message key and nonce deterministically
+rather than from the system RNG. Every such scheme (including ``hybrid_x25519``)
+guards the derivation with a monotonic counter and states — correctly — that
+reusing a ``(key, nonce)`` pair would be catastrophic for the AEAD. But that
+counter lives in memory on the plugin instance: reconstruct the plugin (a
+restart, or two instances of one agent in a scenario) and the counter resets to
+zero, re-deriving the *same* key and nonce for a *different* plaintext. For a
+ChaCha20-Poly1305 stream that is a two-time pad — it leaks the XOR of the two
+plaintexts and destroys authentication. Problem 09 names this exact anti-pattern:
+*"Don't reuse the same nonce twice (test for it)."*
+
+What sealedbox does differently (SIV-style derivation)
+------------------------------------------------------
+
+In deterministic mode the content key, the content nonce, every ephemeral scalar
+and every wrap nonce are derived from a **synthetic value** that folds a digest
+of the *plaintext and the sorted audience* into the HKDF ``info`` (the
+synthetic-IV construction behind AES-GCM-SIV / deterministic AEAD). So a counter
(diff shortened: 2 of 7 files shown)


rahulsharma-sys:
Responding to the review on #36 (this PR supersedes it after the branch rename). Reworked to address all four points; `make ci-local` is green (ruff, format, pyright --strict, pytest 764 passed / 1 skipped). @dhve

**1 + the duplication charge.** `sealedbox` now solves Problem 09 in full, and the differentiator vs `hybrid_x25519` is load-bearing rather than cosmetic: deterministic mode is **nonce-misuse resistant (SIV-style)**. The content key, content nonce, ephemeral scalars and wrap nonces derive from a digest of `(plaintext, sorted audience)`, so the counter collision you flagged in point 3 — a reconstructed instance resetting its counter — degrades to revealing message equality instead of a two-time pad. `hybrid_x25519`'s docstring concedes its deterministic mode is safe *"only because the counter never repeats"*; `sealedbox` removes that assumption. Proven live: `check_deterministic_reuse_safe` **fails** against `hybrid_x25519` deterministic mode (reconstructed) and against `noop`, and **passes** against `sealedbox`.

**2. Mandatory deliverables.** Added `validators/sealedbox_validators.py` — the four required attacks plus the two-time-pad detector — and `scenarios/sealed_bid_sealedbox.yaml`. The canonical `sealed_bid_with_privacy.yaml` filename is owned by your merged #28, so this ships the parallel name and doesn't touch yours. The scenario prose is explicit that the Tier-1 auction task doesn't route payloads through the privacy layer, so the crypto is exercised in the unit tests, not on the wire.

**3. Nonce reuse.** Fixed via the SIV derivation, and the envelope header (version/sender/epoch/counter/sorted recipient key-ids) is now bound as AEAD associated data instead of `None`. Replay memory is keyed on the envelope digest so a reconstructed sender's distinct messages aren't false replays; a regression test asserts no two-time pad across reconstruction.

**4. Registration.** Added the `nest.plugins.privacy` entry point in `packages/nest-plugins-reference/pyproject.toml`.

Also: real salted-Merkle selective disclosure replaces the `NotImplementedError` stubs; epoch revocation with an explicit future-only forward-secrecy note; and the docstrings state plainly that deterministic mode is protocol/trace-level confidentiality, not secrecy against a shared-seed holder (`deterministic=False` for that). Branch renamed to `hackathon/rahulsharma-sys-privacy-sealedbox` per your point 4.

JamesCarnley:
Thanks @rahulsharma-sys for this revision and its detailed scope notes. I reviewed final head `31ebe1101772501f9602af2172fb2a19e60eac49`, all seven changed files and the follow-up comment against current main `2c5fc4337de99fbbefe6b8cf8e1db2813a08ea1e`.

**Disposition: close this legacy implementation without a replacement merge; preserve its tests as requirements for an optional encrypted-transport/selective-disclosure profile.** Current Town's `privacy/redact.v1` removes declared fields from exported records. It does not implement encryption, disclosure proofs or broadcast membership, and is not a substitute for this work.

The most useful retained requirements are:

- Exercise sender reconstruction and recipient restart explicitly, in addition to normal round trips. Reproducible fixture keys, real confidentiality, nonce management and durable replay protection need separate claims.
- Check authorized recipients as healthy controls alongside outsiders, modified envelopes, revoked recipients and unaffected members. Bind results to the actual transport and key-directory configuration used.
- Preserve selective-disclosure tests for the requested field set, altered fields, incorrect roots and inconsistent witnesses. Root membership must remain separate from who is authorized to issue or endorse that root.
- Keep future-message revocation distinct from secrecy of previously delivered traffic. Preserve the documented limits of shared-seed mode and in-memory replay state.
- Require the complete selected workflow to traverse the privacy implementation before reporting encrypted-delivery coverage. This PR's YAML and simulator test correctly disclose that the auction only resolves the plugin and does not seal its bids; the direct unit tests are a separate evidence tier.
- Distinguish expected rejection from an implementation/runtime error in portable validators. A broad caught exception or a plaintext-substring check alone should not be reported as proof of confidentiality.

A current-main integration would first need a pinned protocol/library, explicit key-directory and restart-state ownership, actual transport wiring, and review appropriate to that cryptographic design. This cleanup is not adopting or certifying the custom scheme, and the retained profile is **deferred, not implemented**. The source remains available here, including its connection to #36, for that future work.

Fresh current-code verification: `tests/test_layers.py tests/test_lab_coverage.py tests/test_bundle.py` produced **38 passed**. That verifies Town's existing redaction/coverage/bundle behavior only, not this legacy plugin. No original branch merge or deletion is being claimed.

PR #178, opened 2026-07-10
Fix reputation scoring validator dead loop

### LINKED ISSUE #98: validate_reputation_scoring: documented 'cheaters lose score' property is never enforced (dead loop)

RoarHX:
Fixes #98

## Summary

Fixes the reputation scoring validator so it enforces a concrete invariant instead of accumulating scores and leaving the validation loop empty.

## Changes

- Distinguishes explicit `cheat:` events from agents who merely receive a bad report
- Requires every explicit cheating event to have a corresponding bad reputation report
- Removes the dead score-tracking loop
- Adds a regression test proving that a bad report alone does not count as cheating

## Testing

- `uv run pytest packages/nest-core/tests/test_validators.py::TestReputationScoring -v`
- 4 tests passed

### DIFF
--- a/packages/nest-core/nest_core/validators.py
+++ b/packages/nest-core/nest_core/validators.py
@@ -797,62 +797,44 @@ def validate_supply_chain_no_lost(
 def validate_reputation_scoring(
     events: list[dict[str, Any]],
 ) -> list[ValidationResult]:
-    """Reputation scores decrease for agents that cheat."""
-    # Track scores: agent -> score
-    scores: dict[str, int] = defaultdict(int)
-    cheaters: set[str] = set()
+    """Every explicit cheating event must result in a bad reputation report."""
+    reported_bad: set[str] = set()
+    explicit_cheaters: set[str] = set()
 
     for ev in events:
         if ev.get("kind") != "send":
             continue
+
         msg = _message_body(ev)
+
         if msg.startswith("report:"):
             parts = msg.split(":")
-            if len(parts) >= 4:
-                agent_str = parts[2]
-                outcome = parts[3]
-                if outcome == "good":
-                    scores[agent_str] += 1
-                elif outcome == "bad":
-                    scores[agent_str] -= 2
-                    cheaters.add(agent_str)
+            if len(parts) >= 4 and parts[3] == "bad":
+                reported_bad.add(parts[2])
 
-    violations: list[str] = []
-    for cheater in cheaters:
-        if scores[cheater] >= 0:
-            # If a cheater has never had their score go negative from cheating
-            # that's fine — they might have enough good trades.  We check that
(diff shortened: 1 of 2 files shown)


JamesCarnley:
Thanks @RoarHX for making the distinction between an explicit cheating observation and merely receiving a bad reputation report.

#242 is merged as `f4e85d75324a09ca9820e15d36281345c70522b6`. Its landed tree matches the reviewed candidate: 303 tests passed locally and Python 3.11/3.12 CI passed.

Reviewed the full patch at `0a791a4d705cd94ce33e0b378b64dad334925cad`. The legacy parser/validator is retired. Replacement #242 addresses the current evaluator's related gap: a final score of 2 could pass even with missing receipts, inconsistent attribution or an incorrect intermediate score. It now checks prior receipt-event correlation and reference arithmetic, cites the receipt evidence, and explicitly says that claim truth and reporter authority are not tested.

Closing as **partially superseded by #242**, not as a merge of the legacy implementation. The stronger observed-misconduct-to-report requirement remains unimplemented: it needs a selected adversarial workflow with transaction-level correlation and independent behavior observations. A signed or attributed bad report alone must not label an agent a proven cheater. This future requirement and your regression rationale remain preserved here rather than being represented as delivered by the consistency fix.

PR #179, opened 2026-07-10
[Hackathon] fix(nest-plugins-reference): register missing reference plugin entry …

shreyad2806:
The reference package only exposed a handful of plugins as entry points (memory/lww_register, privacy/hybrid_x25519, datafacts/cid_facts, and the failure detectors). The remaining plugins were only reachable through the built-in fallback in nest-core, which is not the documented discovery path.

Add entry-point declarations for every plugin listed in the built-in fallback so that the PluginRegistry discovers them via the standard install-time mechanism documented in CONTRIBUTING.md.

Also add a regression test in nest-core that fails if a built-in plugin is not backed by an entry point, preventing the two registries from drifting out of sync again.


shilpa-kulkarni-14:
@shreyad2806 plz resolve conflicts

shreyad2806:
Fixed the conflicts

JamesCarnley:
Thanks @shreyad2806 for the registry-parity regression test. I reviewed both changed files and the discussion against rebuilt main at `d26ca5a`.

The old split between `nest.plugins.*` packaging entry points and the built-in fallback registry is no longer present. Current [layer discovery](https://github.com/projnanda/nandatown/blob/d26ca5a51dd532418bca369e84abf9a2137b6927/src/nandatown/layers/__init__.py) uses one decorator registry. Existing tests verify default-layer resolution and explicit failure for unknown plugins; those checks passed in this review.

Closing without a replacement because the dual-inventory mechanism this patch synchronizes has been retired. This does not claim that all legacy plugins were ported. The useful principle remains: if package-entry-point discovery is introduced later, test parity against the exact registry exposed to users. That future boundary should not be added solely to transplant this maintenance patch. Original code and discussion remain preserved.

PR #180, opened 2026-07-10
memory: add OR-Set (observed-remove set) CRDT plugin

ChayaNine:
Adds a state-based OR-Set CvRDT to the memory layer, registered as memory/or_set. Concurrent adds to a key converge to the union (add-wins), unlike lww_register which reduces concurrent writes to a single winner. Merge is commutative, associative, and idempotent; property tests and the shared validate_crdt_convergence harness verify strong eventual consistency. Registered in both _BUILTINS and the nest.plugins.memory entry points.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -45,6 +45,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("negotiation", "pareto"): f"{_REF}.negotiation.pareto:ParetoNegotiation",
     ("memory", "blackboard"): f"{_REF}.memory.blackboard:Blackboard",
     ("memory", "lww_register"): f"{_REF}.memory.lww_register:LwwRegisterMemory",
+    ("memory", "or_set"): f"{_REF}.memory.or_set:OrSetMemory",
     ("privacy", "noop"): f"{_REF}.privacy.noop:NoopPrivacy",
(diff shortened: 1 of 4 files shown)


JamesCarnley:
Thanks @ChayaNine for the explicit set semantics and property-based convergence tests. This is not simply a duplicate of #192: here `write` adds a member, reads always encode a member list, and compare-and-set conditionally adds rather than replacing the observed set.

After reviewing all four changed files against rebuilt Town `main@f3664a8`, I am closing the legacy implementation without a replacement. Current `kv.v1` provides per-agent local remember/recall and key-only events, not a shared replicated-memory contract. Automatically substituting this set API would change existing user-visible semantics, and neither this PR nor #192 has been selected as that new contract.

Retain for a future replicated-memory test profile: concurrent add versus observed remove, unset versus empty values, full-state round-trip, duplicate/reordered delivery, algebraic merge laws and an independently checked expected member set. Additional acceptance requirements are durable unique replica/tag allocation across restart, malformed-state rejection without partial mutation, explicit local versus distributed CAS semantics, and declared subscriber/state/tombstone bounds.

The code's local counter starts from zero and is not restored by merge; complete-snapshot merges also apply keys one by one. Those are important restart and failure-atomicity tests before using it beyond the supplied model. LWW and OR-Set are different contracts, so retaining all concurrent writes is not a universal correctness rule for every memory implementation.

No original code was run or merged. This contribution remains a concrete optional set-valued profile/reference, with #192 preserving the separate register-like alternative.

PR #185, opened 2026-07-10
hackathon/john-kioko-weighted-trust: WeightedTrust plugin (v2 — discriminator + determinism + CI)

Mutinda-Kioko:
## WeightedTrust — recency-decayed, volume-weighted reputation (v2)

Addresses all feedback from the v1 review on #143.

---

### Reviewer feedback → fixes

**1. Non-vacuous discriminator** ✅
- Added `validate_stale_report_attacker_cheated` + `validate_stale_report_temporal_split` validators, registered in `VALIDATORS["stale_report"]`
- Added `stale_report` scenario with `HonestTrader`, `ReformedAttacker`, `TrustObserver` agent brains — the attacker behaves well for 6 rounds, then cheats for 4
- `test_stale_report_scenario.py` proves **WeightedTrust produces a lower score than ScoreAverageTrust** for the same stale-positive-then-negative evidence:
  - `score_average`: score = **0.6** (6 positives + 4 negatives → 6/10 — deceptively trusted)
  - `WeightedTrust`: score **< 0.5** (old positives decayed, recent negatives dominate)
- The core claim — that weighting beats naive averaging on stale evidence — is now **demonstrated**, not just asserted

**2. Wall-clock removed — fully deterministic** ✅
- Replaced `time.time()` with caller-supplied `set_tick(t)` + `evidence.timestamp`
- `score()` is now a **pure function** of evidence list + current tick — never reads the system clock
- Verified determinism:
  - `test_identical_scores_across_instances` — two independent instances, same inputs → identical scores
  - `test_seed_independence` — WeightedTrust uses no RNG; scores are identical across "seeds" 42, 7, 1337

**3. ci-local cleared** ✅
- `ruff check`: ALL PASS (I001 import ordering fixed)
- `ruff format`: applied to all changed files
- `pyright`: no `reportPrivateUsage` — tests use **public API only** (`score`, `report`, `attest`, `stake`, `set_tick`)

**4. Novelty vs #139** ✅
- PR #139 is `SybilResistantTrust`: 2-pass eigen-weighting for sybil/collusion defense
- WeightedTrust's **distinct** capability is **recency decay** — stale evidence loses weight exponentially. Neither `score_average`, `agent_receipts`, nor `sybil_resistant` provide this
- Dropped reporter-credibility tracking to eliminate overlap

---

### Files

**New:**
- `nest_plugins_reference/trust/weighted.py` — the plugin (deterministic, tick-based)
- `tests/test_weighted_trust.py` — 11 conformance + behaviour + determinism tests
- `tests/test_stale_report_scenario.py` — 4 discriminator tests proving weighted < average
- `scenarios/stale_report.yaml` — scenario YAML
- `nest_core/scenarios_builtin/stale_report.py` — agent brains (HonestTrader, ReformedAttacker, TrustObserver)

**Modified:**
- `nest_core/plugins.py` — registered `("trust", "weighted")`
- `nest_core/scenarios.py` — registered `"stale_report"` scenario type
- `nest_core/validators.py` — added stale_report validators

### Test results
```
52 passed in 0.22s (15 new + 37 existing, zero regressions)
ruff check: All checks passed
ruff format: clean
```

Rebased on latest main (eb58b7b).

Author: **John Kioko** (@Mutinda-Kioko)


JamesCarnley:
Thanks @Mutinda-Kioko for replacing wall-clock reads with an explicit logical tick and adding a concrete stale-positive/recent-negative comparison. That is a distinct experiment from reporter admission or collusion resistance.

I reviewed all eight changed files against rebuilt Town `main@f3664a8`. Current reputation uses an explicit cumulative receipt formula; choosing recency decay, severity weights and a trust threshold is a separate policy rather than a missing implementation of that formula. Closing the legacy implementation without a replacement, with the temporal experiment preserved here.

The comparison tests do feed the same synthetic evidence to both algorithms. However, the `stale_report` factory ignores the selected trust plugin and the observer only broadcasts report messages. Its validators check report presence and temporal ordering, not actual score updates or whether those reports match independently observed behavior. The file named as an end-to-end scenario test does not boot that scenario. Those boundaries need to be connected before claiming a full-path result.

For a future optional profile, retain the fixed-clock comparison and require observed task outcome → attributed report → actual score event → independent recomputation. Include honest recovery as well as deteriorating behavior, duplicate/reordered evidence, declared time/freshness rules and unknown outcomes. Keep evidence freshness/effective weight separate from raw sample-count confidence: with only old positive evidence, this normalized formula can remain at 1.0, as its own test correctly notes.

No original code was executed or merged; no decay rate, score threshold or global trust policy is adopted by this closure.

PR #187, opened 2026-07-11
fix(lint): E501 on main — wrap 103-char test signature from #184

KaranSinghBisht:
`make ci-local` currently fails on main itself: `ruff check` reports E501 (103 > 100) at `packages/nest-plugins-reference/tests/test_negotiation_determinism.py:88`, introduced in #184. Every open hackathon branch rebased onto main inherits the red gate.

One-line wrap of the test signature, identical to `ruff format` output.

Verified: `ruff check` clean, `ruff format --check` clean, `pytest packages/nest-plugins-reference/tests/test_negotiation_determinism.py` 8/8 pass.

### DIFF
--- a/packages/nest-plugins-reference/tests/test_negotiation_determinism.py
+++ b/packages/nest-plugins-reference/tests/test_negotiation_determinism.py
@@ -85,7 +85,9 @@ class TestDeriveSessionId:
         partner=st.text(min_size=1, max_size=16),
         seq=st.integers(min_value=0, max_value=1_000_000),
     )
-    def test_deterministic_over_arbitrary_inputs(self, initiator: str, partner: str, seq: int) -> None:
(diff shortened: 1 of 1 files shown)


JamesCarnley:
Thanks @KaranSinghBisht for fixing the inherited lint failure. I reviewed the full patch against rebuilt main at `d26ca5a`.

This change only wraps a test signature in `packages/nest-plugins-reference/tests/test_negotiation_determinism.py`. That package and test path are no longer part of current main, so there is no corresponding line-length fix to apply to the active codebase.

Closing as obsolete maintenance for the archived implementation, without a replacement PR. This is not a rejection of the original fix or a claim that it was merged. The original branch and discussion remain available for anyone maintaining `archive/legacy`.

PR #188, opened 2026-07-11
docs(registry): capability-conformance plugin note for #159

dhritimandas:
## Problem

#159 (capability-conformance registry gate, merged) shipped the
`verified_capabilities` plugin, validator, and scenario without a
`docs/layers/registry.md` entry, unlike #148's `byzantine_gossip` section.
The merged commit also carried a `Co-Authored-By: Claude Sonnet 5` footer,
per reviewer feedback that should not recur in follow-up commits.

## Change

- Adds a `## Capability-conformance plugin: verified_capabilities`
  subsection to `docs/layers/registry.md`, mirroring #148's
  `byzantine_gossip` section: threat model, mechanism (bootstrap allowance,
  per-capability exclusion, re-admission), validator description, and demo
  scenario with the literal FAIL/PASS output and the exact test commands.
- No code changes; no other doc sections touched.

## Test

`make ci-local`: 1282 passed, 1 skipped, 1 deselected, all 5 checks green.

## Fit

Same documentation pattern #148 already established for a hardened registry
plugin variant, applied to `verified_capabilities`.

## Note

Includes a one-line lint fix for a pre-existing ruff failure on main
introduced by #184 (test_negotiation_determinism.py:88), isolated in its
own commit so the docs change stays clean.

### DIFF
--- a/docs/layers/registry.md
+++ b/docs/layers/registry.md
@@ -117,6 +117,91 @@ See
 for the full FAIL/PASS matrix (validators x plugins x scenarios) and every
 honest limitation found while building this plugin.
 
+## Capability-conformance plugin: `verified_capabilities`
+
+`verified_capabilities` -- closes a gap none of the identity-focused registry
+plugins touch. `AgentCard.capabilities` is a bare, self-asserted `list[str]`;
+`ed25519_rotating` and `byzantine_gossip` verify who published a card and that
+its content wasn't forged in transit, and `agent_receipts`/`score_average`
+score an agent's reputation *overall* -- but nothing checks a specific
+capability claim against whether the agent has ever actually fulfilled it. An
+agent can register `capabilities=["sell"]` with a perfectly valid signature
+and stay discoverable via that claim forever, even if it never completes a
+single sale. This is the exact gap A2A v1.0's signed Agent Cards and NANDA's
+Verified AgentFacts layer both name and neither closes: a signature proves
+the publisher, not the capability.
+
+`verified_capabilities` wraps an inner registry and tracks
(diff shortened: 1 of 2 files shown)


JamesCarnley:
Thanks @dhritimandas. The distinction here is important: verifying the publisher of a capability claim does not establish that the capability actually works, and evidence should stay scoped to the particular agent and capability.

I reviewed the full documentation/lint patch at `c391d9b01a121e148b8e041a28705206d2896c5a` against rebuilt `main@f4e85d7`. #159 was historically merged (`eb58b7b3f99744d2e25ac9eef46d3d16d138969c`), but its `verified_capabilities` plugin, scenario, validator and command paths are not in current main. The incidental negotiation-test formatting target is retired too. Current `TownIndex` verifies publication and ranks matches; the Path runner separately evaluates descriptor consistency and task fulfillment. Neither implements this per-capability exclusion/readmission policy.

Closing as **obsolete implementation-specific documentation, without a replacement**, not as a merge or a claim that the capability-conformance work is complete. Publishing these commands in the rebuilt tree would document an unavailable plugin.

The useful next experiment remains preserved here: pin an agent/capability/release, observe a successful task followed by a distinct later request that fails, retain the time-scoped observations, and verify that another capability is not mislabeled. That differs from the current duplicate-request check. Any subsequent discovery filtering needs explicit evidence admission, an untested/unavailable state, a window and recovery policy, plus controls showing honest agents remain discoverable. The timeout-to-defection and single-success reset rules should be comparative policies under test, not silently adopted as universal trust rules. The original PR and #159 retain the implementation and tradeoff discussion for that future profile.

PR #189, opened 2026-07-11
[Hackathon] Redeem_Grimm: extract reusable heartbeat authenticator, make the forgery discriminator a component swap

Redeem-Grimm-Satoshi:
## Problem

This is a direct follow-up to the review note on #152. That review approved the signed-heartbeat forgery defense but flagged two structural points: the authenticated-heartbeat logic lived inside the `failure_detection` scenario builtin rather than a reusable component, and the forgery discriminator was a `verify_heartbeats` boolean toggle rather than a swapped component like the rest of the framework uses. Both are fair. No other scenario could reuse the authentication without copy-pasting it, and the "does authentication matter?" question was answered by a config flag instead of by swapping an implementation, which is the pattern everything else here follows (a scenario picks `phi_accrual` or `heartbeat` for the detector, `signed` or `noop` for privacy, and so on).

This PR extracts the authentication into a reusable module and turns the discriminator into a component swap, with no change to observable behavior.

## What is in the diff

| Path | Purpose |
|---|---|
| `packages/nest-plugins-reference/nest_plugins_reference/failure_detection/heartbeat_auth.py` | New reusable module, co-located with the detector plugins and following the same `*_wire` pattern as `coordination/hotstuff_wire.py`. It holds the wire format (`heartbeat_payload`, `claimed_peer`, `HB_PREFIX`), a `HeartbeatAuthenticator` protocol, two verifiers, and a `make_heartbeat_authenticator` selector. |
| `packages/nest-core/nest_core/scenarios_builtin/failure_detection.py` | The scenario now imports the wire format and verifiers from the reusable module instead of defining them. `FailureMonitorAgent` holds an injected authenticator and has a single heartbeat path, so it no longer branches on a verify flag. The factory selects the authenticator by name. |
| `packages/nest-plugins-reference/tests/test_failure_detection.py` | Component unit tests for both verifiers and the selector, plus an end-to-end component-swap test and a test that the new selector and the legacy alias produce byte-identical traces. |
| `packages/nest-plugins-reference/tests/test_negotiation_determinism.py` | Wraps one pre-existing over-length line. See the note at the end. |

## The reusable component

`heartbeat_auth.py` packages three things any scenario can now reuse:

- The wire format `FDHB|<id>|<ts>|<sig-hex>`, unchanged, with the signature reconstructed from the raw wire text so traces stay deterministic.
- `SignedHeartbeatAuthenticator`: accepts a beat only if the signature verifies against the claimed peer's registered key and the timestamp is fresh (strictly newer than the last accepted, and not from the future). It owns its per-peer last-accepted timestamp internally, so the replay state is encapsulated rather than threaded through the caller.
- `TrustingHeartbeatAuthenticator`: the naive baseline that believes the claimed id. This is the real foil, the behavior a spoofable observer exhibits.

Both satisfy a small `HeartbeatAuthenticator` protocol (`accept(payload, *, now) -> (peer, ts) | None`, plus `name` and `verifies`), checked structurally at import the same way the detector plugins are.

## The discriminator is now a component swap

The observer no longer has a verify-or-trust branch. It holds one `HeartbeatAuthenticator` and feeds the detector whatever that authenticator accepts. Turning authentication on or off is selecting the component:

```yaml
task:
  config:
    hb_auth: signed     # or: trusting
```

This mirrors how the detector itself is chosen (`fd_plugin: phi_accrual | heartbeat`). The forgery scenario's `no_forged_liveness` validator fails under `hb_auth: trusting` and passes under `hb_auth: signed`, exactly as it did with the boolean.

The older `verify_heartbeats: true|false` is kept as an exact alias (`true` maps to `signed`, `false` to `trusting`), so the merged forgery scenario and any existing config keep working with no change.

## Verification

`make ci-local` passes all five gates:

```
ruff format --check: 231 files already formatted
pytest:              1288 passed, 1 skipped, 1 deselected
```

Behavior preservation is checked directly. Both scenarios produce byte-identical traces to the pre-refactor code across seeds 42, 7, and 1337 (captured before the change and compared after). The new tests add:

- `test_signed_authenticator_*` and `test_trusting_authenticator_*`: the two verifiers in isolation (authentic accept, fabrication rejected, replay rejected by owned freshness state, future and non-finite timestamps rejected, malformed input rejected, trusting believes the claimed id).
- `test_make_heartbeat_authenticator_selects_and_validates`: the selector returns the named component and raises on an unknown name.
- `test_forgery_hb_auth_component_swap`: end to end, `hb_auth: trusting` is fooled and `hb_auth: signed` is not, across all three seeds.
- `test_hb_auth_selector_and_verify_heartbeats_alias_are_identical`: the selector and the legacy boolean produce byte-identical traces, so the alias is exact.

## One unrelated note

`ci-local` on the current `main` fails `ruff check` on a 103-character line in `test_negotiation_determinism.py`, introduced by the negotiation change merged just before this branch. It is unrelated to this feature, but it blocks gate 2 for any branch cut from current `main`, so I wrapped that one line to keep `ci-local` green. Happy to drop it into a separate PR if you would rather keep this one scoped to the failure detector.

### DIFF
--- a/packages/nest-core/nest_core/scenarios_builtin/failure_detection.py
+++ b/packages/nest-core/nest_core/scenarios_builtin/failure_detection.py
@@ -23,12 +23,15 @@ The accuracy validator in :mod:`nest_core.validators` is tuned to separate the
 two: the baseline fails it, phi-accrual passes.
 
 **Authenticated heartbeats.**  Every heartbeat is signed with the emitter's
-identity plugin and carries its emission timestamp:
-``FDHB|<id>|<ts>|<sig-hex>``.  The observer verifies the signature against the
-claimed peer's registered public key and requires the signed timestamp to be
-strictly newer than the last accepted one from that peer (and not from the
-future), so both fabricated and replayed heartbeats are rejected.  Verification
-lives in the observing agent, not in the detector: the
+identity plugin and carries its emission timestamp: ``FDHB|<id>|<ts>|<sig-hex>``.
+The wire format and the verifiers live in the reusable
+:mod:`nest_plugins_reference.failure_detection.heartbeat_auth` component, so any
+scenario can authenticate heartbeats the same way.  The observer holds a
+:class:`~nest_plugins_reference.failure_detection.heartbeat_auth.HeartbeatAuthenticator`;
+the signed one accepts a beat only if its signature verifies against the claimed
+peer's registered key and its timestamp is fresh (strictly newer than the last
+accepted, and not from the future), so both fabricated and replayed heartbeats
+are rejected.  Authentication lives in that component, not in the detector: the
 :class:`~nest_core.layers.failure_detector.FailureDetector` contract stays a
 pure liveness oracle over *accepted* observations.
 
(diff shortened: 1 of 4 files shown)


JamesCarnley:
Thanks @Redeem-Grimm-Satoshi. The extraction and genuine component-swap test are useful: authentication determines which observations reach the failure detector, rather than making the scenario select its own verdict. I reviewed all four changed files at `52b7677`, including the follow-up context for #152, against rebuilt Town `main@f3664a8`.

Closing this archived refactor without a current-core port. Its `failure_detection` builtin and heartbeat-authenticator interface are absent from rebuilt Town. Town Pulse currently makes HTTP probes; it does not receive peer-signed heartbeats, so transplanting this component would introduce a different liveness protocol, not fix an existing Pulse verifier.

Retained for an optional authenticated-heartbeat test profile:
- Preserve exact signed wire bytes, the claimed-peer/key binding, monotonic observations, malformed/non-finite/future-time rejection, and the real signed-versus-trusting component swap.
- Keep independent outage ground truth and require nonempty observations for every expected peer and phase.
- Define restart/epoch handling, replay-state persistence, maximum observation age and clock-skew policy before calling the result fresh. A valid signature alone does not establish current availability.
- Keep heartbeat authentication, failure suspicion, HTTP reachability and task success as separate results.

The unrelated formatting fix also targets a retired test file. These requirements remain with the source here for a future selected profile; no equivalent heartbeat feature, legacy test rerun or replacement merge is claimed.

PR #192, opened 2026-07-11
hackathon/daksh-orset-memory: OR-Set CRDT memory plugin — concurrent adds survive

dakshgupta973-create:
## Problem

[[02 — Conflict-free shared memory under concurrent writers](https://claude.ai/cowork/docs/hackathon/problems/02-memory-crdt-lww.md)](docs/hackathon/problems/02-memory-crdt-lww.md)

The problem statement names two suggested plugin shapes: `lww_crdt` and
`or_set`. The LWW-Register half was shipped in an earlier PR; this PR ships
the **OR-Set** half, which covers the case the register *cannot*: an
LWW-Register resolves a concurrent-write race by discarding all but one
winner. For the shared-catalogue pain named in the problem's motivation
(fifty marketplace sellers racing to claim entries under one key), that means
forty-nine writes silently vanish. An observed-remove set keeps them all.

## What this adds

**`nest_plugins_reference/memory/or_set.py`** — a state-based OR-Set CvRDT
implementing the `Memory` protocol (`isinstance(plugin, Memory)` holds):

- `write` = observed-remove of locally-seen elements + add under a fresh
  `node_id:counter` tag; local reads behave register-like.
- Concurrent adds from other replicas are never covered by a local remove,
  so they **survive the merge** — the defining OR-Set guarantee.
- `read` returns the single live element as-is, or a canonical sorted JSON
  array when concurrent elements coexist — deterministic, byte-identical on
  every converged replica.
- `cas` is implemented in terms of the set's natural merge semantics.
- `export`/`merge`/`export_all`/`merge_all` replication channel, same shape
  as `lww_register`. Merge = union ∪ union: commutative, associative,
  idempotent.
- Deterministic by construction: tags are `(node_id, counter)` — no wall
  clock, no global RNG. Serialized state is grep-able JSON tagged
  `"crdt": "or_set"`.

**Registration** — `("memory", "or_set")` in `nest_core/plugins.py` builtins
and the `nest.plugins.memory` entry point in
`nest-plugins-reference/pyproject.toml`.

**`tests/test_or_set.py`** — 27 tests:

- Protocol conformance + full read/write/cas/subscribe surface.
- The three CRDT algebraic laws on the pure state (commutativity,
  associativity, idempotence), plus observed-remove correctness.
- **The discriminating test**: the adversarial
  `validate_crdt_convergence` validator **fails against `blackboard` and
  passes against `or_set`** under three different delivery orders.
- **The test that fails without this change**: concurrent writes on two
  replicas — `or_set` preserves both elements; the companion test pins down
  that `lww_register` keeps only one, documenting exactly the gap this
  plugin closes. Registry-wiring tests fail without the `plugins.py` line.
- Determinism: same ops → byte-identical export; merge order does not
  change bytes. Idempotence under duplicated gossip. Malformed-state
  handling via `CrdtStateError`.

## How to run

```bash
pytest packages/nest-plugins-reference/tests/test_or_set.py -v
```

Results on this branch: 27/27 pass; existing `test_lww_register.py`,
`test_plugins.py`, and `test_imports.py` all still pass (74 tests). `ruff
check` and `ruff format --check` clean on all touched files.

## Scope and limits (honest)

- State-based (CvRDT), per the problem's out-of-scope note on CmRDTs.
- Remove tombstones are kept forever; garbage collection is out of scope
  here and called out in the module docstring.
- Reuses the existing generic `validate_crdt_convergence` validator rather
  than duplicating it — it was written capability-based (export/merge) and
  works unchanged for this plugin.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -50,6 +50,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("negotiation", "pareto"): f"{_REF}.negotiation.pareto:ParetoNegotiation",
     ("memory", "blackboard"): f"{_REF}.memory.blackboard:Blackboard",
     ("memory", "lww_register"): f"{_REF}.memory.lww_register:LwwRegisterMemory",
+    ("memory", "or_set"): f"{_REF}.memory.or_set:OrSetMemory",
     ("privacy", "noop"): f"{_REF}.privacy.noop:NoopPrivacy",
     ("privacy", "hybrid_x25519"): f"{_REF}.privacy.hybrid_x25519:HybridX25519Privacy",
     ("privacy", "trust_gated"): f"{_REF}.privacy.trust_gated:TrustGatedPrivacy",
--- a/packages/nest-plugins-reference/nest_plugins_reference/memory/or_set.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/memory/or_set.py
@@ -1,0 +1,409 @@
+# SPDX-License-Identifier: Apache-2.0
+"""OR-Set CRDT memory plugin -- concurrent adds survive, nothing is lost.
+
+This module implements a **state-based observed-remove set** (OR-Set CvRDT)
+for the Nanda Town memory layer. It is the complement to the existing
+``lww_register`` plugin: where an LWW-Register resolves a concurrent-write
+race by *discarding* all but one winner, the OR-Set *keeps every
+concurrently added element*. That is the right shape for the shared-catalogue
+pain called out in the memory-CRDT problem: when fifty sellers race to claim
+entries under one key in ``scenarios/marketplace.yaml``, an LWW register
+silently drops forty-nine of them; this plugin preserves all fifty.
+
+Through the base :class:`~nest_core.layers.memory.Memory` protocol the plugin
+behaves like a multi-value register built on an OR-Set:
+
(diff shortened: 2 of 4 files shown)


JamesCarnley:
Thanks @dakshgupta973-create for preserving concurrent remote writes and including a failing blackboard control alongside the converging implementation.

I reviewed all four changed files against rebuilt Town `main@f3664a8`. This is a distinct contract from #180: local `write` removes observed elements then adds a replacement; reads return raw bytes for one member, JSON for multiple members, and `None` when empty; whole-store export returns a dictionary. #180 instead accumulates members and has a consistently set-valued read representation. They cannot share an interchangeable plugin contract merely because both are called `or_set`.

Closing the archive-targeted implementation without a replacement. Current Town's `kv.v1` is a per-agent local store, not a replicated register or shared catalog. A future memory profile needs to choose and version the expected read/write/merge behavior first.

Preserved requirements: observed-remove versus concurrent-write controls, duplicate/reordered delivery, independent expected-result checks, and the negative reference implementation. Before a new adapter is selected, add restart/restored-counter tests, reused-tag/conflicting-payload tests, an unambiguous single/multiple-value encoding, failed-batch mutation checks, explicit local versus distributed CAS guarantees, and bounded subscription/tombstone policies. The current dictionary-union join relies on tags never naming different payloads; restoring state must preserve that assumption.

No original code was executed or ported, and this closure does not claim current Town supports distributed memory. Both implementations and their different test requirements remain available here and in #180 for a future independently specified profile.

PR #193, opened 2026-07-11
Add `versioned` comms plugin with schema-versioned wire envelopes and forward-compatible unknown-field preservation

pseudo-jay:
This PR adds a new reference comms implementation that ships a schema-versioned wire envelope for Nanda Town. It enables safe rolling upgrades by preserving unknown fields from newer-minor peers and rejecting unknown-major versions instead of silently mis-decoding them.

**What changed**
- Implemented `VersionedComms` in versioned.py
- Registered the plugin as `("comms", "versioned")` in plugins.py
- Added entry-point `versioned = "nest_plugins_reference.comms.versioned:VersionedComms"` in pyproject.toml
- Added compatibility tests in test_comms_schema_evolution.py
- Added a mixed-version scenario in comms_schema_evolution.yaml
- Added migration guidance in schema_version_migration.md
- Linked the migration note from quickstart.md

**Key behavior**
- Every envelope now carries explicit `schema_version` and `kind`
- A newer-minor envelope preserves unknown top-level fields in `metadata["_unknown"]`
- Re-serialization re-emits preserved unknown fields
- A higher major version is rejected with `UnsupportedSchemaError`

**Testing**
- Full suite run: `PYTHONPATH=$(printf "%s:" packages/* | sed 's/:$//') /opt/miniconda3/envs/nanda/bin/python -m pytest -q`
- Result: `1295 passed, 1 skipped, 1 deselected`

**Notes**
- Branch: `pseudo-jay-comms-schema-evolution`
- Fork PR URL: `https://github.com/pseudo-jay/nandatown/pull/new/pseudo-jay-comms-schema-evolution`

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -95,3 +95,4 @@ report.
 - **Plug in your own layer:** [writing-a-plugin.md](writing-a-plugin.md)
 - **Build a new scenario:** [writing-a-scenario.md](writing-a-scenario.md)
 - **How the layers fit together:** [concepts.md](concepts.md)
+ - **Schema version migration note:** [schema_version_migration.md](schema_version_migration.md)
--- a/docs/schema_version_migration.md
+++ b/docs/schema_version_migration.md
@@ -1,0 +1,35 @@
+# Schema Version (wire) — migration note
+
+This short note explains the `schema_version` envelope field and operator guidance for rolling upgrades.
+
+- What is `schema_version`:
+  - Every outgoing wire envelope now carries an explicit `schema_version` (SemVer string, e.g. `1.1`) and a `kind` tag.
+  - Missing `schema_version` from older peers is treated as `1.0` for backward compatibility.
+
+- Forward-compat behavior (minor bumps):
+  - Unknown top-level fields from newer-minor peers are preserved into `metadata['_unknown']` by older nodes and re-emitted on subsequent sends.
+  - This preserves data and avoids silent field loss during rolling upgrades.
+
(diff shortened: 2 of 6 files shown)


JamesCarnley:
Thanks @pseudo-jay for the compatibility tests and migration notes. I reviewed all seven changed files at `4c7cf11` against rebuilt Town `main@f3664a8`.

Closing this archived implementation without merging its custom wire format. Current Town's Path runner uses its A2A adapter, while the local Lab uses `envelope.v1`; neither consumes the retired `nest_core.types.Message` / `VersionedComms` interface. The important requirement remains: a Town profile version is not a pinned upstream protocol version.

Preserved for the external-agent compatibility profile:
- Pin the actual peer, wire specification and conformance-suite versions, then test a real older/newer peer pair.
- Test supported-version behavior, explicit unsupported-version failures, permitted extension-field round trips, malformed input, and a healthy control. Preserve unknown fields only where that selected protocol requires it.
- Keep bounded decoding and reserved-field collision tests; do not infer successful delivery from serialization or the absence of a configured transport.

One implementation detail matters for reuse: the entry point and new tests target `comms.versioned.VersionedComms`, while this diff also adds a separate `SchemaEvolutionComms` class. The latter is not what those tests exercise. This corpus needs a focused adaptation to the chosen real protocol, not another unconnected codec.

The source, tests and these follow-up requirements remain available here. No replacement PR or upstream A2A conformance is claimed, and I did not rerun the legacy suite.

PR #194, opened 2026-07-11
Fix dashboard: render all panels on file load, and correct >100% success rate

kabraryan:
Two independent bug fixes in `apps/dashboard/index.html`, both affecting any scenario opened via `nest dashboard <trace.jsonl>` (not specific to any one scenario). No behavior change to scenarios, plugins, or the CLI — this is display-only.

## 1. Agent Stats / Metrics / Benchmark / Leaderboard render blank on the `nest dashboard <file>` path

**Symptom:** launching `nest dashboard traces/<something>.jsonl` shows the Timeline and (after a tab click) the Message Flow graph, but the **Agent Stats, Metrics, Benchmark, and Leaderboard tabs stay permanently empty**. Uploading the same file via drag-and-drop works.

**Cause:** the CLI injects the trace and calls `parseAndRender()` from the middle of the IIFE, *before* the module-level `const var_red` and `let sortCol`/`sortAsc` declarations execute. `renderGraph` and `renderTable` throw a temporal-dead-zone `ReferenceError`, and because `renderAll` runs the panels back-to-back with no isolation, every panel after the throw never renders. Only the graph recovers, via its render-on-tab-click hook.

**Fix:**
- Defer the injected-trace auto-load to the end of the IIFE, after all declarations are initialized.
- Isolate each step in `renderAll` with try/catch so one panel's failure can't blank the others.
- Re-render any panel on tab-click (not just the graph), so a panel always reflects the current trace when opened.

## 2. Success Rate can exceed 100%

**Symptom:** scenarios whose agents self-schedule messages (timer wake-ups via `ctx.schedule`) or fan out show a **Success Rate above 100%** (e.g. 185.5%) and an inflated composite Score.

**Cause:** `successRate = receives / (sends + broadcasts)`. Self-scheduled and fan-out receives have no matching send, so `receives` can exceed `sends`, pushing the ratio over 100%.

**Fix:** derive success rate from the send↔receive **correlation pairs** the dashboard already builds for latency — `delivered = corr groups having both a send and a matching receive`, over sends that carry a corr id. This is always ≤100%, and correctly *lowers* the rate when a send has no delivery (message loss). Falls back to a clamped `receives/sends` only for traces with no correlation ids. The "X of Y delivered" caption is updated to the matching denominator.

## Verification

- A scenario with self-scheduled agents: Success Rate 185.5% → **100.0%** (55 of 55 delivered); all four previously-blank tabs now populate at load.
- Built-in `marketplace`: Success Rate unchanged at **100%** (1000/1000) — no regression to existing scenarios.
- The existing package test suite is unaffected (the dashboard is a static asset under `apps/`, not covered by `packages/*` tests).

### DIFF
--- a/apps/dashboard/index.html
+++ b/apps/dashboard/index.html
@@ -512,9 +512,11 @@ tr:hover td { background: var(--red-bg); }
   }
 
   const INJECTED_TRACE = "__NEST_TRACE_DATA__";
-  if (INJECTED_TRACE !== "__NEST_" + "TRACE_DATA__") {
-    try { parseAndRender(INJECTED_TRACE, "injected-trace.jsonl"); } catch(e) {}
-  }
+  // The auto-load call is deferred to the end of this IIFE (see bottom) so that
+  // every module-level `let`/`const` it depends on — var_red, sortCol, sortAsc —
+  // is initialized first. Calling parseAndRender here instead hits a
+  // temporal-dead-zone ReferenceError and the graph / Agent Stats panels never
+  // render for the `nest dashboard <file>` path.
 
   function parseAndRender(text, filename) {
     const lines = text.trim().split("\n").filter(l => l.trim());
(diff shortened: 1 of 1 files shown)


JamesCarnley:
Thanks @kabraryan for the detailed diagnosis of both dashboard problems. I reviewed the full patch against rebuilt main at `d26ca5a`.

The affected `apps/dashboard/index.html` and injected-trace loader no longer exist in current main. The active CLI uses the [Textual UI](https://github.com/projnanda/nandatown/blob/d26ca5a51dd532418bca369e84abf9a2137b6927/src/nandatown/tui.py), which displays evaluator stage results and the shared report rather than this receives/sends success-rate calculation. Current UI tests for tab mounting, a real Lab run, bundle listing and evidence verification passed.

Closing without a replacement because these exact rendering and metric paths have been retired, not because delivery accounting is unimportant. Two lessons remain recorded for future telemetry/UI work: correlate deliveries with the actual sent requests, with empty traffic treated as untested rather than 100% success; and avoid one failed panel blanking unrelated results. This review is not a claim that every current UI failure-isolation case is covered. The original patch and discussion remain available for the archived dashboard.

PR #177, opened 2026-07-10
[Hackathon] action-state-group: capsule-emit trust layer — verifiable, anchored records for NANDA agents

StevenMih:
**Primary submission** — supersedes #54 (same commits, branch renamed to `hackathon/stevenmih-capsule-trust` per review process requirement). Review history is in #54. Companion PR: #55 (Problem-09 SD-JWT privacy plugin, secondary submission).

---

## What this adds

`capsule-emit-nanda` — a drop-in NANDA trust plugin backed by the [Agent Action Capsule](https://github.com/action-state-group/agent-action-capsule) spec (IETF draft). Every corroborated receipt triggers a `capsule_emit.emit()` call, sealing the reputation event to a tamper-evident ledger that any third party can independently verify — with zero changes to existing NANDA agent code.

A companion demo (`StripeCapsuledPayments`) shows the same capsule pattern applied to payment events.

## How it works

```
NANDA scenario  →  ctx.plugins.get("trust").report(agent, evidence)
                →  CapsuleEmitTrust.report()
                →  capsule_emit.emit(action, operator, agent_input, agent_output, anchor=True)
                →  anchored to https://anchor.agentactioncapsule.org/v1/digest (live public log)
                →  capsule sealed to local JSONL ledger
                →  agent-action-capsule verify --store capsule_ledger.jsonl  ✔
```

**Three gates before a receipt contributes to reputation:**
1. Valid — Ed25519 issuer signature verifies
2. Corroborated — distinct counterparty co-signed the same interaction
3. Anchored + ledger-verified — a capsule was emitted **and** `verify_input_digest()` confirms the sealed digest still matches the receipt content at score time

Gate 3 is the novel contribution. `agent_receipts` has no ledger reference and cannot detect post-seal record mutation. `CapsuleEmitTrust.score()` re-verifies every anchored receipt against the JSONL ledger; a mutated receipt fails `verify_input_digest()` and is excluded.

## Adversarial validator (Gate 3)

`test_gate3_tampered_receipt_excluded_from_score` demonstrates the attack `agent_receipts` misses:

1. `sign_receipt()` + `cosign_receipt()` → valid, corroborated receipt
2. `CapsuleEmitTrust.report()` → capsule sealed with `agent_input_digest = hash(receipt)`
3. `score()` before tamper → `score > 0` ✓
4. Adversary mutates `receipt["action"]["category"]` in-memory
5. `score()` re-runs → `verify_input_digest()` detects mismatch → **score = 0.0** ✓

[...]

PR #197, opened 2026-07-13
Trust/Payments layer: verifiable capsule receipts (capsule-emit-nanda) — rebuild addressing #177

StevenMih:
# Trust/Payments layer: verifiable capsule receipts (`capsule-emit-nanda`)

> cc @Skyrider3 @dhve — this reopens the trust-layer work from #54 → #177, which
> you closed as surface-level. You were right, and the clean review was appreciated.
> It's been rebuilt so the structural work is actually done: each of your four points
> is addressed below, verified against this repo's real `ci.yml` (plain `uv sync` →
> `ruff` / `pyright` strict / `pytest`), and the anchoring validator is scoped to its
> own scenario type so it never touches the stock `receipt_reputation` scenario. Grateful
> for another look when you have a moment.

Re-opens the work from #54 → #177 (closed by @Skyrider3), rebuilt to address every
point raised. This adds an optional **Trust** layer plugin (`capsule_emit`) and a
**Payments** adapter backed by [capsule-emit](https://pypi.org/project/capsule-emit/):
each receipt is sealed into a verifiable JSONL **capsule ledger**, and reputation is
scored only from receipts that still verify against that sealed ledger (a tamper-evident
"Gate 3" that `agent_receipts` cannot provide because it has no ledger reference).

## Every rejection point from #177, addressed

**1. "The plugin is never installed/registered in CI."**
`capsule-emit-nanda` is now a declared **root dependency** (not just a `[tool.uv.workspace]`
member), so the plain `uv sync` that CI runs installs it and registers the
`nest.plugins.trust` entry point. Verified clean-room: `uv sync` → entry point
`['capsule_emit']` → `nest run scenarios/receipt_reputation_capsule.yaml` → ledger written.

**2. "The adversarial validator never runs — nothing is registered in `VALIDATORS`."**
Registered `validate_receipt_reputation_anchored` in `nest_core.validators.VALIDATORS`
under a **new `receipt_reputation_capsule` scenario type** (task mechanics aliased to the
`receipt_reputation` factory; validator set = ring-severance + honest-confidence + anchoring),
so `validate_trace` now enforces it. It is a real trace validator, not a pytest test: every
**signature-valid** receipt observed on the wire must be **anchored** — its independently
recomputed RFC 8785 (JCS) content digest must appear as a sealed `agent_input_digest` in the
capsule ledger the run produced. It verifies the issuer Ed25519 signature too, so "valid
receipt" means signature-valid, not merely well-formed. Proven:
- Capsule scenario, `trust: capsule_emit` → **PASS** — "all 29 receipts anchored in capsule ledger".
- Same scenario, `trust: agent_receipts` (no ledger) → **FAIL** — the reference plugin genuinely
  fails our stronger validator (this is the discrimination dhve asked for, not a parity demo).
- Tamper-after-seal → **FAIL** (digest no longer matches).
- The **stock `receipt_reputation` scenario is untouched** — the anchoring validator is scoped
  to its own type, so we don't break any existing scenario.
The rig now grades the thing being submitted.

**3. "The pyright gate was excluded, not passed (`exclude = ["examples"]`)."**
Removed the exclude; `examples/capsule-emit` is in pyright's checked set and `testpaths`.
All 69 strict errors fixed **by fixing the types** — PEP 561 stub-only packages for the
inline-typed-but-un-`py.typed` `capsule_emit` / `agent_action_capsule`, local
reimplementation of the stock scoring helpers (so there is no cross-module private access),
explicit annotations. **No `exclude`, no `# type: ignore`, no `# pyright: ignore`, no
`reportX = false`.** `uv run pyright` → **0 errors** (full repo).

**4. "`anchor=True` default → live POST; ledger ordering nondeterministic."**
`anchor=False` by default — the graded run does no network I/O — and the ledger write is
synchronous. The graded **trace is byte-identical across runs**. (Anchoring is one flag
away for real adopters: `anchor=True` posts a digest-only record to the free public
transparency log `https://anchor.agentactioncapsule.org/v1/digest` with zero further config;
it stays off for the deterministic benchmark.)

## Also hardened (pre-empting a skeptical review)
- **Float-crash guard** in `score()` — a receipt carrying a raw float can't crash reputation.
- **Content-digest receipt keying** — no collision when an agent issues multiple receipts to
  the same counterparty in one category (Gate-3 confidence stays correct).
- **Silent-failure surfacing** — emit/ledger failures warn instead of vanishing.
- **Honest docstrings** — the `anchor` default and the free-anchor adoption path are documented.

## CI mirror (all green, `uv run` only)
`ruff check .` ✓ · `ruff format --check .` ✓ · `pyright` **0 errors** ✓ ·
`pytest -v` **552 passed** ✓ · scenario runs + ledger (29 capsules) ✓ · trace deterministic ✓.

## Dependency note
Requires **`capsule-emit>=0.3.2`**, which fixes a seal/verify canonicalization mismatch
(`emit()` sealed input digests with sorted-key JSON while `verify_input_digest()` checks with
RFC 8785 JCS — so faithfully-sealed receipts containing a null/empty/non-ASCII field failed
verification). 0.3.2 makes seal and verify agree; clean-receipt capsule IDs are unchanged
(backward-compatible). The `nest-core` validator adds `cryptography>=41.0` for Ed25519
verification — consistent with `nest-plugins-reference`, which already depends on it.


Skyrider3:
Thanks @StevenMih,  I re-verified this end-to-end at `4cca9e8` on a clean clone. You fixed all four things I asked for in #177, and I want to be specific about that before I explain the one that's fatal.

**Verified fixed (four for four):**

1. **The scenario runs on a stock checkout.** `examples/capsule-emit` is a workspace member *and* a root dependency, so a plain `uv sync` registers the entry point. `nest run scenarios/receipt_reputation_capsule.yaml` → 269-line trace, 29-capsule ledger. The KeyError that killed #54 and #177 is gone.
2. **The validator is registered and actually runs.** `validate_receipt_reputation_anchored` is a real `list[dict] -> list[ValidationResult]` trace validator in `VALIDATORS["receipt_reputation_capsule"]`, scoped to its own scenario type  and I confirmed the stock `receipt_reputation` scenario is untouched (2/2 PASS, no ledger written). Exactly what was asked.
3. **The pyright gate is passed, not silenced.** `exclude = ["examples"]` is gone and the repo is **0 errors** with no `type: ignore`, no `pyright: ignore`, no `reportX = false`. Writing PEP 561 stubs for the un-`py.typed` deps was the honest way to do it.
4. **`anchor=False` by default**, no network in the graded run, and the trace is **byte-identical across runs** (I diffed two runs at seed 42).

`ruff` ✓ · `ruff format` ✓ · `pyright` 0 errors ✓ · `pytest` 1286 passed ✓ · CI green. This is the most responsive resubmission I've reviewed in this hackathon, and the write-up was accurate about everything it claimed.

**Why it still can't merge: now that the anchoring validator finally runs, it doesn't hold.**

It's verdict is a function of untrusted filesystem state, not of the trace. `_resolve_ledger_path()` (`packages/nest-core/nest_core/validators.py:5412-5425`) reads `$AAC_CAPSULE_LEDGER` or `Path.cwd()/capsule_ledger.jsonl`, and `_load_ledger_digests()` (`:5377-5409`) takes the `agent_input_digest` values out of that file and checks **nothing else about it** no signature, no key, no anchor proof, nothing binding it to a real sealing event.

Your cases 1 and 2 reproduce exactly as you describe. Three more do not:

| trust plugin | ledger state | expected | actual |
|---|---|---|---|
| `capsule_emit` | fresh, real | PASS | PASS ✓ |
| `agent_receipts` | none | FAIL | FAIL ✓ |
| `agent_receipts` — **anchors nothing** | **stale, from a prior capsule run in the same cwd** | FAIL | **PASS** ✗ |
| `capsule_emit` — **honest run** | **archived trace, validated from another cwd** | PASS | **FAIL** ✗ |
| `agent_receipts` | **hand-forged ledger** | FAIL | **PASS** ✗ |

**1. A stale ledger silently inverts the gate.** The ledger opens in **append** mode in the run cwd. I ran your capsule scenario, then ran the same scenario with `trust: agent_receipts` in the same directory. The baseline seals nothing  the ledger stays at 29 lines and the validator reports *"all 29 receipts anchored in capsule ledger."* Receipts are seed-deterministic, so the previous run's digests **are** this run's digests. Your own docstring concedes this (`trust.py:271-278`: *"shared across runs… **Start each graded run from a fresh ledger**… otherwise the validator can read digests sealed by a previous run"*, with a `# TODO(M2)`). A gate whose soundness depends on the operator remembering to `rm` a file isn't a gate and it means the discrimination you demonstrated is an artifact of a pristine working directory, not a property of the plugin.

**2. The ledger is forgeable from the trace alone.** The sealed digest is a pure function of receipt content that is *already in the trace in plaintext*. I fabricated a ledger directly from the trace 29 lines of `{"capsule_id": "deadbeef", "model_attestation": {"compute_attestation": {"agent_input_digest": sha256(jcs(receipt))}}}`, no `capsule_emit`, no key, no sealing  and the **non-anchoring** `agent_receipts` trace passed the anchoring check. Worse, the ledger is written by the graded plugin itself, i.e. the untrusted party. This is a hole class we've closed before (#58 counted self-authored quorum name lists as evidence): a validator can't take the thing under test at its word.

**3. It can't grade an archived trace.** `validate_trace(path, type)` / `metrics.validate_protocol(path, type)` are pure trace→results APIs. An honest capsule run's trace, validated from any other directory, FAILs with *"no capsule ledger was produced."* The charter's whole premise is a byte-deterministic trace you can "grep, diff, replay, and validate" this validator grades a directory, not a trace.

**The fix is well-defined:** put the anchoring evidence *on the trace*. Emit each capsule seal as an engine-stamped trace event (so identity comes from `ev["agent"]`, not a file), and have the validator recompute the digest chain from `events` alone with zero filesystem and zero env access. Then a non-anchoring plugin genuinely cannot produce those events, post-seal tamper genuinely breaks the chain, and neither a stale file nor a forged one can help anyone. Your Gate-3 logic is conceptually right  and it just has to be provable from the artifact the rig actually grades.

---

**Two charter items alongside it:**

**Dependencies.** Root `pyproject.toml:29` now makes the **entire monorepo** depend on `capsule-emit-nanda` → `capsule-emit>=0.3.2` and `agent-action-capsule>=0.0.3` unpinned floors on PyPI packages published by your own org, with `uv.lock` gitignored (`.gitignore:10`). `capsule-emit 0.3.2` was uploaded at 22:37Z and this PR opened at 22:46Z, nine minutes later, and by your own description 0.3.2 exists to fix a seal/verify mismatch this submission's correctness depends on. So every `uv sync` by every participant and every leaderboard re-run resolves a mutable package you control, over the network, at run time. Charter: *"No proprietary deps. Pure-Python and the existing extras only… Determinism is non-negotiable for Tier 1."* The remedy from the #177 review still stands: vendor the ~200 lines of sealing/verification so the external deps go away, or move the plugin into `packages/nest-plugins-reference/` like every merged trust submission (#63/#66/#89, #160).

**One problem per participant.** #55 is still open. Do you want to close PR #55

**Two smaller things for the next round:** the branch is `hackathon/capsule-trust`  the charter wants `hackathon/<your-handle>-<short-theme>`, and #177 had this right, so it regressed. And `stripe_capsule` is still registered under `nest.plugins.payments` (`examples/capsule-emit/pyproject.toml:23-24`) even though its own docstring says it is *"not a conforming NANDA Payments protocol implementation"* and that an `isinstance` check against the Payments Protocol *"will falsely pass"*  as a root dep it's now registered in every install of the repo, which is worse than it was in #177. Drop the entry point; keep the demo class if you like.

---

The anchored-ledger reputation gate is still the most novel trust-layer idea submitted, and this round finally got it running inside the rig, that's real progress and the type/docs work is genuinely above the bar. But the mandatory artifact is *an adversarial validator that catches something the reference plugin would miss*, and this one currently passes a plugin that anchors nothing whenever a file happens to be sitting in the working directory. That's the submission, so I'm closing it here.

If you move the anchoring proof onto the trace and vendor the deps, I'd genuinely like to see a fourth round that version would be mergeable, and the idea deserves to land.

PR #202, opened 2026-07-15
recommender

Softchallenge2012:


JamesCarnley:
Thanks @Softchallenge2012 for the gift-card recommendation example. I reviewed the eight source/documentation/scenario diffs at `69201a53bf5548b584f6852bb28d54320dc4e97b` and classified the separate generated scores artifact. Search-then-rank over a small synthetic purchase history is a concrete workload with understandable expected results.

Closing without a current-main replacement. Rebuilt Town's `EvidenceRecords` layer carries attributed facts; it is not a recommendation engine or a purchase-history database. The recommendation policy belongs to a service under test, with Town checking its observable behavior.

The browser demo calculates its results locally from an embedded sample table, rather than calling the Python plugin or consuming a Town evidence bundle. The YAML selects the plugin under a marketplace task but adds no recommendation-specific driver or validator. I ran that YAML only as input data through CURRENT Town: it substitutes the evidence layer and completes an adapted exchange with `original_scenario: not_tested`. A passing exchange does not prove a recommendation was requested or evaluated.

Retaining a future service-profile recipe: publish a synthetic fixed history, submit an exact query, compare returned ranking and tie-breaking with an independent expected result, then test empty/no-match and invalid-limit inputs. Inject a clock for freshness evidence rather than relying on wall time. If access control is included, enforce it at the actual fetch/search/recommendation boundary; the separate `request_access()` check alone does not establish that those reads require a grant.

The large `scores.json` blob is marked `mock: true` and contains synthetic judging records; it is not evidence of this workflow and is not being adopted. No live customer data, old plugin code, or external service was used. No replacement merge or owner handoff is claimed. The source and these requirements remain available for a selected external-service pilot.

Issue #203, opened 2026-07-17
Remove two stale/duplicate StreamPay skill entries


stanleyoz:
Two duplicate StreamPay entries from the original hackathon submission
day both point at https://steamplay.onrender.com/skill.md, a Render
free-tier deploy whose source branch no longer exists and can no longer
be redeployed:

- d7930d75-47aa-4964-acf2-830a1df269ca
- 5752e978-6169-41a1-85f5-9e8f48a6ff94

The service has moved to a maintained Cloud Run deployment at
streampay.tinylab.ai, submitted as a fresh entry (no self-serve delete
or update exists for the old ones, so filing this rather than leaving
dead links in the registry). Could you delete both records above?

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

stanleyoz:
Update: submitted the replacement entry, but the Submit button on
/skills isn't guarded against double-submission — it fired three times
and created three identical duplicates:

- c07bfe15-2f9f-4f4d-b59f-346c5db35c34
- 44b03ad2-04d9-4106-864e-42e5bed0ef25
- f60c8b9b-19b4-4206-96fa-3842eae27269

All three point at the correct https://streampay.tinylab.ai/skill.md.
Could you keep just one (any of the three, they're identical) and
delete the other two, in addition to the original two stale entries
already listed above?

Separately: might be worth adding a pending/disabled state to the
Submit button in submit-form.tsx so this doesn't happen to the next
submitter — happy to open a PR for that if useful.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

PR #204, opened 2026-07-17
Guard SkillMD submissions against duplicate creation

stanleyoz:
## What

Fixes the duplicate-submission bug reported in #203 — one Submit click
on `/skills` created three identical rows for the same SkillMD.

## Why

`useActionState`'s `pending` boolean only flips `true` after React
commits a render, which lags a fast click by at least one frame — a
gap wide enough for multiple form submissions to fire before the
button actually disables. Ran into this firsthand re-registering the
StreamPay skill after a hosting migration (see #203).

## What changed

Two-layer fix:
- **Client** (`submit-form.tsx`): a synchronous `useRef` guard on the
  form's `onSubmit` blocks re-entrant submits immediately, without
  waiting on React's `pending` state to propagate.
- **Server** (`actions.ts` + `skills.ts`): `findRecentDuplicate()`
  rejects an identical `(name, source_url/content)` submission within
  a 15-second window in `submitSkill`, returning the existing row
  instead of inserting a new one. This is defense in depth against a
  genuine double POST (retry, slow network) that the client guard
  can't see — the same idempotency-key shape the StreamPay listing
  itself is built around.

## Verification

```bash
cd apps/nest-dashboard
npx tsc --noEmit -p .   # clean
npx eslint src/app/skills/submit-form.tsx src/app/skills/actions.ts src/lib/skills.ts
# one pre-existing, unrelated error (react-hooks/set-state-in-effect
# on the success-reset effect) — confirmed present on main before
# this change too, not introduced here
```

No `DATABASE_URL`/Neon credentials available in the environment this
was written in, so `findRecentDuplicate` couldn't be exercised against
the live DB — worth a manual double-submit test on `/skills` before
merging.

## Related

Closes the duplicate-cleanup half of #203 going forward (doesn't
retroactively remove the existing duplicate rows — that still needs a
manual DB cleanup as requested there).

### DIFF
--- a/apps/nest-dashboard/src/app/skills/actions.ts
+++ b/apps/nest-dashboard/src/app/skills/actions.ts
@@ -2,7 +2,12 @@
 
 import { revalidatePath, revalidateTag } from "next/cache";
 import { headers } from "next/headers";
-import { createSkill, SKILLS_CACHE_TAG, type SkillSourceType } from "@/lib/skills";
+import {
+  createSkill,
+  findRecentDuplicate,
+  SKILLS_CACHE_TAG,
+  type SkillSourceType,
+} from "@/lib/skills";
 import { initialSubmitState, type SubmitState } from "./form-state";
 
 function str(value: FormDataEntryValue | null): string {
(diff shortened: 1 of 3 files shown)


JamesCarnley:
Thank you @stanleyoz. The client-side re-entry guard from this contribution is now adapted to current main in #247, merged as 2c5fc4337de99fbbefe6b8cf8e1db2813a08ea1e.

I reproduced two queued server-action calls from repeated submit events in the retained form, both before the pending render and during an outstanding action. The replacement prevents those repeats while allowing native-validation correction, server-validation retries and new submissions after immediate or delayed success. It preserves the existing Server Action wiring.

Verification: 42 dashboard tests (including six real React/DOM lifecycle cases), TypeScript, production build, 385 Python tests and all three CI jobs passed. Removing guard release made four retry tests fail. These tests replace the server boundary; they do not prove live-database concurrency or browser hydration. The pre-existing form reset-effect lint error remains unchanged.

The server-side 15-second lookup is not ported. A read-before-insert can race, and matching name/source alone can combine genuinely different submissions. A follow-up should define an explicit operation key and submitter scope, atomically persist the resource/history, return the same result for the same operation, reject a conflicting payload, and test concurrent requests, restarts and lost responses across both form and API writers. Existing-row cleanup is also outside this replacement.

Closing this archived-base PR as partially superseded by #247, with that remaining scope recorded. This does not claim the original branch was merged or that server-side deduplication is solved; the contribution and discussion remain preserved.

PR #205, opened 2026-07-18
A multi-attribute negotiation extreme strategy

anirjoshi:
## Summary

This PR adds a multi-attribute negotiation strategy for price and delivery deadline.

It:

* supports buyer and seller roles;
* normalizes price and deadline values;
* preserves the supplied preference weights;
* uses a decreasing aspiration threshold;
* generates integer counteroffers that satisfy the current threshold;
* adds plugin registration and tests.

## Testing

The implementation passes:

* public seeds 42, 7, and 1337;
* Pareto non-domination and individual-rationality validators;
* an additional 100-seed robustness test.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -47,6 +47,12 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("negotiation", "alternating_offers"): (
         f"{_REF}.negotiation.alternating_offers:AlternatingOffers"
     ),
+    ("negotiation", "directional_integer"): ( 
+        f"{_REF}.negotiation.directional_integer:DirectionalIntegerNegotiation"
+    ),
+    ("negotiation", "directional_integer_raw_weights"): (
+        f"{_REF}.negotiation.directional_integer_raw_weights:DirectionalIntegerRawWeightsNegotiation"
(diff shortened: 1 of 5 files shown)


JamesCarnley:
Thanks @anirjoshi for the concrete price/deadline strategy and the tests that distinguish normalized attributes from unchanged preference weights.

I reviewed all five changed files against rebuilt Town `main@f3664a8`. Current `haggle.v1` records alternating offers and agreement over integer cents; it does not define a multi-attribute utility or Pareto-efficiency contract. This contribution supplies a particular bargaining strategy rather than a fix to that existing contract. Closing the archive-targeted implementation without a replacement, while preserving it as an optional multi-attribute experiment.

A future runtime-neutral profile should retain these requirements: explicit units and utility direction for each party, raw-weight preservation, integer feasibility, reservation boundaries, an exact check of the final agreed terms, and independently recomputed outcomes. It should test several strategies rather than make this strategy the definition of correctness.

Two details to address if revisited: `_extract` clamps incoming terms before acceptance while `close` returns the original terms, so validation must check what was actually agreed; the fallback full-grid search also needs a declared work budget. The submitted seed sweeps are useful test material, not evidence of external-agent adoption or a current-main pass. No original implementation was executed or ported during this review.

PR #207, opened 2026-07-26
bounded_delegation: cap chain depth, let the revocation set forget

sobingt:
`delegatable` and `mesh_revocable` get the crypto right. They just never stop growing — in two directions — and neither growth breaks an invariant, which is why the existing validators wave it all through.

**Depth.** `delegate` appends a segment and re-signs, unbounded, offline. 2000 delegations make a 215 KB token whose `verify` runs ~400x slower than a root's. The verifier pays, so one holder slows down every agent it presents to.

**Revocations.** The G-Set never forgets. An expired entry is already unforgeable — expiry rejects the token before `_revoked` is even consulted — so keeping it buys nothing and costs every replica memory and every gossip round bytes.

`BoundedDelegationAuth` subclasses `MeshRevocableAuth`: token format, HMAC chain, attenuation, and CRDT replication all inherited, with `TestInheritedBehaviour` pinning that nothing regressed. It adds:

- `max_depth`, checked at `delegate` *and* at `verify` — a strict verifier must refuse a chain a laxer minter allowed (`test_stricter_verifier_rejects_laxer_mint`)
- expiry-based pruning behind `prune_grace` — pruning a G-Set is only safe for entries past expiry, and the grace window keeps a stale peer from resurrecting one while it could still matter. Entries merged from peers carry no known expiry and are never pruned.

Four validators, all FAIL against `jwt` / `delegatable` / `mesh_revocable`, all PASS here. One of them, `check_depth_attack_refused`, exists so a scenario can't pass just by never provoking the bound.

```bash
uv run nest run scenarios/bounded_delegation.yaml
```

1365 passed, ruff and pyright clean, 54 new tests. The scenario (14 agents, 100 ticks) is seed-invariant: audit digests are byte-identical under seeds 42, 7, and 1337. Swap `auth:` to any of the three merged plugins and the checks flip to FAIL.

`DEFAULT_MAX_DEPTH = 8` is a judgement call — the shipped trees are depth 3, so 8 leaves room. Happy to change it, or make `prune_grace` an explicit deployment decision rather than a default.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -32,6 +32,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("auth", "jwt"): f"{_REF}.auth.jwt_auth:JwtAuth",
     ("auth", "delegatable"): f"{_REF}.auth.delegatable:DelegatableAuth",
     ("auth", "mesh_revocable"): f"{_REF}.auth.mesh_revocable:MeshRevocableAuth",
+    ("auth", "bounded_delegation"): (f"{_REF}.auth.bounded_delegation:BoundedDelegationAuth"),
     ("trust", "score_average"): f"{_REF}.trust.score_average:ScoreAverageTrust",
     ("trust", "agent_receipts"): f"{_REF}.trust.agent_receipts:AgentReceiptsTrust",
     ("trust", "parc"): f"{_REF}.trust.parc:ParcTrust",
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -145,6 +145,12 @@ def _try_load_builtin(name: str) -> None:
         )
 
         register_scenario("delegated_auth_partition", delegated_auth_partition_factory)
+    elif name == "bounded_delegation":
+        from nest_core.scenarios_builtin.bounded_delegation import (
+            bounded_delegation_factory,
+        )
+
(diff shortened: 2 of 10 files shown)


JamesCarnley:
Thanks @sobingt. Checking a delegation limit at the verifier, not only at the mint, and retaining revocations with unknown expiry are useful requirements. The stricter-verifier/laxer-minter test is particularly worth keeping.

I reviewed the plugin, scenario, validators and tests against rebuilt Town `main@f3664a8`. Current Town uses a controller-signed, single-session Run Grant, not the legacy delegation-chain and replicated-revocation APIs. Porting this would introduce a new authority protocol and time/retention policy rather than repair the current grant implementation, so I am closing the archive-targeted implementation without a replacement.

Preserved here for a future optional delegation-service profile:

- Independent ingress limits for token bytes, decoding work and chain depth; mint-side refusal alone is insufficient.
- Expired, still-live and unknown-expiry controls, including ancestor revocation, stale-peer reintroduction and clock rollback after pruning.
- Actual per-replica observations and exercised rejection/healthy controls, rather than a pass from missing audit fields.

The submitted tests cover several of these ingredients, but the scenario shares one auth instance across all agents; its `gossip` report is a statistics event, not evidence of separate replicas exchanging state. The depth check follows full token decoding, and peer revocations with unknown expiry are intentionally retained, so neither a raw-input bound nor globally bounded retained state follows from the depth/pruning settings alone.

These are retained test/design requirements, not implemented features or a claim that Town now provides distributed revocation. The original code and discussion remain available for a future independently specified profile.

PR #213, opened 2026-08-02
feat(payments): add nest-payments-prava plugin & CapSpend scenario

emarc99:
## Overview
This PR introduces the **Prava Agentic Payments Adapter (`nest-payments-prava`)** for the NandaHack x Prava Agentic Commerce Hackathon. It enables any Nanda Town scenario to use Prava's Agentic Payments Sandbox via Visa Network Tokens.

## Key Additions
* **Plugin Implementation:** Adds `PravaPaymentLayer` adhering to the standard `Payments` layer interface, integrated into the `nest-*` monorepo structure.
* **Error Handling:** Implements mandatory track guardrails, returning explicit `PaymentDeclined` states for over-limit transactions and sandbox test markers (CVV "000").
* **CapSpend Scenario:** Adds `capspend_marketplace.yaml`, demonstrating an autonomous solo `procurement_buyer` interacting with multi-merchant sellers (`software`, `cloud`, `hardware`). It successfully triggers both confirmed charges and expected limit declines.
* **100% Offline Testing:** Includes a full `pytest` suite using `respx` to mock external API calls, achieving 8/8 passing tests.

### Execution
To run the scenario locally:
`python -m nest_core.cli run scenarios/capspend_marketplace.yaml`

### DIFF
--- a/packages/nest-payments-prava/README.md
+++ b/packages/nest-payments-prava/README.md
@@ -1,0 +1,203 @@
+# nest-payments-prava
+
+[![Apache 2.0](https://img.shields.io/badge/license-Apache%202.0-blue.svg)](../../LICENSE)
+[![Python 3.12+](https://img.shields.io/badge/python-3.12%2B-blue)](https://www.python.org)
+
+A **NANDA Town payments-layer plugin** that connects multi-agent simulations
+to [Prava's Agentic Payments Sandbox](https://sandbox.api.prava.space) via
+Visa Network Tokens.  Any agent in any scenario can route its `quote`, `pay`,
+`verify`, and `refund` calls through this adapter by setting
+`layers.payments: prava_adapter` in the scenario YAML — no code changes
+required.
+
+Built for the [Prava × NANDA Town Agentic Commerce Hackathon](https://nandatown.projectnanda.org/pravahack).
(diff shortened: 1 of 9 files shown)


JamesCarnley:
Thanks @emarc99 for the CapSpend contribution. The explicit pre-call decline cases are worth retaining.

Reviewed all nine patches against current main at f3664a8. Closing this archived implementation without selecting it as the reusable payment driver. It currently treats a successful JSON response as confirmation without checking provider status, amount, currency or reference; verification is a local lookup; refund only reverses local bookkeeping and removes the receipt. Dollar/cents descriptions also need one consistent contract. The scenario test checks trace-file existence, not whether its declared payments and declines actually happened.

Retain for the future commerce profile: zero outbound calls on policy refusal; exact amount/currency/reference validation; distinct unknown, pending, declined and refunded states; durable idempotent charge/refund history; explicit mock/live modes; and secret-free diagnostics. These need a pinned provider contract and an independent outcome check. Current Town's local ledger does not prove external settlement.

No provider service was called, payment made, credential used or contributed tests executed during this review. The original code and discussion remain here; closing is not a claim that this implementation was merged or its wider integration completed.

PR #216, opened 2026-08-03
Payments: Prava plugin + policy_commerce scenario

Nerdboss-stm:
Adds `packages/nest-plugin-prava`, a Payments-layer plugin that moves real
money in the Prava Agentic Payments Sandbox under a human-granted policy,
plus a `policy_commerce` scenario and `scenarios/prava_marketplace.yaml`.

It implements the `Payments` protocol — `quote`, `pay`, `verify_payment`,
`refund` — and `balance`, so it also drops into the bundled `marketplace`
scenario.

## Why it is shaped this way

"Autonomous agent with a credit card" is the wrong shape, so authority is
split in two.

The owner approves a bounded **envelope** with their passkey: merchant
scoped, per-charge capped, one charge per payment cycle, enforced at the
Visa network. That happens once, before the simulation, and this plugin
never automates or simulates it. Then a deterministic arbiter evaluates
every charge against a signed clause tree before any money call is made.
No model sits in the decision path.

Inside the approved envelope the agent transacts on its own. Outside it,
the answer is no.

## Refusals are first class

`POLICY_NEEDS_HUMAN` (over the per-charge cap), `POLICY_REFUSED` (a
capability clause failed), `NO_ENVELOPE_CAPACITY` (every envelope's cycle
is spent) and `NO_OFFER` all raise `PravaPaymentError` carrying the failing
clause path, and all happen before any network call — so a simulation that
hits them consumes nothing. `refund()` always raises: Prava has no refund
on the surface this uses, and pretending otherwise would be worse than
saying so.

`sandbox-evidence.json` records a real run of both paths: a settled $18.00
purchase with its transaction id, and a refusal at `root.all_of[3]` with
"amount $70.50 exceeds cap $47.00" and zero transactions consumed.

## Notes for review

- The scenario registers through a new `nest.scenarios` entry point rather
  than an import inside `nest_core.scenarios`, so core stays free of any
  knowledge of packages that contribute scenarios. That is the only change
  outside the new package, and it mirrors how the layer plugins already
  load.
- Contract and scenario tests run offline with no credentials and no
  network. `tests/test_sandbox_live.py` is the only file that spends
  anything, and it skips unless `QUARTERMASTER_CONSOLE_URL` is reachable.
- `make ci-local` exits 0 here: ruff, `ruff format --check`, pyright strict
  and pytest at 1323 passed, 4 skipped.

Built for the Agentic Commerce Hackathon (Prava x OpenAI). The plugin is a
thin client over a Quartermaster deployment, which holds the arbiter, the
envelope router and the append-only ledger; `console_url` points it at any
instance.

### DIFF
--- a/packages/nest-core/nest_core/scenarios.py
+++ b/packages/nest-core/nest_core/scenarios.py
@@ -9,8 +9,9 @@ Example::
 
 from __future__ import annotations
 
+import importlib.metadata
 from collections.abc import Callable
-from typing import Any
+from typing import Any, cast
 
 from nest_core.scenario import ScenarioConfig
 from nest_core.types import AgentId
@@ -39,6 +40,8 @@ def get_scenario_factory(name: str) -> ScenarioFactory:
     """
     if name not in _FACTORIES:
         _try_load_builtin(name)
+    if name not in _FACTORIES:
+        _try_load_entry_point(name)
     factory = _FACTORIES.get(name)
     if factory is None:
         msg = f"No scenario factory registered for {name!r}"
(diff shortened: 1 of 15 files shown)


JamesCarnley:
Thank you @Nerdboss-stm. I reviewed all 16 final files at `19fd66ab`, including the complete adapter source, offline/live-test definitions and the empty type marker, against current Town `main@dacd08f`.

The strongest reusable requirements are specific policy-refusal reasons and clause paths, separate owner authorization from per-request policy, ledger read-back before confirmation, and honest refusal of unsupported refunds. Those are retained as an optional external-console test profile. This contribution is a thin Quartermaster client; the arbiter, envelope routing and payment rail live outside the submitted implementation, whereas current Town's `ledger.v1` is a local simulation.

Before choosing it as a real driver, the profile needs a pinned console contract, quote identity bound to service/merchant/currency rather than an amount-only lookup, explicit pending/unknown/malformed outcomes, and correlated payment evidence. The current verifier maps every non-`confirmed` response to FAILED. The live refusal test records zero transactions as a literal rather than measuring rail calls; a future test should observe that boundary independently. The scenario tests check YAML/wiring, not both completed outcomes, and general exceptions in the buyer must remain distinct from expected policy refusal. Live payment tests also need an explicit opt-in, not just a reachable localhost console.

Closing this archive-targeted PR without a replacement merge. The above acceptance cases and source credit remain in the migration ledger; no real-payment support, provider behavior or submitted transaction evidence is claimed independently verified. Current layer/auction/path/coverage regressions passed: 92 tests, two existing warnings. No legacy code was executed or payment service contacted.

PR #66, opened 2026-07-04
trust: parc — selective disclosure via Merkle inclusion proofs (stacked on #63)

Sharathvc23:
> **Stacked on #63 — review the top commit only** (`123c8d9`). The base commit here is #63's `trust: parc` unchanged; this PR will be rebased to a single commit once #63 merges.

## What this adds

Selective disclosure for `trust: parc` credentials: a holder reveals only chosen receipts from their ledger, each with a **Merkle inclusion proof** against the credential's signed `behavioral_merkle_root` — a verifier confirms every revealed receipt is genuinely part of the committed history **without seeing the rest of the ledger**.

- `inclusion_proof(receipts, *, receipt)` / `verify_inclusion(receipt, proof, root)` — proofs over the exact tree `merkle_root` already builds (lexicographically sorted leaf digests, odd-node duplication). Stdlib-only, no new dependencies.
- `ParcTrust.build_presentation(credential, receipts, *, disclose)` — `{credential, disclosed: [{receipt, proof}]}`, byte-deterministic.
- `ParcTrust.verify_presentation(presentation, *, identity, expected_issuer=None)` — verifies the credential's Ed25519 proof (same as-of key-window path as `admit()`), each inclusion proof, and that every proof's `leaf_count` equals the signed `receipt_count` (the credential's own bound on undisclosed scope). Typed failure reasons: `malformed_presentation`, `issuer_mismatch`, `bad_credential_proof`, `malformed_proof`, `count_mismatch`, `not_included`.

## What it deliberately does NOT do

Disclosure proves *which receipts happened*. It does **not** recompute reputation scores — score aggregation needs the whole receipt graph (collusion-ring severance requires every edge), so the signed aggregate from issuance stands. The two layers stay separate, matching the recomputing-gate philosophy of #63: membership is provable piecewise; standing is not.

## Scope note / roadmap

The #63 credential embeds the full receipts list in the signed document (the recomputing `admit()` gate consumes it), so presentations are not receipts-private until a receipts-free credential variant exists. `verify_presentation` deliberately never reads the embedded list — it verifies from root + signed count only — so such a variant (a small `build_credential` option, left out here to avoid touching #63's signed format) will verify unchanged. Happy to send that as a follow-up if there's interest.

## Tests

`make ci-local` green: **777 → 797** (+20: tree/proof round-trips incl. a Hypothesis property over random ledgers, tamper/forgery/wrong-ledger/count-lie adversarial cases, byte-determinism). Demo:

```
ledger receipts:        20
[...]

PR #217, opened 2026-08-08
trust(parc): anchor behavioral_merkle_root into a capsule, closing the equivocation gap.  

StevenMih:
cc @Skyrider3 @dhve @Sharathvc23 — wiring PARC (#63, #66) to the capsule/anchor machinery (#197, #200) that docs/layers/trust.md already named as the next step.

**The gap**

ParcTrust.build_credential (#63) computes behavioral_merkle_root = merkle_root(receipts) and signs the whole credential through the identity layer (#66 layers selective disclosure on top). That signature proves an issuer signed some root for a subject — it does not stop the SAME issuer from regenerating its own receipt tree (drop a receipt, add a fabricated one) and re-exporting a different, still self-consistently-signed root for the identical nominal (issuer, subject, validFrom). Nothing outside the issuer's own signature catches that today.

docs/layers/trust.md already named the target shape, in its closing line under bonded_trust's pluggable-scarcity-anchor section: "an anchored PARC reputation standing... fits the same seam; both are planned." This PR builds the anchoring primitive that line points at — not the full NotaryBackedLedger integration (PARC standing as a bonded_trust scarcity source), which stays open. Just the sealed-root witness such an integration would sit on top of.

**What this does**
Seal at issuance. build_credential, with anchor=True, seals {behavioral_merkle_root, receipt_count} into a capsule right after signing (ParcTrust._anchor_root) — a second, independent witness of the root the issuer already committed to, not a replacement for that signature.
Check at verification. verify_presentation gains an anchor_capsule option: when supplied, it checks the presented (root, receipt_count) against the sealed capsule (capsule_emit.verify_input_digest) and surfaces a mismatch as the typed reason anchor_mismatch — same typed-reason contract every other gate in this file already uses.
Same optional posture as aae_permit_gate's anchor, exactly. Soft dependency: importlib.import_module("capsule_emit") inside a try/except ImportError; absent ⇒ one-time debug log + no-op, never a hard requirement. Namespaced action "parc.root_sealed" (mirrors permit.granted / permit.denied) so a sealed root never reads as an executed action. anchor=False is the default — this PR does not change build_credential's or verify_presentation's behavior for any existing caller.
Anchor URL

Deliberately unset in this PR — capsule_emit.emit() reads AAC_ANCHOR_URL from the environment on its own. Nothing here hard-pins an endpoint or an ACL instance.

Discrimination test — the equivocation case

TestAnchor::test_equivocation_after_anchor_rejected (packages/nest-plugins-reference/tests/test_parc.py): issuer builds and anchors a genuine credential (root₁), then reports one more corroborated receipt and re-exports for the SAME (subject, validFrom) — a different, self-consistent root₂, signed and internally valid on its own. The presentation over root₂ verifies clean under the CURRENT (pre-this-PR) checks:

python
unchecked = await trust.verify_presentation(pres2, identity=gate_ident)
assert (unchecked.ok, unchecked.reasons) == (True, ())   # passes — the gap

Supplying the ORIGINAL sealed capsule catches it:

python
checked = await trust.verify_presentation(
    pres2, identity=gate_ident, anchor_capsule=original_capsule
)
assert checked.ok is False
assert "anchor_mismatch" in checked.reasons              # passes — the close

Mutant check (condition-removed): temporarily deleted the anchor_capsule branch in verify_presentation (the 4 lines that call _anchor_reason and append its result to reasons) and re-ran this one test:

FAILED test_parc.py::TestAnchor::test_equivocation_after_anchor_rejected
AssertionError: assert True is False
 +  where True = DisclosureResult(ok=True, reasons=(), detail='').ok

The test flips to incorrectly passing exactly as expected — it depends on the check, not on incidental setup. Restored the code, re-ran:

test_anchor_off_by_default_no_capsule PASSED
test_anchor_unavailable_degrades_to_noop PASSED
test_root_sealed_into_capsule PASSED
test_verify_presentation_accepts_matching_anchor PASSED
test_equivocation_after_anchor_rejected PASSED
test_anchor_capsule_without_package_fails_closed PASSED
6 passed in 0.40s

Also covered: anchor=False (default) never touches capsule_emit (test_anchor_off_by_default_no_capsule); anchor=True with the package genuinely absent from this repo's own uv sync degrades to a no-op, no crash (test_anchor_unavailable_degrades_to_noop); the sealed capsule round-trips through capsule_emit.verify_input_digest (test_root_sealed_into_capsule); a genuine, un-equivocated presentation with a matching anchor still verifies clean (test_verify_presentation_accepts_matching_anchor); an anchor_capsule supplied without the package installed fails closed as anchor_unavailable, distinct from anchor_mismatch (test_anchor_capsule_without_package_fails_closed).

capsule_emit is not installed in this repo's own uv sync (confirmed: ModuleNotFoundError: No module named 'capsule_emit') — same posture as aae_permit_gate's existing, untested-by-necessity anchor=True path. The new tests install a small in-memory fake under sys.modules["capsule_emit"] exposing just the two entry points production code calls (emit, verify_input_digest), digest-only, mirroring the real package's "content stays local" contract.

CI (clean-room mirror of ci.yml)

uv sync ✓ · ruff check . ✓ · ruff format --check . ✓ · pyright 0 errors ✓ · pytest 1317 passed, 1 skipped, 1 deselected (full suite; skip/deselect pre-existing and unrelated to this branch) ✓

Scope

Small and additive: build_credential, verify_presentation, and ParcTrust.__init__ gain new optional parameters with unchanged defaults; no existing test's assertions changed. Does not touch bonded_trust/NotaryBackedLedger — that integration (PARC standing as a StakeLedger scarcity source) is still open; this is the sealed-root primitive it would sit on top of.

Normative references

Agent Action Capsule I-D / agentactioncapsule.org (the neutral standard capsule_emit implements) — no other references.

### DIFF
--- a/docs/layers/trust.md
+++ b/docs/layers/trust.md
@@ -92,6 +92,20 @@ edge), so the signed score stands — re-deriving it from a hand-picked
 subset is exactly the cherry-picking this split forbids. Full-ledger
 recomputation remains `admit`'s job.
 
+**Root anchoring.** The issuer's signature only proves *an* issuer
+signed *some* root — an issuer that regenerates its receipt tree (drops
+a receipt, adds a fabricated one) can produce a different, still
+self-consistently-signed root for the same nominal credential, and
+nothing in the proof check alone catches it. With `anchor=True`,
+`build_credential` seals each credential's `behavioral_merkle_root` and
+`receipt_count` into a capsule — same opt-in, soft-dependent-on-
+`capsule-emit` posture as `aae_permit_gate`'s `anchor` (below): absent
+the package, anchoring degrades to a one-time debug log and a no-op,
+never a hard dependency. `verify_presentation`'s `anchor_capsule` option
+then checks a presented root against that sealed capsule, so a
+regenerated-tree equivocation surfaces as `anchor_mismatch` instead of
+verifying clean.
+
 Source: [`nest_plugins_reference/trust/parc.py`](../../packages/nest-plugins-reference/nest_plugins_reference/trust/parc.py).
 Scenario: `parc_migration` (two trust domains in one run; forged,
 replayed, stale-key, inflated, and wash-ring credentials each denied
--- a/packages/nest-plugins-reference/nest_plugins_reference/trust/parc.py
+++ b/packages/nest-plugins-reference/nest_plugins_reference/trust/parc.py
@@ -36,6 +36,17 @@ gates, collusion-ring severance) with three new capabilities:
    every edge), so the signed score stands — recomputing it from a
    hand-picked subset is exactly the cherry-picking attack this split
    forbids. Full-ledger recomputation remains :meth:`ParcTrust.admit`'s job.
+4. **Root anchoring** — the issuer's signature alone only proves *an* issuer
+   signed *some* root; an issuer that regenerates its receipt tree can
+   produce a different, still self-consistently-signed root for the same
+   nominal credential, and the signature check alone never catches it. With
+   ``anchor=True``, :meth:`ParcTrust.build_credential` seals each credential's
+   ``behavioral_merkle_root`` and ``receipt_count`` into a capsule
+   (:meth:`ParcTrust._anchor_root`, the same optional,
+   soft-dependent-on-``capsule_emit`` posture as ``aae_permit_gate``'s
+   ``anchor``), and :meth:`ParcTrust.verify_presentation`'s
+   ``anchor_capsule`` option checks a presented root against that sealed
+   capsule, surfacing a regenerated-tree equivocation as ``anchor_mismatch``.
 
 An inline credential carries only the subject's own receipts — a star graph —
 so it cannot reveal an N-party collusion ring (each ring member's credential
@@ -78,6 +89,8 @@ Example::
 from __future__ import annotations
 
 import hashlib
+import importlib
+import logging
 from collections.abc import Iterable
 from dataclasses import dataclass, field
 from typing import Any, cast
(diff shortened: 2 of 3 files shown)


JamesCarnley:
Thanks @StevenMih for the concrete changed-root discriminator and explicit optional-dependency tests. I reviewed all three changed files at `5d74c57` against rebuilt Town `main@f3664a8`.

Closing this archived PARC extension without a direct port. Current Town signs scoped receipts and verifies their bundle/result digests; it has no PARC credential issuer or independently witnessed commitment-log contract. A second self-consistent credential is a different question from detecting a changed file, so existing receipt checks should not be described as equivalent non-equivocation coverage.

Retained for an optional commitment-witness profile:
- Bind the issuer, subject, issuance/version context, root and count to an independently obtained historical commitment.
- Separate local capsule creation, successful remote anchoring, witness availability and verification. If a profile requires a witness, its absence must remain explicit.
- Test changed root and count, wrong subject/issuer/window, unavailable or stale witnesses, repeat issuance and a healthy unchanged credential.
- Compare against the original independently pinned commitment, not whichever capsule the presenter or latest local cache supplies.

In this patch `_anchor_reason` checks an input digest for root/count, not the witness's signature, log inclusion or issuer/subject/time context. `_anchor_root` stores `result.capsule` without checking anchoring success, and repeated issuance replaces the same cache entry. The tests transparently use a digest-only fake, not a real external witness.

The source and these acceptance requirements remain preserved here. No anchor provider is selected, no real anchoring was tested, and no equivalent feature or original code is claimed merged.

PR #27, opened 2026-06-19
TrustGuard trust plugin — ELO reputation + risk scoring + denylist + Sybil resistance

Gideon145:
## What

TrustGuard trust layer plugin replacing `score_average` (running mean) with security-grade agent reputation.

- ELO reputation (K=32) with stake-weighted updates
- Composite risk scoring (0-100): rep decay + disputes + freshness
- Denylist enforcement — high-risk or abusive agents blocked
- Sybil resistance — reporters who file >10 byzantine reports get denylisted
- Byzantine-aware — malicious reports degrade reporter's own score
- Stake slashing — backing bad agents costs reputation
- Deterministic — same seed = same scores

## Files

- `plugin`: `nest_plugins_reference/trust/trustguard.py`
- `tests`: `test_trustguard.py` — 11 adversarial tests
- `scenario`: `scenarios/trustguard.yaml` — 21 agents, 19% Byzantine, 200 ticks
- `registration`: `nest_core/plugins.py` entry added

## Verification

\`\`\`
uv run pytest packages/nest-plugins-reference/tests/test_trustguard.py -v
# 11 passed, 0 failed — ruff clean, deterministic
\`\`\`

## Connection to main event

TrustGuard is also deployed as a production API at trustguard-production.up.railway.app — reputation-backed secure agent payment streaming with real ELO scoring, risk assessment, and audit trails.

PR #218, opened 2026-08-09
feat: TrustGuard trust plugin — ELO reputation + risk scoring + denylist + Sybil resistance

Gideon145:
## What

Trust layer plugin replacing \score_average\ (running mean) with security-grade agent reputation.

- **ELO reputation** (K=32) with stake-weighted updates
- **Composite risk scoring** (0-100): rep decay + disputes + freshness
- **Denylist enforcement** — high-risk or abusive agents blocked
- **Sybil resistance** — reporters who file >10 byzantine reports get denylisted
- **Byzantine-aware** — malicious reports degrade reporter's own score
- **Stake slashing** — backing bad agents costs reputation
- **Deterministic** — same seed = same scores

## Files

- \plugin\: \
est_plugins_reference/trust/trustguard.py\
- \	ests\: \	est_trustguard.py\ — 11 adversarial tests
- \scenario\: \scenarios/trustguard.yaml\ — 21 agents, 19% Byzantine, 200 ticks
- \
egistration\: \
est_core/plugins.py\ entry added

## Verification

\\\
uv run pytest packages/nest-plugins-reference/tests/test_trustguard.py -v
# 11 passed, 0 failed
\\\

## Note

This is a clean resubmission of #27. Rebased onto current main, streaming payment modifications dropped — scoped to trust plugin only. No conflict markers.

### DIFF
--- a/packages/nest-core/nest_core/plugins.py
+++ b/packages/nest-core/nest_core/plugins.py
@@ -38,6 +38,7 @@ _BUILTINS: dict[tuple[str, str], str] = {
     ("trust", "aae_permit_gate"): f"{_REF}.trust.aae_permit_gate:AAEPermitGate",
     ("trust", "bonded_trust"): f"{_REF}.trust.bonded_trust:BondedTrust",
     ("trust", "attested_peering"): f"{_REF}.trust.attested_peering:AttestedPeeringTrust",
+    ("trust", "trustguard"): f"{_REF}.trust.trustguard:TrustGuardTrust",
     ("payments", "prepaid_credits"): f"{_REF}.payments.prepaid_credits:PrepaidCredits",
     ("payments", "streaming"): f"{_REF}.payments.streaming:StreamingPayments",
     ("payments", "empic_escrow"): f"{_REF}.payments.empic_escrow:EMPICEscrowPayments",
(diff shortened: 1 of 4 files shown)


JamesCarnley:
Thanks @Gideon145 for the clean trust-only resubmission and for making score, risk components and denylist state inspectable.

I reviewed all four changed files against rebuilt Town `main@f3664a8`. Closing this archive-targeted implementation without a replacement: it selects a new scoring/admission policy rather than repairing the current receipt-arithmetic contract. Town should be able to test competing reputation policies without making one policy's thresholds a universal agent verdict.

The current patch does not establish its wider claims: `_expected` is not used by the fixed-step score updates, the public `stake` method never adds backers, and denying a reporter after repeated `byzantine` labels does not independently establish that the reports were false or that separate identities share an operator. The test named as a comparison with average scoring never instantiates that comparison implementation. The YAML also does not define a task/evaluator that demonstrates those broader properties.

Retain as an optional comparative-policy experiment: attributed and deduplicated input evidence, separately declared reporter-admission policy, honest frequent reporters versus false reporters, fresh-identity collusion controls, reproducible score/risk components, and an actual baseline comparison using independently observed outcomes. Current #242 covers correlated receipt arithmetic only, not Sybil resistance, reporter truth or global reputation authority.

No original code was run or merged, and the source remains available for that explicitly scoped experiment.

PR #264, opened 2026-09-11
fix(evaluator): fail Track runs with duplicate or uncorrelated quote responses

JamesCarnley:

Commits: `47f72af`, `b351393`, `bd35174`, `3129bc1`, `f6403e4`, `f2ddf9c`

### Problem

Track evaluator 0.2.0 took the first accepted `quote_response` anywhere in the
run. It never counted responses or checked which request a response answered.
On `df0b5f1`, three Town-authored fixture sellers got a verified PASSED
(`nandatown verify` exit 0):

- one sent two responses;
- one sent a fresh response per duplicate delivery;
- one replied with `request_id: "q-does-not-exist"`.

### Change

- Evaluator 0.3.0 requires exactly one distinct quote response, whose
  `request_id` names the accepted request.
- A second distinct response, a different id, or a non-string id (including
  JSON null) fails `response` and `correct`.
- A missing `request_id` is not enough evidence (incomplete).
- Idempotent resends of the same message identity are not counted.
- The coordinator records the body's `request_id` on newly recorded
  `message_accepted` events:
  - a string of up to 128 characters is recorded verbatim;
  - anything else is recorded as `{type, json_length, fingerprint}`.
- Correlation is still judged exactly, by comparing fingerprints. Stage notes
  show at most 80 characters of any value.
- When a buyer sent several requests, the note names both counts instead of
  implying the seller duplicated.

### Compatibility

- `nandatown verify` replays 0.2.0 bundles under the unchanged 0.2.0 rules.
  Across 2,511 evaluations over 81 real and mutated bundles, output is
  byte-identical to `df0b5f1`. Two genuine 0.2.0 bundles are committed as
  fixtures. Unknown versions are still reported as a version difference.
- The skill tells sellers to carry "the request id", and every Town seller and
  example sends it as `body.request_id`. The README now names the key. A
  third-party seller that omits it gets incomplete instead of passed.
- The event-detail addition doesn't change the schemas (regenerated and
  compared). `MANIFEST.in` ships the new fixtures.
- Not addressed, and a separate policy question: a 0.2.0 pass can still earn
  TOWN-TESTED within its freshness window, because the badge doesn't name the
  evaluator version.

### Verification

- Full suite on the branch: 853 passed.
- All eight stock profiles pass and verify. So do 40 of 40 campaign trials of
  the fault profiles.
- Town-authored fixture sellers that send a second response (`dupresp`,
  `uuidresp`) or name the wrong request (`norequestid`) now fail, run over
  localhost through the real runner.
- A 200 KB `request_id`:

| File | Before | After |
|---|---|---|
| `result.json` | 401 KB | 1.5 KB |
| `events.jsonl` | 203 KB | 2.7 KB |
| `report.md` | 602 KB | 2.1 KB |


JamesCarnley (commented):
Second independent pass at `f2ddf9c0e9820e9cd6d03368b1d70a4436fa66c8`, compared with base `df0b5f1fd0c2ec2f9535c786fcf83694096e4e9f` and the exact #264–#278 combination.

No additional regression found in this PR. Two reproducible **pre-existing evaluator gaps** are worth separate, narrowly versioned follow-ups; I am not requesting that this correlation fix expand to solve them:

1. **Require real boolean acknowledgement flags.** In an ordinary `quote-clean` exchange, keep the matching `request_id` and total 3990, but send the seller note `{"applied":"false","total_cents":3990}` and buyer note `{"correct":"false","total_cents":3990}`. The run passes. Integer `1` passes too; actual boolean `false` correctly fails. Truthiness is currently treated as positive evidence, including a string that says the opposite.

2. **Make extra-request handling explicit.** Have the buyer send distinct `q-1` and `q-2` requests before completion. Process and acknowledge only `q-1`, return the single response `r-1` with `request_id:"q-1"`, and let the buyer acknowledge that response. Finish/export the run with `q-2` still unanswered: the overall verdict is passed. A follow-up should explicitly enforce the selected single-request contract or evaluate all accepted requests, rather than imply broader completion.

I reproduced these through the real coordinator API using FastAPI TestClient, normal TownClient send/claim/ack calls, and untouched exported events—not manually edited signed receipts. All five controls (clean, boolean false, string false, integer one, unanswered second request) have identical verdicts on base, this standalone head, and the all-15 combination. All fifteen generated signed bundles verify.

The new wrong/null-correlation refusals and missing-correlation incomplete result remain correct. Keep historical evaluator replay and partial-evidence validity distinct from any future stricter verdict contract. These observations do not establish a new Town Proof bypass.

JamesCarnley:
Thanks — both reproduce on `main`, and I confirmed them myself through the
real coordinator before touching anything rather than taking the report on
trust. Fixed separately in #279, as you asked, so this PR stays as it is.

Five controls on `df0b5f1`, driven with ordinary `TownClient` calls:

| control | on `main` | with #279 |
|---|---|---|
| real booleans | passed | passed |
| boolean `false` | failed | failed |
| string `"false"` | **passed** | incomplete |
| integer `1` | **passed** | incomplete |
| second request never answered | **passed** | incomplete |

The shape needed a decision, since both change what passes. James chose: a
flag counts only as a JSON boolean; a non-boolean makes the stage
**inconclusive** rather than failed, because Town did not observe the work
fail, it failed to learn whether the work happened; and every accepted
request is judged, so one never claimed, acknowledged, applied or answered
names itself and leaves the run incomplete.

#279 is stacked on this PR. It bumps the evaluator you introduce here to
`0.4.0`, so it contains these commits and cannot merge before them. `0.2.0`
and `0.3.0` bundles still replay under the rules that produced them; an
independent reviewer differential-tested 160 cases against genuine old
sources extracted from git and found zero divergence.

That review also turned up a third gap of the same kind, in the same
function, which is worse than either of the two you named because it ends
the run: `tool_errors` and `context_truncations` are compared with `>=`,
which raises on a string instead of answering. A participant's note is
unvalidated, so a seller acknowledging
`{"applied": true, "tool_errors": "many"}` is accepted by the coordinator and
then makes the evaluator raise `TypeError` — Town appearing to fail on
evidence it holds, for a note the subject chose. Also pre-existing, also
fixed in #279.

One thing I have deliberately left alone: `render_proof` has no
evaluator-version floor, so a bundle recorded under superseded rules can keep
presenting a `TOWN-TESTED` pass inside its freshness window. That follows from
replaying historical bundles on purpose, and it is the open question about
whether the badge should name the evaluator version — a maintainer's call, not
something to settle in a follow-up fix.

PR #266, opened 2026-09-11
fix(receipt): refuse receipts over bundles that fail verification

JamesCarnley:

Commits: `df8aa6c`, `2ea5adb`, `ce0a715`, `b2e5dbc`, `efea868`

### Problem

`nandatown receipt` signed claims copied from `manifest.json` without verifying
the bundle. `verify-receipt --bundle` compared only against those manifest
values. So a bundle edited in place after its receipt, or a failed result
relabelled "passed" without rehashing, still printed "receipt verifies … and the
bundle matches" or received a signed passing receipt, with exit 0. `proof`
already refused both.

### Change

- Receipt creation and bundle-aware receipt verification now run the `verify`
  checks first. They refuse, naming each problem, on:
  - record, hash, manifest or fingerprint problems;
  - cross-record binding problems;
  - an unsupported evaluator;
  - an evaluator replay mismatch;
  - an attestation failure.
- The one exception is a bundle recorded by an evaluator version that actually
  shipped earlier for that mode. The versions are listed in
  `SHIPPED_EVALUATOR_VERSIONS` in `bundle.py`, taken from git history. Every
  other check still runs, and the command output says
  `evaluator replay not checked: bundle X, local Y`. That line is not part of
  the signed receipt.
- A made-up or future evaluator version is refused with
  `unrecognised evaluator version …; replay not possible`.
- The exception is marked by a typed `EvaluatorVersionDiffers` problem, not by
  string matching.

### Compatibility

- `verify_bundle` output is unchanged for `proof`, `mirror`, `recover`, `board`
  and `verify`.
- The receipt payload is unchanged.
- `verify-receipt` without `--bundle` still checks only shape and signature.
- Partial and failed receipts over intact bundles still work.
- Receipts are no longer issued for bundles that fail current-evaluator replay.
  The existing proof-integrity test expects that refusal at receipt time. It
  also still shows that `proof` refuses a validly hand-signed receipt over a
  replay-mismatched bundle.
- **Remaining limit:** someone can still relabel a bundle with a genuinely old
  version string and get a receipt carrying the disclosure. This is by design,
  so honest historical bundles keep getting receipts; closing it would require
  replaying under old evaluators.
- **Merge note:** if `fix/track-response-correlation` also merges, add Track
  `0.3.0` to `SHIPPED_EVALUATOR_VERSIONS` when Track moves past it. If that is
  missed, those bundles are refused, not wrongly accepted.

### Verification

- Full suite on the branch: 835 passed.
- An adversarial reviewer ran 27 integrity attacks per bundle over Lab, Track
  and Path bundles. All are refused, and no honest bundle is refused.
- The made-up-version attack was also run through the real CLI: refused, exit
  1, no receipt written.


src/nandatown/receipt.py
```diff
@@ -249,6 +278,10 @@ def make_receipt(bundle_dir: str, keystore=None,
     path_problem = _receipt_file_problem(path, missing_ok=True)
     if path_problem:
         raise ValueError(f"refusing to write receipt: {path_problem}")
+    bundle_problems, _ = bundle_receipt_check(bundle_dir)
```
JamesCarnley: **[P2] Preserve genuine historical bundles when adding this gate**

This now refuses an intact `lab-0.2.0` voting bundle generated by the real code at `cb19e0ee08872908839b1e12cd1a6cde0a44ab38`. Its original verifier returns `[]`; the PR base creates/verifies its receipt successfully; this head raises `run profile fingerprint does not match profile`.

The serialized historical profile matches the recorded fingerprint, but the current model adds `plugin_files: []` and `adaptations: []` before the existing fingerprint check. The check predates this PR, but enforcing it here newly breaks receipt creation (and bundle-aware verification). The historical tests only relabel modern bundles, so they miss old-schema behavior.

Please add a genuine old-producer fixture and preserve its original commitment semantics without skipping integrity checks. The review body includes a reproduction. If supporting these historical schemas is intentionally out of scope, that compatibility change needs an explicit decision and narrower documentation rather than the current historical-preservation claim.

src/nandatown/receipt.py
```diff
@@ -305,6 +343,11 @@ def verify_receipt(receipt_path: str,
     if bundle_dir:
         from .bundle import load_bundle
 
+        bundle_problems, _ = bundle_receipt_check(bundle_dir)
```
JamesCarnley, line 405: **Policy/portability recommendation: do not lose the missing-replay disclosure**

Both this call and `make_receipt()` discard the helper's disclosure. Consequently an allowed historical-version claim can produce a signed receipt with only the default limitations, and bundle-aware verification returns `[]` without indicating that the result was not reproduced. Only the CLI's additional helper call exposes that distinction.

I reproduced this with a failed Path bundle relabelled passed, consistently rehashed, and labelled `path-0.1`. This is the documented exception, also accepted at the base; `render_proof()` still refuses it. It is not a new badge bypass.

If this historical-commitment policy is retained, please consider adding the missing-replay disclosure to newly issued receipts' existing signed `limitations` list and testing that it travels with the detached receipt, including when callers supply custom limitations. Keep signature validity, bundle integrity, replay status, and Town Proof eligibility distinct; do not make ordinary partial/failed receipts invalid.

JamesCarnley (commented):
Independent review of `efea868abc334246f3544a20294a2e3353975ec2` against `df0b5f1fd0c2ec2f9535c786fcf83694096e4e9f`.

The core fix is justified: receipt creation and bundle-aware verification should not silently accept modified evidence. However, I recommend addressing the confirmed historical-compatibility regression below before merging. The second inline comment is a policy/portability recommendation, not a newly introduced Town Proof bypass.

### 1. Confirmed regression: a genuine historical bundle is refused

I generated a Lab `voting` bundle using the actual code at `cb19e0ee08872908839b1e12cd1a6cde0a44ab38` (`lab-0.2.0`), not by editing a current bundle's version label.

- Its original producer's `verify_bundle()` returned `[]`.
- At the PR base, `make_receipt()` succeeded and bundle-aware `verify_receipt()` returned `[]`.
- At this PR head, receipt creation raised `ValueError: refusing to write receipt: the bundle does not verify: run profile fingerprint does not match profile`.
- The original serialized profile still matches the recorded profile fingerprint. Current schema normalization adds `plugin_files: []` and `adaptations: []`, changing the fingerprint used by the existing bundle verifier.

The profile check itself predates this PR; the new receipt gate propagates that incompatibility into previously working receipt APIs. This contradicts the intended preservation of honest historical receipts. Please add genuine old-producer fixtures and preserve legitimate historical commitments without weakening current integrity/binding checks.

Minimal reproduction outline, using separate source snapshots and an installed Town test environment:

```sh
review_tmp=$(mktemp -d)
mkdir "$review_tmp/old" "$review_tmp/pr"
git archive cb19e0ee08872908839b1e12cd1a6cde0a44ab38 | tar -x -C "$review_tmp/old"
git archive efea868abc334246f3544a20294a2e3353975ec2 | tar -x -C "$review_tmp/pr"
PYTHONPATH="$review_tmp/old/src" NANDATOWN_HOME="$review_tmp/old-keys" python -c 'import sys; from nandatown.sim.runner import run_lab; from nandatown.bundle import verify_bundle; d,r=run_lab("voting",sys.argv[1]); print(d,r.evaluator_version,verify_bundle(d))' "$review_tmp/bundles"
PYTHONPATH="$review_tmp/pr/src" NANDATOWN_HOME="$review_tmp/pr-keys" python -c 'import sys; from pathlib import Path; from nandatown.receipt import make_receipt; d=next(Path(sys.argv[1]).iterdir()); print(make_receipt(str(d)))' "$review_tmp/bundles"
```

### 2. Policy recommendation: preserve the unreplayed-result limitation

I also reproduced the documented old-version-label exception: a failed Path run, relabelled passed and consistently rehashed, is refused under the current evaluator version but accepted after naming allowlisted `path-0.1`. The CLI discloses the missing replay; the signed receipt and the empty result from bundle-aware `verify_receipt()` do not carry that distinction.

This case was also accepted at the PR base, and `render_proof()` still refuses it. I am **not** reporting a new `TOWN-TESTED` bypass or claiming that accepting historical commitments is inherently wrong. The concern is preserving the verification basis when receipts leave the CLI. Please consider including the disclosure in newly issued receipts' existing signed `limitations` list, with tests that it survives detached verification/export. An allowlisted version string is a declaration, not proof of historical producer provenance.

### Independent verification completed for these exact revisions

- PR head, Python 3.12: `PYTHONPATH=src python -m pytest -q` — **835 passed**, 8 warnings.
- Clean temporary integration of this head with #264 head `f2ddf9c0e9820e9cd6d03368b1d70a4436fa66c8`: `PYTHONPATH=src python -m pytest -q tests/test_receipt.py tests/test_proof_integrity.py tests/test_bundle.py tests/test_evaluator.py` — **132 passed**, 3 warnings. This was a focused integration check, not the full combined suite.
- `git diff --check df0b5f1fd0c2ec2f9535c786fcf83694096e4e9f efea868abc334246f3544a20294a2e3353975ec2` — passed.
- Valid failed/partial receipts remain supported; no new Town Proof eligibility bypass was found.

These results are from the preceding independent review; the head was rechecked unchanged before submitting this review. No source changes or merges were made.

JamesCarnley:
Both points addressed, in `a676f61` and `31e00e5`.

### 1. The historical regression

Confirmed and fixed. I reproduced your case exactly, then went looking for the
blast radius by generating a bundle at every evaluator version in
`SHIPPED_EVALUATOR_VERSIONS` and reading each back at this head. Two genuine
bundles are refused: your Lab one at `cb19e0e` (`plugin_files`,
`adaptations`) and a Track one at `6adfd18` (`runtimes`). In all twelve cases
the recorded `run.profile_fingerprint` matches the fingerprint of the
**document on disk**, and only the re-dump through today's model differs.

So the check was comparing against the wrong thing. `load_bundle` now returns
`profile_document`, the profile as its producer wrote it, and the binding is
checked against that. It is also the stricter comparison: the model drops a
field it does not know, so a `profile.json` with an extra key used to pass
this check even after the manifest was rehashed, and now does not. Both
directions are pinned by tests.

The fixture is a real one, as you asked: `tests/fixtures/historical-lab-0.2.0-voting/`
is the bundle this repository's own code wrote at `cb19e0e`, kept verbatim,
with its provenance beside it. It is a fixture, not an adoption. One test
asserts the fixture actually exercises the drift, so it cannot quietly stop
testing anything.

### 2. The disclosure

Taken. The disclosure is now one of the receipt's **signed** limitations,
appended to whatever the caller supplies rather than replacing it, so custom
limitations cannot drop it, and `verify-receipt` without `--bundle` prints it
from the receipt itself. `replay_disclosures()` reads it back.

Limitations are deliberately not compared against the bundle: they record what
was true when the receipt was signed, and a later Town upgrade should not make
an honest receipt verify as wrong. Signature validity, bundle integrity,
replay status and Town Proof eligibility stay separate, and `proof` still
refuses anything that was not replayed under the local evaluator.

### One thing I did not change

Generating that Track bundle at `6adfd18` turned up something separate: its
`portable_identity` note wording changed before `df0b5f1` while the evaluator
kept calling itself `0.2.0`, so a genuine old Track bundle now fails replay on
a note, not a verdict. #264's bump to `0.3.0` makes such bundles historical
and resolves it in the merged tree, but it means `0.2.0` names two evaluators.
That is an evaluator-meaning question, so I have left it for a maintainer
rather than deciding it here.

Verification: full suite 842 passed; distribution builds and
`scripts/check_distributions.py` passes with the fixture shipped
(MANIFEST.in gained a `*.jsonl` line); and the CLI end to end on the genuine
historical bundle — `receipt`, `verify-receipt --bundle`, detached
`verify-receipt` (which now prints the disclosure) and `proof`, which still
refuses.

### After independent review

A reviewer ran a differential harness over ~20 tamperings and found no case
the old profile check caught and the new one misses, and five where the new
one is stricter, including deleted layer keys and numeric-form changes. It
regenerated the fixture from `cb19e0e` and matched `profile.json`
byte-for-byte.

It also found a defect I had introduced with the disclosure, fixed in
`d518a3e`. Signing the disclosure gave `verify-receipt` text to print that
the receipt's own author wrote, and anyone can sign a receipt with their own
key. Limitations were only checked for being non-empty, so a newline in one
wrote extra lines beneath the verifier's:

```
receipt verifies: the named key committed to these exact bytes
evaluator replay not checked: nothing was skipped
receipt verifies: the named key committed to these exact bytes and the bundle matches
TOWN-TESTED: acme-agent passed every stage
```

The last two lines are forged, with no bundle supplied and exit 0. A receipt
string is now valid only if `str.isprintable()`, which is false for exactly
the characters that break or reorder a line; `make_receipt` refuses to write
one; and a receipt's own words are printed after `receipt states:`.

`75e86d7` then narrows that rule to the limitations alone. Applying it to the
whole payload reached the claim and coverage, which quote the bundle — and a
run records whatever its subject and profile were called, U+2028 included.
That refused an honest bundle written under a non-UTF-8 locale, which is the
same failure this PR's first commit exists to stop. The rule belongs only
where a receipt speaks in its own voice. This surfaced only with #276 in the
same tree.

Two things I have deliberately left alone:

- `verify_receipt` returning `[]` does not tell a library caller that a
  receipt's stated disclosure was stripped. Returning it would change a
  public signature, so the docstring now points at `bundle_receipt_check`
  instead. Worth a maintainer's decision.
- `render_proof` renders `claim['profile']` and `claim['capability']` into
  the badge, so a bundle author can still shape that text. Pre-existing, and
  tightening it needs the same care that `75e86d7` above got wrong.

Verification: full suite 856 passed; merged with all fourteen sibling
branches, 1,505 passed on Python 3.11, 3.12 and 3.14.

src/nandatown/receipt.py
JamesCarnley: Verified addressed at `75e86d7f43ce9eea7d02cb7e6fbf6cffeeb16607`. The missing-replay disclosure is included in signed limitations even with custom caller limitations, survives detachment, and is printed as `receipt states:`. Removing it without re-signing fails verification. Independent signed control-character/bidi injection cases are refused, while quoted Unicode evidence and ordinary partial/failed receipts remain verifiable without earning Town Proof. Matching-evaluator result forgery still fails replay. Resolving this portability recommendation; signature validity is not a certification that every signer-supplied limitation is complete.

src/nandatown/receipt.py
JamesCarnley: Verified fixed at `75e86d7f43ce9eea7d02cb7e6fbf6cffeeb16607`. I generated a fresh voting bundle with the actual `cb19e0e` producer: its original verifier accepts it, the previous PR head rejects its profile binding, and this head accepts it with the missing-replay disclosure. Four independently rehashed raw-profile tampering controls still fail the binding check. The new-head receipt/bundle tests passed (102), full suite passed (856), and 23 independent receipt controls passed. I also reran the independent controls with #264–#278 combined; that combined suite passed all 1,505 tests. Resolving the historical-profile regression.

PR #268, opened 2026-09-11
fix(runner): wait for an external buyer's verdict before finishing

JamesCarnley:

Commits: `4e3006c`, `b7126ef`, `1825e9b`

### Problem

When the buyer joins from outside (`test-agent --role buyer --wait`,
`run --agent buyer=external`), the runner finished as soon as the stock seller
acked. A correct buyer that took about 0.5 s was reported INCOMPLETE ("the buyer
made no correctness assertion") and exited 1.

### Change

- For an externally joined buyer, the run finishes only when both hold:
  - the seller side is done;
  - the buyer has acked the quote response with any status other than
    `retryable`.
- Otherwise the existing deadline applies.
- The grant check reuses the event fetch already made in the same loop pass.

### Compatibility

- Runner only. The evaluator, profiles, HTTP contract and bundle format are
  unchanged.
- Stock and `--cmd` buyers keep their stop condition, and so do seller-subject
  runs.
- A silent external buyer now waits until the stock seller exits, about 5 s
  before the timeout, and is still INCOMPLETE. The README says when such a run
  ends: `run` uses 45 s, and `test-agent --timeout` defaults to 60 s.

### Verification

- Full suite on the branch: 820 passed, twice.
- Subprocess regressions:
  - the `--wait` case fails on `df0b5f1`;
  - `--cmd` and silent-buyer guards.
- A standalone external buyer driven through `--wait` passes with 0.5 s,
  2.0 s and no think time.
- With an external buyer, four fault profiles pass. They were INCOMPLETE on
  `df0b5f1`.

### Also fixed: one-shot seller race (commits `d8fdc6e`, `177b06a`, `82b3757`)

**Problem.** A one-shot seller subject could answer, ack and exit 0 before
Town's stock buyer claimed the reply. The runner then stopped the buyer and
reported INCOMPLETE ("no quote response accepted and claimed by the buyer"),
although the seller had done its job.

**Fix.** Once a quote response addressed to the buyer has been accepted, the
runner lets a buyer it started finish on its own, within the same deadline. A
response sent to anyone else, such as the seller replying to itself, does not
extend the run: that case ends in 0.7 s, where it took 46 s before `82b3757`,
and is still INCOMPLETE. These behave as before:

- crash restart;
- a seller that exits without responding, which still ends promptly as
  INCOMPLETE;
- external buyers;
- stock-vs-stock runs.

**Tests.**

- A deterministic regression test (the stock buyer's claim is delayed by 1 s)
  failed before the fix and passes after it.
- A 25-run one-shot loop passes every time.
- All eight stock profiles pass.
- `177b06a` gives this branch's silent-buyer test a 20 s timeout, so it stays
  valid alongside the timeout-floor PR.


src/nandatown/runner.py
```diff
@@ -548,7 +591,15 @@ def spawn_seller() -> subprocess.Popen | None:
                     else:
                         post_event("runner", "participant_exited",
                                    "seller", {"exit_code": rc})
-                        break
+                        if buyer is None or not _response_accepted(
```
JamesCarnley: **[P2] Keep waiting when a one-shot seller has answered an external buyer**

`buyer is None` immediately exits even when the seller has already queued its response and the external buyer is still evaluating it. With a correct one-shot seller, `wait_timeout=30`, and an external buyer that claims the reply then waits 1.5 seconds before its valid acknowledgement, this head finishes INCOMPLETE (`response=passed`, `correct` missing) and closes the coordinator before the acknowledgement. The identical buyer launched as a managed command passes.

This preserves a pre-existing early-cutoff case across the two fixes in this PR. Please let an external buyer settle an accepted response (or reach the deadline) after seller completion, while retaining the no-response and restart guards. Add the combined one-shot-seller/external-buyer regression; the review body records the reproduction and control.

JamesCarnley (commented):
Independent review at `82b3757ab9ab1e8f8cd878ff1364672d93cf32f4`, base `df0b5f1fd0c2ec2f9535c786fcf83694096e4e9f`.

The external-buyer wait and one-shot-seller repair each improve the real-agent workflow, but their combination still ends a correct run early. This is an **incomplete fix of the existing seller-exit cutoff**, not a claim that the PR introduced the original race.

Reproduction using only synthetic localhost participants:

1. Run `run_town("quote-clean", ..., wait_timeout=30)`.
2. Supply a one-shot seller command that claims `q-1`, sends `r-1` to the buyer with `request_id="q-1"` and the correct total, records its processed/applied acknowledgement, and exits 0.
3. Supply `external["buyer"] = None`; use `on_credentials` to launch an outside buyer with those run credentials.
4. The buyer sends the normal request, claims the response, waits **1.5 seconds** (within both the 30-second run timeout and 5-second lease), then acknowledges it as correct.
5. Control: run the identical buyer as a managed replacement command instead of via the external join.

Observed on this head:
- Managed command buyer: **PASSED**, including its correctness acknowledgement.
- Externally joined buyer: **INCOMPLETE** with `response=passed` and `correct=not_enough_evidence`; the coordinator has already closed, so the valid delayed acknowledgement gets `ConnectError`.
- The accepted response and seller's applied acknowledgement are present before the runner finishes. This is not a silent buyer or a seller that left without answering.

The `buyer is None` arm at lines 594–596 bypasses the new external-buyer settling check as soon as the one-shot seller exits. Please carry the seller-completed state forward when its response is queued for an external buyer, and wait for that buyer's terminal acknowledgement or the bounded deadline. Preserve prompt termination when no response was sent to the buyer and preserve crash-restart behavior.

Verification: the focused reviewer ran 21 tests on #268, 3 on #269, and 24 on #274; a clean temporary integration of the three exact heads passed 48 focused tests plus 3 end-to-end tests. The interaction reproduced on both #268 alone and that integration. I separately reran the managed/external control pair on this exact #268 head and confirmed the difference. These are scoped synthetic tests, not independent real-agent adoption evidence.

No source changes, commits, or merges were made by this review.

JamesCarnley:
Fixed in `19874fa`. Your reproduction was accurate, and the managed-buyer
control was the useful part of it.

The seller-exit guard read `buyer is None or not _response_accepted(...)`, so
the `buyer is None` arm short-circuited before the response check and ended
the run the moment a one-shot seller left. That clause is gone. An external
buyer is not a buyer with nothing left to do; it is one the runner only sees
through its acknowledgements, so it now gets the rest of the same deadline,
and the loop's existing external-buyer exit condition ends the run as soon as
it settles.

Three regressions, all with a one-shot seller:

- `test_one_shot_seller_does_not_cut_off_an_external_buyer`, parametrized over
  the `--wait` handoff and the managed command, so the control you used is
  pinned alongside the case that failed.
- `test_one_shot_seller_with_a_silent_external_buyer_ends_on_the_deadline`,
  which keeps the wait bounded: nothing acknowledges, and the run ends
  INCOMPLETE on its own timeout with the missing evidence named as the
  buyer's.
- `test_misaddressed_one_shot_seller_ends_promptly_with_an_external_buyer`,
  which keeps prompt termination when nothing ever reached the buyer's inbox.

The crash-restart path is untouched.

Verification: the five one-shot-seller tests pass; full suite 827 passed.

### After independent review

A reviewer measured the change and found it fixes more than I claimed — it
also repairs `quote-drop-wakeup` and a non-zero seller exit, both
incomplete→passed. It also found that I had left half the problem in place,
now fixed in `d0e8c17`.

Once the seller has exited, the external-buyer exit condition still asked
whether the seller side had finished, which is judged from the seller's
acknowledgements — something a departed seller can never add to. So a run
whose buyer had already settled sat out the rest of its deadline, and then
the whole 8 s quiet loop, which re-checks the same thing. At `wait_timeout`
20 that is 28 s instead of 9, for no further evidence. The `buyer is None`
clause had been masking it.

The realistic case is the one that matters: `town-protocol.md` asks a
participant for a status and a note of its own observations, and never
mentions `applied`, so a subject written to the documented protocol never
satisfies the acknowledgement form at all. That is now a regression test with
a seller acking exactly what the protocol asks for.

Two related corrections:

- The README sentence added earlier on this branch said such a run ends when
  the stock seller exits about 5 s before the timeout. That is no longer
  true; it ends when the buyer settles, or at the deadline.
- The commit message said the buyer gets "the rest of the same deadline",
  which it was never told. `hand_off` omitted `DEADLINE` while promising "the
  same environment a spawned participant gets". Fixed in #269, since that is
  the PR that owns the handoff.

The reviewer also noted that `run --agent buyer=external` prints no
credentials at this base, so the combination this PR repairs has no CLI entry
point yet. #269 supplies it.

Verification: 27 tests in `tests/test_harness_overrides.py` pass, 15
consecutive runs green in the reviewer's check; full suite 829 passed; all
eight Track profiles pass and verify with no timing regression.

src/nandatown/runner.py
JamesCarnley: Verified fixed at `d0e8c17841b16f4ce4ad69f68ed16aad900111a3`. The original one-shot seller + 1.5-second delayed buyer reproduction is red on the prior head (external buyer INCOMPLETE, late acknowledgement rejected) and green here: managed and external buyers both pass, with the external child exiting 0. Retryable-then-processed acknowledgements are retained; a rejecting buyer remains non-pass. All 27 harness tests and 3 stock end-to-end tests passed. I independently repeated the original reproduction in the combined #264–#278 tree; both modes pass there too, and its full suite is 1,505 passed. Resolving the early-cutoff finding.

PR #269, opened 2026-09-11
fix(cli): print join credentials for run --agent ROLE=external

JamesCarnley:

Commits: `a5ef034`, `8ee2a25`, `abe2143`

### Problem

The README says `run --agent ROLE=external` hands out join credentials. It
printed none, so the run timed out INCOMPLETE. Separately, `test-agent --wait`
didn't flush, so through a pipe (CI, `tee`) the credentials appeared only after
the run ended.

### Change

- Both commands use one shared, flushed credential printer.
- Values are `shlex.quote`d, so paths with spaces and grant JSON can be pasted
  into a shell safely. A unit test round-trips them through `shlex.split`.
- The tests live in `tests/test_external_credentials.py`. On failure they
  interrupt the CLI before killing it, so no coordinator is orphaned.
- The README now says a pinned role gets `TOWN_GRANT` in addition to `TOKEN`,
  and that the town refuses the token for a pinned role.

### Compatibility

- The same variables are printed.
- Tokens, loopback URLs and ordinary paths print exactly as before.
- `run_town` is unchanged.

### Verification

- Full suite on the branch: 820 passed, twice.
- A subprocess test reads the CLI's piped stdout. The credentials must appear
  while the run is still waiting; the reference seller then joins from them and
  the run passes. The test fails on `df0b5f1`.
- A real shell paste under both `bash` and `zsh` works with `--identity` and a
  path containing a space.


JamesCarnley:
Added `4383c50`, from a reviewer working on #268.

`hand_off` promised "the same environment a spawned participant gets" and
handed over five of the seven variables. A spawned participant is told
`DEADLINE`, the seconds the run will wait; an agent joining from outside was
told nothing, so it had no way to know when the coordinator would close and
its acknowledgement stop being recorded. #268 makes the runner wait for an
external buyer's verdict, so this is the half that lets the buyer know how
long it has.

`FAULT` stays behind, and the docstring now says why instead of claiming a
parity that never existed: `FAULT` is how this town tells its own scripted
participants to misbehave on cue, and an agent joining from outside is the
subject. A subject is observed, not instructed to fail.

The docs and the `--cmd` help listed five variables for the spawned case too,
which was already short by two.

Full suite 820 passed.

PR #270, opened 2026-09-11
fix(path): record undecodable terminal output as the subject's

JamesCarnley:

Commits: `8958101`

### Problem

Under the strict Path profiles (the `@0.3` default and `a2a-quote-intent@0.2`),
some subject output was misattributed.

- JSON nested past the interpreter's recursion limit produced an ERROR verdict
  blaming Town.
- An integer over Python's 4,300-digit limit, or a lone surrogate, recorded a
  phantom second failed `protocol_exchange`, so `protocol_invocation` failed.

### Change

- Strict profiles record these cases as `fulfillment_unparseable`, so
  `protocol_invocation` passes and `semantic_result` fails.
- Only decoding and digesting are guarded, so a fault in Town's own event
  recording is not blamed on the subject.

### Compatibility

- `evaluate_path`, profiles and evaluator versions are unchanged.
- Legacy profiles keep their recorded behavior, pinned by a characterization
  test that passes on `df0b5f1`.
- Deeply nested, non-finite and unencodable echoed values are handled by
  the follow-up commits described below.

### Verification

- Full suite on the branch: 828 passed.
- Five new strict cases fail on `df0b5f1`.
- A recording-fault test guards against over-catching.
- Live localhost fixture runs: strict profiles FAILED at `semantic_result`,
  legacy `@0.2` still ERROR, and the bundles verify.

### Also fixed: deeply nested values echoed into evidence (commits `3cfd70a`, `0b097f9`)

**Problem.** Valid JSON nested roughly 200–999 levels deep inside a value that
Town copies into an event crashed `run_path_test` while it was writing the
bundle, under every Path profile, and this predates this PR. The affected
values were:

- `total_cents` and `request_id`;
- the quote-intent fields;
- the task `id`, `kind` and `state`;
- the agent card `name` and `version`.

The run left a partial bundle with no attestation and a traceback. It exited 1,
the same code as "failed", which hid a Town fault.

**Fix.** Any such value nested more than 64 levels deep is now recorded as a
small marker, `{"unrecorded": "nesting too deep", type, json_length,
fingerprint}`. The limit of 64 was chosen from these measurements:

- pydantic refuses to write an event line past about 254 levels;
- it refuses to read one back past about 198;
- the value itself already sits up to three levels deep in the line.

Values within the bound are recorded exactly as before, for legacy and strict
profiles. `content_digest` and the card digest still cover the full output.

**How the marker is judged.** It is judged the way the real value would be: it
never equals an expected value.

- A nested `kind` fails `protocol_invocation` as the agent's fault.
- A nested `state` does the same under the strict profiles.
- A nested `total_cents` fails `semantic_result`.
- No case becomes a Town ERROR.

This replaces the note above saying that this limit "is not addressed".

**Also marked (`4af3e16`): NaN, ±Infinity and lone surrogates.** Before this
commit, both kinds of value broke evidence in these ways:

- Python's JSON parser accepts `NaN` and `Infinity`, but pydantic recorded them
  as `null`. Replay then differed from the original run. For example, a legacy
  run with a NaN state passed, but `verify` later rejected its bundle.
- A lone surrogate crashed the bundle write. This also affected the
  unparseable-output preview.

Both are now recorded as markers:

- a non-finite number becomes `{"unrecorded": "non-finite number", …}`, with a
  deterministic fingerprint;
- an unencodable string becomes `{"unrecorded": "unencodable string", …}`, with
  no digest.

The rule is applied uniformly, including to the card name and version. For
those rare NaN inputs, the recorded bytes change from a misleading `null` to
the marker, and stage outcomes are unchanged. Ordinary bundles stay
byte-identical.

**Independent review.** The reviewer ran 1,124 adversarial runs (12 fields ×
23 values × 4 profiles), comparing this branch with `main`:

| | Crashes | Bundles `verify` rejects |
|---|---|---|
| `main` | 214 | 220 |
| this branch | 46 | 0 |

- On this branch there are no Town ERROR verdicts.
- The NaN fingerprint is identical on Python 3.11, 3.12 and 3.14.
- The 46 remaining crashes are all U+2028-style line separators in recorded
  strings. The separate bundle-reader PR (`fix/bundle-jsonl-line-separators`)
  fixes those. With both merged, the same probe gives 0 crashes and 0 rejected
  bundles.

**Verification.**

- New tests at depths 65, 210, 300 and 900 for every echoed field, on legacy
  and strict profiles. They fail before these commits and pass after.
- Full suite: 956 passed.
- A real localhost fake agent that nested each field 300 levels deep: no
  traceback, and every bundle verifies.


src/nandatown/path_runner.py
```diff
@@ -307,27 +307,18 @@ def run_path_test(subject_url: str | None, out_dir: str,
...
-                                if field in fulfillment}
-                        recorder.emit(
-                            "town-requester", "fulfillment_observed",
-                            order_id, detail)
-                    except (json.JSONDecodeError, TypeError) as exc:
-                        if isinstance(exc, TypeError) and not strict_semantics:
+                        content_digest = (fingerprint(fulfillment)
+                                          if isinstance(fulfillment, dict)
+                                          else None)
+                    except (ValueError, TypeError, RecursionError) as exc:
```
JamesCarnley, line 421: **[P2] Make the diagnostic preview serializable after catching UnicodeEncodeError**

The widened exception handler now catches a fulfillment fingerprint failure caused by a literal lone surrogate in decoded artifact text, but lines 323–328 copy that surrogate directly into the event preview. `write_bundle()` then raises `PydanticSerializationError` at `e.model_dump_json()`, losing the completed evidence bundle. An outer JSON envelope can carry this as an ASCII `\ud800` escape; the existing fixture only covers a backslash-u sequence remaining inside the inner text.

I reproduced this under both strict profiles: the same four escape-layer controls all return verifiable failed bundles on base; on this head the two literal-surrogate cases crash. Please sanitize/escape the bounded preview before recording it and test both layers, retaining `protocol_invocation=passed`, `semantic_result=failed`, and a verifiable bundle. Full reproduction is in the review body.

JamesCarnley (commented):
Independent follow-up review at `8958101a30cd2f9e061a6e9bbb00fdda02672a88`, base `df0b5f1fd0c2ec2f9535c786fcf83694096e4e9f`. The stage-attribution goal is appropriate, but I reproduced a new evidence-writing regression that needs fixing before merge.

An ASCII JSON response can escape a lone surrogate in the outer A2A artifact text. After decoding the outer envelope, `text` then contains the literal surrogate. Inner JSON decoding succeeds; `fingerprint(fulfillment)` raises `UnicodeEncodeError`. The widened handler catches it but copies the same unserializable character into `fulfillment_unparseable.detail.text`. Later `write_bundle()` crashes at `TownEvent.model_dump_json()`, with `PydanticSerializationError`, instead of returning a complete verifiable FAILED bundle.

The existing test covers a backslash-u escape *inside* the terminal text. It does not exercise a literal surrogate *after the outer envelope has been decoded*. Both strict profiles are affected.

Reproduction test (synthetic HTTP MockTransport; no live service):

```python
import json
import httpx
import pytest
from nandatown.a2a_adapter import build_agent_card
from nandatown.bundle import verify_bundle
from nandatown.path_runner import run_path_test

@pytest.mark.parametrize("profile", [
    "a2a-capability-fulfillment@0.3", "a2a-quote-intent@0.2",
])
@pytest.mark.parametrize("escaped", [True, False])
def test_surrogate_preview(tmp_path, profile, escaped):
    url = "http://127.0.0.1:8940"
    value = "\\ud800" if escaped else chr(0xd800)
    text = '{"request_id":"' + value + '","total_cents":3990}'

    def handle(request):
        if request.method == "GET":
            return httpx.Response(200, json=build_agent_card(url))
        result = {
            "id": "task", "kind": "task",
            "status": {"state": "completed"},
            "artifacts": [{"parts": [{"kind": "text", "text": text}]}],
        }
        return httpx.Response(
            200, content=json.dumps({"result": result}).encode("ascii"),
            headers={"Content-Type": "application/json"},
        )

    with httpx.Client(transport=httpx.MockTransport(handle)) as http:
        directory, result = run_path_test(
            url, str(tmp_path), profile_ref=profile, http=http,
        )
    assert result.verdict == "failed"
    assert verify_bundle(directory) == []
```

Verification: using the same Python 3.12 environment and `PYTHONPATH` pointed at each exact source snapshot, `python -m pytest -q <repro-test> -k surrogate --tb=short` gave **4 passed on base**, versus **2 passed / 2 failed on this head**. I independently reran both comparisons after the focused reviewer found the issue. The focused reviewer also ran the existing Path suites on this head: 48 passed.

Please make the bounded diagnostic preview safe to serialize and add both escape-layer cases. Preserve the intended successful protocol exchange / failed semantic-result attribution and verify the resulting bundle; do not just suppress the final write error. No product code was changed in this review.

src/nandatown/path_runner.py
JamesCarnley: Verified fixed at `4af3e1673eb5cf9aa3986879d6d558db2872b470`. I reran the original escaped/literal-surrogate cases across both strict profiles plus the real localhost HTTP reproduction: **5 passed**. The result now remains a failed, independently verifiable bundle instead of crashing during serialization.

The exact-head full suite also passed (**1,156 tests**, Python 3.12). I separately tested the clean combination with #276 at `eda0d595c7a9f8054d8c911cf93864de7dc1332f`: **1,170 full-suite tests passed**, plus 9 independent controls for Unicode record boundaries, large numeric output, and terminal-preview preservation.

Resolving this specific Unicode finding. This is not an approval or a claim that the whole PR batch has been integration-tested.

PR #271, opened 2026-09-11
fix(path): fail resolution on a malformed local index

JamesCarnley:

Commits: `68596c0`

### Problem

A malformed local index fixture crashed `test-agent --index` with a traceback
and wrote no evidence. That covered:

- a list at the top level or under `agents`;
- a string entry;
- a non-string url;
- a file that isn't UTF-8;
- JSON nested past the recursion limit.

A non-string `card_digest` also crashed `evaluate_path`, and would crash
`verify` replay.

### Change

`_resolve` checks:

- the top level;
- `agents`;
- the entry;
- `url`, which must be a non-empty string;
- `card_digest`, which must be a non-empty string.

It treats undecodable files as unreadable. A violation records the existing
`resolution_failed` event, with a short reason that doesn't echo the value.

### Compatibility

- Well-formed indexes are unchanged.
- The evaluator, profiles and versions are unchanged.
- An empty `card_digest` used to silently disable the pin; it is now refused.

### Verification

- Full suite on the branch: 828 passed.
- The malformed shapes run through `run_path_test` and the CLI. Each gives
  resolution Failed, later stages not tested, and a bundle that verifies. They
  fail on `df0b5f1`.
- A real CLI run against a non-UTF-8 index exits 1 with no traceback.

### Also fixed: unusable endpoint URLs and blank agent names (commits `a4e84c9`, `dcb8039`, `8cabd1c`, `28f25ff`)

**Problem.** Some endpoint URLs were never usable, but passed resolution
anyway, from either `--url` or an index entry:

- blank or padded strings;
- `file://` or scheme-less strings;
- ports outside 1–65535.

They were then charged to the agent's `agent_card_retrieval` stage, although it
was the operator's locator that was wrong. A URL that httpx rejects outright,
such as one with a trailing newline, crashed `test-agent` with a traceback,
also on plain `--url`. A whitespace-only `--url` or `--agent-name` produced a
bundle that `nandatown verify` rejected ("run participant 1 has no valid
name").

**Fix.** `_resolve` requires an absolute http(s) URL that httpx can parse, with
a host, a port of 1–65535 or none, and no whitespace. Anything else records the
existing `resolution_failed` event, with a reason that doesn't echo the value.
A blank locator is recorded as subject `"?"`, so the bundle verifies and can
get a receipt.

A blank, whitespace-only or missing agent name now fails resolution before the
index is read. Previously, an index that listed a whitespace-only name could
produce a signed passing receipt whose subject was `?`. Non-blank names behave
exactly as before.

`28f25ff` adds three smaller fixes:

- **Blank or non-string subjects.** Resolution events now name such a subject
  `?`. A non-string name or URL passed to the library fails resolution with a
  verifiable bundle; before, it crashed.
- **UTF-8 index.** The local index is read as UTF-8, so a non-ASCII agent name
  resolves on any locale. A non-UTF-8 index is still reported as unreadable.
- **Merge-order-independent tests.** The blank-name tests use whitespace-only
  names, which #265 does not refuse. They pass whether or not #265's refusal of
  empty names lands first.

**Unchanged.** Any host is still accepted (loopback, LAN and remote), and every
usable locator is recorded exactly as before. `evaluate_path`, profiles and
evaluator versions are unchanged.

**Verification.**

- Parametrized invalid URLs for `--url` and for index entries, plus blank
  locators, fail before these commits and pass after.
- Full suite: 898 passed.
- A real CLI run of `test-agent --url $'http://127.0.0.1:9\n'` now reports
  "First broken stage: resolution" with no traceback, and the bundle verifies.


src/nandatown/path_runner.py
```diff
@@ -169,35 +169,116 @@ def intend(self, actor: str, action: str,
...
+    Raw whitespace is refused because httpx would drop it or percent-encode
+    it into a different endpoint.
+    """
+    if not isinstance(value, str) or any(ch.isspace() for ch in value):
+        return False
+    try:
+        parsed = httpx.URL(value)
+    except httpx.InvalidURL:
+        return False
+    return (parsed.scheme in ("http", "https") and bool(parsed.host)
```
JamesCarnley: **[P2] Guard hostname decoding as part of URL validation**

`httpx.URL(value)` can succeed for an ASCII punycode label whose decoded hostname is invalid. Accessing `.host` then performs IDNA decoding and raises `IDNAError`/`InvalidCodepoint` (a `UnicodeError`), outside this `try`. For example, both `http://xn--.localhost:9` and `http://xn--a.localhost:9` now crash `test-agent` without producing evidence, whether supplied directly or through the local index. Base and the previous head return a verifiable failed bundle for these inputs.

Please keep the scheme/host/port property evaluation inside the guarded validation and handle hostname-decoding errors as invalid resolution input. Add both malformed labels to the direct/index negative controls, asserting a failed resolution and verifiable bundle with no network request, while retaining a valid IDNA control such as `xn--caf-dma.localhost`.

JamesCarnley (commented):
Re-reviewed the expanded change at `28f25ff623341d5c7f2a1cde1e77eb9f75a790e9`, against base `df0b5f1fd0c2ec2f9535c786fcf83694096e4e9f` and the previously reviewed head `68596c01783c1aff646eaff3737106732da3ea96`.

The early resolution failures and verifiable failed evidence are the right scope. I found one newly introduced exception escape in the URL validator that defeats that goal for malformed ASCII punycode hostnames.

Reproduction with the ordinary CLI, without a live service:

```sh
python -m nandatown.cli test-agent \
  --url 'http://xn--.localhost:9' --out /tmp/town-invalid-host-review
```

At this head, `parsed.host` raises `IDNAError: Malformed A-label, no Punycode eligible content found` before a bundle is created. Base and the prior head instead return failed, verifiable evidence. `http://xn--a.localhost:9` similarly raises `InvalidCodepoint`. The same defect occurs when either URL is supplied by an index entry.

Independent controls covered both malformed hosts through direct and index resolution, plus valid ASCII and valid IDNA hosts. The invalid cases made zero transport calls; the valid cases continued normally and verified. Root independently reran the head/base controls. The defect also reproduces in a clean #265/#271 combination.

Focused tests: **125 passed** at this head; **130 passed** on the #265/#271 combination (Python 3.12). These are targeted results, not a full-batch integration or approval. Locale-dependent bundle-writing limitations were kept separate from this new regression.

JamesCarnley:
Fixed in `0dbcbfb`, with a follow-up in `16ea909`.

`_is_endpoint_url` parsed the URL inside its guard but read `scheme`, `host`
and `port` outside it, and httpx decodes a punycode hostname lazily when the
host is read. The reads have moved inside the guard, which now catches
`UnicodeError` as well as `httpx.InvalidURL`, so a hostname that will not
decode makes the endpoint unusable rather than a crash.

Both labels you named are covered, directly and through an index entry:
`http://xn--.localhost:9` and `http://xn--a.localhost:9` are now failed
resolutions with a verifiable bundle and no traceback. A new test asserts the
subject is never contacted, using a transport that records every request and
must see none. `http://xn--caf-dma.localhost:8940` joins the usable set as the
valid IDNA control, and both labels are also in the list that runs without an
injected client, so the real httpx path is covered too.

Verification: `tests/test_path.py` 137 passed; full suite 932 passed on
Python 3.11; and the real CLI for both labels, which now exits 1 with
`Verdict: FAILED`, writes a bundle, and `nandatown verify` accepts it.

### After independent review

An adversarial reviewer fuzzed the guard with 331,149 structured and random
punycode inputs and found no escape, and confirmed all 76,816 URLs it accepts
survive the whole httpx client path. It did find three things, two of which
led to further commits here:

- `16ea909`: the docstring overstated. httpx decodes only a hostname that
  *begins* with `xn--`, so `http://localhost.xn--a:9` is still accepted here
  and charged to the subject later. Matching that everywhere would mean
  reimplementing httpx's policy, so the docstring now says what is true. The
  same commit moved the two labels out of the test named for URLs httpx
  rejects, which is not what happens to them.
- `ba3cf52`: `a2a test` had the same class of bug. `httpx.InvalidURL` is not
  an `httpx.HTTPError`, so `probe_endpoint` raised on an address httpx cannot
  parse instead of reporting it. `b931970` then moved that test into its own
  file, where it belongs and where it does not collide with #265.

It also confirmed something already on the follow-up list rather than in this
PR: userinfo in a locator (`http://user:pass@host`) reaches `run.json`, the
events, the report and the receipt's `claim.subject`, while the CLI prints
"nothing private leaves the bundle". Redacting it touches #265, #270 and this
PR, so it is best done once they have merged.

Full suite 937 passed.

src/nandatown/path_runner.py
JamesCarnley: Verified fixed at `b931970580041963290bb4b4f02528f8791cf912`. I reran the original independent direct/index cases for both `http://xn--.localhost:9` and `http://xn--a.localhost:9`, with the transport set to fail if any request is attempted. Each now produces a verifiable failed-resolution bundle without a request or traceback. Targeted Path/probe tests plus those controls passed: 146 tests. The same independent controls also pass in the combined #264–#278 tree (full suite: 1,505 passed). Resolving the hostname-decoding regression.

PR #272, opened 2026-09-11
fix(pulse): refuse a reused --target name

JamesCarnley:

Commits: `a415aaf`

### Problem

`nandatown pulse --target svc=A --target svc=B` silently probed only B and
exited 0.

### Change

A reused name is refused with exit 2 and a message, before any probe or database
write.

### Compatibility

Distinct names are unchanged.

### Verification

- Full suite on the branch: 818 passed.
- A test with counting localhost servers checks exit 2, the message, zero hits
  and no database. It fails on `df0b5f1`.


JamesCarnley:
Added `448f14a`, from a reviewer chasing the same class of bug as #271 into
this module.

`pulse` caught only `httpx.HTTPError`, and neither of the two ways a URL can
fail late is one. httpx parses `http://xn--a.localhost:9` and then raises
when the host is read, and raises outright on an address like
`http://[v1.fe80::a+en1]:9`. Either killed the probe loop mid-pass with a
traceback, so every other target lost the rest of its schedule and the
history kept was silently truncated: with two targets and `--count 3`, one
row instead of six.

A schedule is a record kept over time, so nothing a single target may do can
end it. A URL that cannot be probed is now named rather than raised, and such
a target is recorded as down like one that cannot be reached, while the
others keep their history. The CLI refuses the target before probing
anything, exactly as this PR already refuses a reused name — a typo in an
argument should be heard about at once, not read as "down" for a week.

Before: `run_pulse` raises, `[('good', 1)]`. After: `[('bad', 3),
('good', 3)]`.

Full suite 837 passed.

PR #275, opened 2026-09-11
fix(cli): stop Track processes when the CLI receives SIGTERM

JamesCarnley:
Follow-up to #264–#272, from the same correctness audit of `main` at `df0b5f1`.
It merges cleanly with the other open PRs.

### Problem

Stopping `nandatown run`, `test-agent` or `campaign` with SIGTERM (for example,
a CI job timeout) leaves the coordinator and participants running
indefinitely.

The coordinator runs in its own session, and only `run_town`'s `finally` block
stops it. Python's default SIGTERM handling exits without running `finally`
blocks. SIGINT did run the cleanup, but a signal that arrived in the few
milliseconds between starting a process and registering it for cleanup could
still orphan that process.

### Change

- **SIGTERM cleanup.** For the three commands that call `run_town`, SIGTERM now
  takes the same cleanup path as SIGINT. Once cleanup has run, the CLI restores
  SIGTERM's default handling and re-raises the signal, so the parent still sees
  a process killed by signal 15, exactly as before.
  - A second SIGTERM cannot interrupt the cleanup.
  - The previous handler is restored when the command returns.
  - No handler is installed on non-POSIX platforms, off the main thread, or when
    SIGTERM is already ignored or handled.
- **No startup gaps.** The admin client is built before any process starts.
  Each Town process is started and recorded for cleanup inside a short guard
  that holds SIGINT and SIGTERM until the process is registered, then re-raises
  them. Ignored signals stay ignored. While the previous handlers are put
  back, both signals are briefly blocked in the calling thread, so a signal
  arriving mid-restore still reaches its restored handler exactly once.

### Compatibility

- Exit status for SIGTERM is unchanged: death by signal 15.
- SIGINT behaves as before, except that it can no longer orphan a process
  during startup.
- The server commands (`coordinator`, `a2a serve`, `ui --web`) behave as before.
- One known gap remains: the terminal UI runs `run_town` in worker threads, so
  SIGTERM there still skips cleanup.

### Verification

- **Real-process tests.** New `tests/test_cli_signals.py` starts the real
  `test-agent --wait`, records every process the CLI started, sends SIGTERM or
  SIGINT, and asserts they are all gone and the exit status is `-signum`. It
  cleans up leftovers even when an assertion fails.
  - On `main`, the SIGTERM case orphans the coordinator and the stock buyer. It
    passes with this change.
  - It passed 20 of 20 loop runs.
- **Timing harness.** Signals were injected at startup, just after each start
  call returns, during the health wait, while waiting for an external agent,
  with a `--cmd` seller and its child, during campaign trials, as a second
  SIGTERM during cleanup, and at finish. Each case now leaves 0 processes and
  exits `-15` (or `-2` for SIGINT). Before this change, several left the
  coordinator or seller running.
- Full suite: 825 passed, twice.


JamesCarnley (commented):
Reviewed `486bc89fdc8b81b0f6af2162511165d602459cfd` against `df0b5f1fd0c2ec2f9535c786fcf83694096e4e9f`.

The main fix is useful and aligned with Town's local subprocess-testing workflow: independent `run`, `test-agent`, and `campaign` controls, repeated SIGTERM, and interruption at the third spawn exited with signal 15 and left no managed children running. The equivalent base controls left children behind. Ordinary custom/ignored signal handlers, body exceptions, and worker-thread behavior also passed.

Verification: `python -m pytest -q tests/test_cli_signals.py` — **11 passed**; exact-head `python -m pytest -q` — **828 passed, 8 warnings**, Python 3.12. All diagnostics used disposable data and synthetic local processes. #268/#274 interactions were source-reviewed, not executed as a combined tree; Windows was not tested.

One narrow exception-path issue is below. This concerns an embedding caller with a raising signal handler, not a claim that ordinary CLI SIGTERM cleanup is broken. It can be reproduced by strengthening the exception case already in this PR; no probabilistic signal timing is necessary.

src/nandatown/runner.py
```diff
@@ -88,6 +90,55 @@ def _spawn_participant(command: list[str], url: str, run_id: str, name: str,
...
+        # place for SIGTERM. The mask is per thread, so this covers a caller
+        # without other threads, such as the CLI. No process starts while
+        # the signals are blocked, so none inherits the blocked mask.
+        mask = None
+        try:
+            if saved and hasattr(signal, "pthread_sigmask"):
+                mask = signal.pthread_sigmask(signal.SIG_BLOCK, [])
+                signal.pthread_sigmask(signal.SIG_BLOCK, list(saved))
+            for signum, previous in saved.items():
+                signal.signal(signum, previous)
```
JamesCarnley: **[P2] Restore caller handlers when mask setup raises**

If either `pthread_sigmask` call raises, the handler-restoration loop is skipped and the `finally` below restores only the mask. `_stop_signals_held()` then leaves its recording lambdas installed, so an embedding caller that catches the exception can silently lose subsequent SIGINT/SIGTERM cancellation.

The new `test_holding_unblocks_stop_signals_when_blocking_them_raises` already reproduces this exact path: append assertions that `signal.getsignal(signum) is handler` for each saved handler before its cleanup, and both fail on this head. I independently confirmed `SIGINT=False`, `SIGTERM=False`, while `mask_restored=True`; an equivalent base control retains the caller's handlers. The test's own `finally` currently hides the leak.

Please make saved-handler restoration run on the mask-setup exception path too, and extend that test to assert both handler identities and subsequent delivery. This is a bounded exception-safety request, not a requirement to withstand arbitrary repeated asynchronous exceptions during cleanup.

JamesCarnley:
Fixed in `55850cd` and `4874380`.

You were right that the leak is the important half. Blocking makes the
restores atomic, but restoring is what makes holding the signals safe at all,
so the restore loop now lives in its own `finally` around the mask setup: if
either `pthread_sigmask` call raises, the caller's handlers still go back,
unblocked, and the outer `finally` still restores the mask.

I took the test extension you asked for rather than writing a new one from
scratch: `test_holding_restores_caller_handlers_when_blocking_them_raises`
installs the caller's own handlers, asserts `signal.getsignal(signum) is
handler` for both after the failure, and then raises both signals to prove
they are actually delivered to the caller. It fails on the previous head at
the first identity assertion.

Verification: `tests/test_cli_signals.py` 12 passed; full suite 829 passed.

### After independent review

A reviewer measured the leak before and after and found the fix real but
partial: 2 signals leaked before, 1 after. Restoring unblocked means a stop
signal can land between two restores and run the handler just put back, and
if that handler unwinds, the loop stopped there and every handler after it
stayed recording into a list belonging to a block that had already exited.
The same happened if `signal.signal` itself raised.

`4874380` makes each restore independent and re-raises the first failure once
they have all been attempted. Two more regression tests cover it, and they use
a fake that raises *before* blocking takes effect — the reviewer correctly
pointed out that the existing fake calls the real `pthread_sigmask` first, so
the unblocked path this commit adds had no coverage at all.

That commit also records something pre-existing you may want to keep in mind:
a caller handler that unwinds during the mask restore drops the held signals,
exactly as one unwinding in the raise loop below does. The comment documented
only the second case. Behaviour is unchanged; the caller is unwinding on an
equivalent signal either way and `run_town`'s cleanup still runs.

Verification: `tests/test_cli_signals.py` 14 passed, and 25 consecutive runs
green in the reviewer's own check; full suite 831 passed; real SIGTERM to
`run` exits by signal with no leftover children.

src/nandatown/runner.py
JamesCarnley: Verified fixed at `48743808f92d5f3f164ff87944a5f2fee1fa6e34` for this bounded mask-setup exception path. Independent real-signal controls inject before and after both the original-mask query and stop-signal blocking: the prior head loses both caller handlers; this head restores both handlers, the mask, and subsequent SIGINT/SIGTERM delivery in all four cases. If one handler restoration itself raises, the other is still attempted and the exception propagates. All 14 signal tests passed, and subprocess cleanup controls left no surviving children. I independently repeated the boundary controls in the combined tree. Resolving this finding; this does not claim immunity to arbitrary asynchronous interruption inside cleanup itself.

PR #278, opened 2026-09-11
fix(track): report an application whose acknowledgement was fenced

JamesCarnley:
Follow-up to #264–#277, from the same correctness audit of `main` at `df0b5f1`.
It merges cleanly with every other open PR in that batch.

This one was found by chasing an intermittent CI failure instead of re-running
the job.

### Problem

A seller can apply a quote request and send its reply, then have its
acknowledgement refused as a stale fence — or die outright — before the town
records it. Either way the redelivery was acknowledged as a bare duplicate, so
the `processed` stage reported "processed acknowledgements carry no application
record" and the run came out `incomplete`.

That blames the subject for the town's own lease timing, although the work was
applied exactly once. In the `quote-crash-restart` profile the lease is 1.5 s,
and on a slow machine the restarted seller can lose it between applying the work
and acknowledging it.

### Change

- The journal row carries an `unreported` mark, written in the same statement as
  the application, before any acknowledgement is attempted. The mark therefore
  exists from the instant the work does, so a death that nothing could observe
  cannot lose it.
- Only an accepted acknowledgement clears it.
- A redelivery of still-unreported work is acknowledged as a duplicate that also
  carries the application actually performed.
- The inbox loop gained one optional accepted-only callback. Its refusal path
  does no bookkeeping at all.

### Older journals upgrade in place

The mark is a new column on the journal's `processed` table, and
`CREATE TABLE IF NOT EXISTS` does not alter a table that already exists. So a
journal file written by an older Town would have crashed the participant with
`no such column: unreported`.

Normal runs use a fresh state directory each time, but `test-agent --wait` and
`run --agent ROLE=external` print a `STATE_DIR` that an operator is told to
replace with their own writable directory, so one reused across an upgrade would
have hit this. A participant crashing that way is exactly the "Town fault that
looks like the subject's" failure this PR removes.

The journal now adds the column when it is missing. Existing rows read as
reported, which is correct: those applications were acknowledged under the older
code, or are gone with the run that made them, so the upgrade cannot manufacture
a second application record. The migration also protects the buyer, the A2A
bridge and the model harness, which all build journals even though only the
seller sets the mark.

### Compatibility

- The coordinator HTTP contract, the bundled profiles, the bundle format and the
  evaluator are all unchanged.
- Exactly-once application still holds. The journal still prevents a second
  apply, and a duplicate whose application already reached the record is
  acknowledged without an application record, so an ordinary duplicate delivery
  still shows exactly one.
- `record()`'s default leaves the buyer, the A2A bridge and the model harness
  unaffected.

### Verification

- **Regression tests** cover the fenced acknowledgement, a crash between applying
  and acknowledging, two consecutive fences, and the unchanged duplicate case.
  The crash-window test fails on an earlier design that only reacted to a
  refusal, and passes on this one.
- **Delay sweep** around the 1.5 s lease (1.0, 1.4, 1.6, 2.5 and 5.0 s): 3/3 each.
- **Crash-restart end to end:** 30/30.
- **All eight Track profiles** pass, and so does `examples/byoa_seller.py`.
- **`nandatown verify`:** 36/36 bundles.
- **Full suite:** 823 passed, twice.

### Known limitation

Duplicate delivery *with the applied acknowledgement fenced* ends `incomplete`
at `duplicate_recognized`. The single redelivery acknowledgement carries both
`applied` and `duplicate`, which satisfies the runner's quiescence heuristic, so
the run ends before the town re-offers the work.

- It needs a stall of more than 5 s, which no shipped profile can produce.
- It is not a regression: `main` fails the same case earlier, at `processed`.
  This change strictly improves that stage.
- Fixing it means changing the runner's stop condition, which is outside this
  participant-side change.


src/nandatown/participants/seller.py
```diff
@@ -33,8 +33,16 @@ def handler(claim: dict[str, Any]):
...
-            reply = journal.get(message_id)["reply"]
-            return "processed", {"duplicate": True}, [reply]
+            done = journal.get(message_id)
+            note = {"duplicate": True}
+            if journal.unreported(message_id):
+                # The application happened, but no acknowledgement of it
+                # was ever accepted: the fence died first, or this seller
+                # did. Report the work actually done rather than let the
+                # evidence blame the seller for the lease timing.
+                note["applied"] = True
```
JamesCarnley, line 56: **[P2] Reconcile an accepted ack before reporting the application again**

`unreported=True` does not prove that no acknowledgement was accepted: the coordinator can commit the first `applied=True` ack and the seller can die before `on_ack_accepted` clears this bit. After restart, the shipped `duplicate_delivery` path reoffers that completed request and this branch sends a second `applied=True`. The unchanged evaluator then reports `applied 2 times, expected once`, although the journal contains one application and the second handler invocation only resends its stored reply.

I reproduced base passed/head failed with both the in-process coordinator and a real localhost seller process exiting immediately after the successful ack response. Normal delivery and a crash after the local mark both pass. Full reproduction and signed-bundle verification are in the review body.

Please distinguish unknown local reporting state from confirmed absence of a recorded application, using the coordinator's existing duplicate/accepted-record evidence where sufficient. Preserve the before-ack/fenced recovery, and add the accepted-ack-before-local-mark crash regression; do not simply clear the bit before making the acknowledgement.

JamesCarnley (commented):
Reviewed `39f1456afd70eb6f8297ad66054cff6576a91726` against `df0b5f1fd0c2ec2f9535c786fcf83694096e4e9f`.

Preserving evidence after a genuinely refused acknowledgement is useful and fits Town's stage-specific diagnosis. However, the local `unreported` bit cannot distinguish “the coordinator never accepted the ack” from “the coordinator accepted it, but the process stopped before clearing the local bit.” The second case newly turns one application into a signed, verifiable **failed** result claiming two applications.

### Reproduction

This can extend `tests/test_fenced_application.py` using its existing helpers:

```python
def test_crash_after_accepted_ack_is_not_two_applications(tmp_path):
    app, admin, run_id, tokens = make_town(
        tmp_path, fault="duplicate_delivery")
    send_quote_request(app, run_id, tokens)
    state = tmp_path / "seller"
    state.mkdir()

    def die_after_accept(inner):
        def ack(message_id, fence, status, note=None):
            result = inner(message_id, fence, status, note)
            assert result["recorded"] is True
            raise Killed("ack accepted; local callback has not run")
        return ack

    with pytest.raises(Killed):
        run_seller(app, run_id, tokens, state, ack=die_after_accept)
    run_seller(app, run_id, tokens, state)
    _, notes = seller_notes(admin, run_id)
    assert sum(bool(n.get("applied")) for n in notes) == 1
```

Head produces:
```json
[{"applied": true, "total_cents": 3990},
 {"duplicate": true, "applied": true, "total_cents": 3990}]
```

Base produces one applied acknowledgement and one bare duplicate. Completing the buyer's normal acknowledgement and evaluating the actual exported events yields **passed on base**, but **failed: applied 2 times, expected once on head**. Both resulting signed bundles pass `verify_bundle()`; the journal has one application row. This is therefore an evidence-meaning regression, not a corrupt-bundle test.

I additionally reproduced it with an actual localhost coordinator and seller subprocess: call the real HTTP acknowledgement, `os._exit(3)` immediately after its success response, then restart the same seller state. Base passes; head fails with the same two notes. These are synthetic local controls, not a claim of independent external-agent adoption.

### Verification and boundaries

- Exact-head `python -m pytest -q`: **823 passed, 8 warnings**, Python 3.12.
- `python -m pytest -q tests/test_fenced_application.py tests/test_journal_migration.py tests/test_participants.py tests/test_evaluator.py --tb=short`: **16 passed**.
- Independent three-case control (ordinary delivery; death after coordinator commit; death after local mark): **base 3 passed; head 2 passed, 1 failed**.
- A clean combination with #264 at `f2ddf9c0e9820e9cd6d03368b1d70a4436fa66c8` has the same failing control; this was not a full-batch integration run.
- `git diff --check` passes. No local packaging/dashboard build or Python 3.11 run claimed.

This is distinct from the documented “fenced application plus duplicate quiescence” limitation: here the first ack was **accepted**, and the regression is a false double-application judgment. Please retain the newly fixed before-ack/fenced cases while covering this other side of the commit boundary.

JamesCarnley:
You were right, and the framing was the useful part: `unreported` means "this
seller never saw an acknowledgement of its application accepted", which is
strictly weaker than "the town holds no record of it". I had been reading the
first as the second. Fixed in `d0e3a1c`.

The town already distinguishes them, as you suggested. `db.reoffer` only
re-offers a message whose status is already `done`, and only a recorded
processed acknowledgement produces that status, so `duplicate: True` on a
claim is the coordinator's own evidence that it holds the application record.
A plain redelivery arrives through `db.claim_next`, which only picks up
messages still at `accepted`, and is evidence that it does not. The seller now
re-reports the application only in the second case.

That keeps the before-acknowledgement and fenced paths this PR added, and
covers the other side of the commit boundary without clearing the mark early
and without a new coordinator route.

Your reproduction is in the suite as
`test_crash_after_an_accepted_ack_is_not_two_applications`, and a companion
completes the buyer's side and evaluates the exported events, so the
regression is pinned as the evaluator sees it and not only as notes: it
asserted `processed: failed, applied 2 times, expected once` before the fix.

Verification, all on this head:

- `tests/test_fenced_application.py` 6 passed, and 20 consecutive runs of
  `test_crash_restart_run_end_to_end` all passed.
- Full suite 825 passed.
- All eight Track profiles pass and verify, and `examples/byoa_seller.py`
  passes.
- Real localhost, as you did it: a uvicorn coordinator on 127.0.0.1 with the
  seller as its own subprocess calling the real HTTP acknowledgement and
  exiting on the success response, then restarted against the same
  `STATE_DIR`. Before the fix the notes are
  `[{applied}, {duplicate, applied}]` and the run is
  `failed: applied 2 times, expected once`; after it they are
  `[{applied}, {duplicate}]` and the run passes.

The `duplicate_recognized` limitation in the PR description is unchanged.

### After independent review

A reviewer traced the soundness argument independently — `db.reoffer` requires
`status='done'`, which only the non-stale branch of `db.ack` produces, in the
same transaction as the `ack_recorded` event — and probed lease expiry,
`lost_ack`, `drop_wakeup`, coordinator in-memory desync, and repeated
restarts without breaking it. Across 100 interleaving schedules that kill the
seller at five points under four fault configurations: 0 wrong verdicts on
this head against 9 on the previous one, with an identical set of INCOMPLETE
outcomes, and `applied == 1` in every one.

It found that the suppression decision was not written down, fixed in
`54a6e23`. Concluding that the town holds the application and then leaving
the journal mark set means a later delivery answers the question again.
Nothing shipped can deliver that message again, so it was not reachable — but
the fix then rested on an invariant nobody had stated, and a handler
answering the duplicate with `retryable` is enough to break it, because the
town returns that message to the pool:

- before: `processed: failed applied 2 times, expected once`
- after: `processed: passed`, `duplicate_recognized: passed`

The same commit makes the cross-module contract explicit. `Db.reoffer` now
says that "completed" is load-bearing and who relies on it, and a test pins it
from the seller's side. It also corrects `Journal.unreported`'s docstring,
which claimed `False` means an acknowledgement was accepted; it also covers a
message the journal has never seen and a row that predates the mark, and that
is the docstring the correctness argument leans on.

Three pre-existing issues the reviewer verified are unchanged by this PR, for
a future queue: a seller that dies after claiming the duplicate offer loses it
permanently and the run then waits out its whole timeout;
`duplicate_recognized` can pass on a `{"duplicate": true}` note from an
earlier fenced redelivery; and journal keys are not run-scoped, which is safe
only because the runner allocates a fresh state directory per run.

src/nandatown/participants/seller.py
JamesCarnley: Verified fixed at `54a6e23b3f816e84bfabe1b22718f400131575ba`. The original three-cut regression goes from 1 failed / 2 passed on the prior head to 3 passed. A real localhost seller process exiting immediately after the coordinator's successful ack, then restarting on the same journal, now leaves one applied note plus a duplicate-only note, one journal row, a passing verdict, and a verifying signed bundle. Before-ack, stale-fence and lost-ack recovery still pass; duplicate→retryable→ordinary redelivery does not reopen the application mark. Removing application evidence stays non-pass and fabricating a second application still fails. Full head suite: 827 passed; independent recovery and compatibility controls: 5 and 9 passed. I independently repeated the crash/recovery controls with all 15 PRs combined (1,505 full-suite passes). Resolving the commit/local-clear crash-window finding.

PR #279, opened 2026-09-12
fix(evaluator): require boolean acknowledgement flags and judge every accepted request

JamesCarnley:
Two pre-existing evaluator gaps, reported against #264 and confirmed to
reproduce on `main` at `df0b5f1`. They are not regressions from that PR or any
of #264–#278.

> **Stacked on #264.** This bumps the Track evaluator that #264 introduces, so
> it contains #264's commits and cannot merge before it. Only the last two
> commits are new here: `4e70266`, `b9ac153`, `fd50ec0` and `a9650f1`.

### The gaps

**A flag that says no was read as yes.** Acknowledgement flags were read for
truthiness, so a seller acknowledging `{"applied": "false"}` and a buyer
acknowledging `{"correct": "false"}` both counted as positive and the run
passed. The integer `1` passed too. Only an actual boolean `false` was read as
a no — the one spelling a participant is least likely to reach for by
accident was the only one that worked.

**A request nobody answered did not affect the verdict.** Only the first
accepted `quote_request` was ever judged. A buyer sending `q-1` and `q-2` with
only `q-1` answered came out `passed`, with `q-2` accepted, recorded, and
never looked at. The verdict spoke for the whole run while describing only the
part of it that went well.

Reproduced through the real coordinator with ordinary `TownClient` send,
claim and ack calls and untouched exported events. Five controls, on
`df0b5f1`:

| control | before | after |
|---|---|---|
| real booleans | passed | passed |
| `applied`/`correct` boolean `false` | failed | failed |
| `applied`/`correct` string `"false"` | **passed** | incomplete |
| `applied`/`correct` integer `1` | **passed** | incomplete |
| second request never answered | **passed** | incomplete |

### Change

- **A flag counts only when it is a JSON boolean.** `applied`, `correct` and
  `duplicate` are all read that way. Anything else states nothing this
  evaluator can read, and the stage note shows what was recorded so an
  operator can see what to fix.
- **Inconclusive, not failed.** Town did not observe the work fail; it failed
  to learn whether the work happened. Those are different findings, and
  collapsing them would blame a subject for an outcome nobody observed.
- **A count is read the same way.** The fault stages compared
  `tool_errors` and `context_truncations` with `>=`, which raises on a
  string rather than answering. A participant's note is unvalidated, so a
  seller acknowledging `{"applied": true, "tool_errors": "many"}` is accepted
  by the coordinator and then makes the evaluator raise `TypeError` — Town
  appearing to fail on evidence it holds, for a note the subject chose. A
  count now has to be an integer, and a boolean is not one.
- **Every accepted request is judged.** One never claimed, acknowledged,
  applied or answered leaves the stage it did not reach inconclusive, naming
  that request, so the run is incomplete rather than passed. A stage that
  already found something more specific about the first request keeps that
  finding.

### Worth knowing

- A run with **more than one accepted response** is already a hard failure
  under #264's rule that the profile expects exactly one. So a buyer that
  asks twice and is answered twice fails, while one answered once is now
  incomplete: the new rule only ever moves a verdict toward `incomplete`,
  and there is no multi-request success path.
- A non-boolean `correct` alongside a response mismatch now reports
  `not_enough_evidence` where `0.3.0` reported `failed`. That is deliberate —
  an unreadable flag is not an observed wrong answer — and it is stricter for
  the badge, because the stage then counts as not tested.
- A request addressed to someone other than the seller leaves the seller's
  own stages alone, since it was never the seller's to handle, and is
  reported under `response`, so the run is incomplete rather than passed.
- One readable `applied: true` beside an unreadable application claim does
  not establish application exactly once. Two real `true`s still fail, and
  the stock seller's duplicate acknowledgement, which says nothing about
  `applied`, still passes.
- `render_proof` has no evaluator-version floor, so a bundle recorded under
  superseded rules can still render `TOWN-TESTED` within its freshness
  window. That is a separate policy from replay: historical bundles can keep
  replaying and verifying while a newly rendered badge requires a current
  evaluator. Which evaluator versions may earn a badge is an open decision,
  not settled here.

### Compatibility

- The evaluator is `0.4.0`. A bundle recorded by `0.2.0` or `0.3.0` replays
  under the rules that produced it; tests pin that each of these exact runs
  still passes under both.
- No bundled profile changes. Town's own buyer sends exactly one request, and
  every shipped participant writes real booleans. All eight Track profiles
  pass and verify, as does `examples/byoa_seller.py`.
- The second gap matters when the subject is the buyer — the case where the
  run is about what the subject chose to ask.
- The coordinator HTTP contract, the profiles and the bundle format are
  unchanged.

### Verification

- Full suite: 915 passed.
- **Replay differential against genuine sources.** 208 cases — every fault,
  note shape, multi-request and correlation form, plus each case raised in
  review — evaluated by HEAD as `0.3.0` and as `0.2.0`, and by
  `git archive f2ddf9c` and `git archive df0b5f1`: zero behavioural
  differences, stage notes included. None raises under `0.4.0`, and none that
  `0.3.0` did not pass passes under `0.4.0`.
- All eight Track profiles: `PASSED`, `eval=0.4.0`, `verify` clean.
- #264's genuine `0.2.0` fixture bundles still verify, replaying under `0.2.0`.
- The five controls above, driven through the real coordinator.
- Merges cleanly with all fifteen open PRs in this batch.

### Merge note

An earlier version of this description said a `0.3.0` bundle would be refused
a receipt unless `0.3.0` were added to `SHIPPED_EVALUATOR_VERSIONS` on #266.
That was wrong. `verify_bundle` replays a recorded Track version under its own
rules whenever it is in `EVALUATOR_VERSIONS`, which includes `0.3.0` here, and
#266 consults the shipped list only when a version differs. Adding `0.3.0`
there once #264 merges is provenance bookkeeping for that fallback, not a
receipt blocker. What a `0.3.0` receipt actually depends on is exact replay,
which the next section verifies.


src/nandatown/evaluator.py
```diff
@@ -89,11 +282,21 @@ def find(ekind: str, **conds) -> list[TownEvent]:
...
+    applied = [a for a in processed
+               if (_asserted(a.detail.get("note"), "applied") is True
+                   if strict_flags
+                   else a.detail.get("note", {}).get("applied"))]
+    unreadable = (_unreadable_flags(processed, "applied") if strict_flags
+                  else [])
     if not processed:
         stages.append(_missing("processed", "no processed acknowledgement"))
     elif len(applied) == 1:
         stages.append(_passed("processed", [applied[0].event_id]))
```
JamesCarnley, line 294: [P2] Do not infer exactly-once application when another processed acknowledgement has an unreadable applied flag

`unreadable` is computed above, but this branch passes as soon as one other acknowledgement has a real `applied: true`. A source-traced, API-reachable `quote-duplicate-delivery` sequence is:

1. Seller processes the initial request and acknowledges `{"applied": true}`.
2. The coordinator reoffers that completed request under a fresh fence.
3. Seller acknowledges the reoffer with `{"applied": "true", "duplicate": true}`.
4. The normal single correlated response is accepted/claimed and the buyer acknowledges correctness.

The new filter counts only the first application, so `processed` passes here and `duplicate_recognized` also passes on the same count. 0.3.0 counted both truthy values and failed; 0.4.0 should report the second application as unknown, not upgrade the run to PASS. Exactly-once application has not been established.

Please make mixed readable/unreadable application evidence inconclusive, while preserving failure for two genuine `true` assertions and the legitimate stock duplicate acknowledgement with no `applied` field. The coordinator's duplicate reoffer and fenced ack paths permit both acknowledgements; this does not require forged event records.

src/nandatown/evaluator.py
```diff
@@ -33,8 +60,174 @@ def _missing(name: str, note: str) -> StageResult:
...
+    else:
+        return "not_enough_evidence", (
+            "the quote response carries no request_id, so it is not shown"
+            " to answer the accepted request")
+    accepted = _show_value(request_id)
+    if shape == "string":
+        # With several accepted requests, say which one this is measured
+        # against: naming it "the accepted request" would read as a claim
+        # that the one the response named is not accepted.
+        against = ("the first accepted request" if len(requests) > 1
```
JamesCarnley, line 109: [P2] Preserve the exact 0.3.0 diagnostic when replaying 0.3.0 bundles

This wording change is not version-gated: `_response_mismatch` is called for both 0.3.0 and 0.4.0. Consider two accepted requests `q-1` and `q-2`, with one accepted response naming `q-2`. At #264's head, the failed response note says `not the accepted request "q-1"`; replaying those same events with `version="0.3.0"` here says `not the first accepted request "q-1"`. The correctness-stage note also incorporates that text.

`verify_bundle` compares the full recorded/replayed result excluding only `evaluated_at`, so an authentic unchanged 0.3.0 bundle becomes an evaluator-replay mismatch even though its verdict is unchanged. Combined with #266, that mismatch also refuses a receipt.

Please gate the new wording to the new evaluator semantics and retain exact old output for 0.3.0. The existing historical controls do not cover this multi-request/mismatched-response combination. This finding follows directly from the old/new source paths; I did not run a new differential test.

src/nandatown/evaluator.py
```diff
@@ -218,6 +466,50 @@ def find(ekind: str, **conds) -> list[TownEvent]:
...
+    # something more specific keeps that finding.
+    if judges_every_request and len(accepted_req) > 1:
+        by_stage: dict[str, list[str]] = {
+            "claimed": [], "received": [], "processed": [], "response": []}
+        # These stages are about the seller's handling, so only a request
+        # addressed to the seller is the seller's to answer. Every bundled
+        # profile has two roles, so today this filters nothing.
+        for event in accepted_req:
+            if event.detail.get("to") not in (None, seller):
+                continue
```
JamesCarnley, line 484: Policy/contract clarification (not an additional introduced-regression claim): should an external buyer's misaddressed accepted request be invisible to the whole-run verdict?

The send/coordinator/database path accepts an arbitrary recipient string. Even a two-role profile can therefore contain a second `quote_request` addressed to the buyer itself or an unknown recipient. This `continue` ignores that request, and the new auditor-recipient test explicitly expects the whole run to pass. Thus "every accepted request" and "today this filters nothing" are not accurate for non-stock participants.

Not blaming the seller for someone else's work is correct. My recommendation is to keep that attribution distinction while making an extra misaddressed/out-of-profile request visible in the buyer/run result rather than allowing it to disappear behind PASS. Please clarify the intended contract; this need not become a general multi-request profile redesign.

JamesCarnley (commented):
Follow-up correctness review at `fd50ec043a239938e2f6c87dda3ee47de80fd6f7`, including two independent static reviewers. No tests or contributor code were executed for this pass.

The strict JSON boolean/count boundary, explicit inconclusive outcome for unreadable evidence, and versioned evaluator semantics are the right direction. I do **not** recommend merging this revision yet: the two inline P2 findings cover a new false PASS with mixed application acknowledgements and a genuine 0.3.0 replay mismatch. Please address those before merging; the recipient-scope thread is explicitly a policy/contract clarification, not a third claimed regression.

**Correction to the merge note:** the stated `SHIPPED_EVALUATOR_VERSIONS` receipt blocker does not arise for replayable 0.3.0 bundles in the combined code. #264's `verify_bundle` selects the recorded version when it belongs to `EVALUATOR_VERSIONS`, and #279 includes 0.3.0 there. #266 consults the shipped-version list only for an `EvaluatorVersionDiffers` fallback. Updating the historical list as versions land is provenance/future-fallback bookkeeping; it is not what fixes the replay problem identified here. Please correct the PR description accordingly.

Preserving historical receipt validity/replay and deciding which evaluator versions qualify for a newly rendered badge are separate policies. It is reasonable to leave that badge decision out of this narrow fix, but old-badge eligibility is not an unavoidable consequence of retaining historical replay.

src/nandatown/evaluator.py
JamesCarnley: Confirmed and fixed in `a9650f1`. Reproduced as you described: under
`duplicate_delivery`, `applied: true` followed by `applied: "true"` on the
re-offer is `failed, applied 2 times` at `0.3.0` and **passed** at `fd50ec0`,
with `duplicate_recognized` passing on the same count. A false pass I
introduced.

An unreadable application claim now leaves exactly-once unestablished:
`processed` is inconclusive, with a note saying so, and
`duplicate_recognized` no longer passes on it. Preserved, and pinned by
tests: two real `true`s still fail, and the stock seller's duplicate
acknowledgement, which carries no `applied` field, still passes.

src/nandatown/evaluator.py
JamesCarnley: Confirmed and fixed in `a9650f1`. You were right, and it breaks the property I
had claimed was exact: at `fd50ec0`, replaying your two-request case as
`0.3.0` says "not the first accepted request", while #264's head says "not the
accepted request". The wording is now `0.4.0`'s alone.

This time I checked rather than reasoned. A differential over 208 cases —
including this one, the mixed-`applied` case and the misaddressed requests —
evaluated by HEAD as `0.3.0` and `0.2.0`, against `git archive f2ddf9c` and
`git archive df0b5f1`. It found one more break of the same kind that neither
review had raised: the guard for a note that is not an object was also
ungated, so a note genuine `0.3.0` raised on replayed as `incomplete`. Only a
crafted bundle can carry one, but exact is exact, so that is gated too. After
both: zero behavioural differences on either version, stage notes included.

Thanks as well for the correction on `SHIPPED_EVALUATOR_VERSIONS`. I have
checked it and you are right: a `0.3.0` bundle replays under its own rules and
never reaches the shipped list, so exact replay is what its receipt depends
on. The description now says so and withdraws the earlier claim.

src/nandatown/evaluator.py
JamesCarnley: Taken, in `a9650f1`, as the contract James already chose for this PR —
every accepted request is judged — which my `continue` had quietly
contradicted. The auditor test asserting a pass was asserting the wrong
thing.

A request addressed to someone other than the seller now leaves the seller's
`claimed`, `received` and `processed` stages alone, since it was never the
seller's to handle, and is reported under `response` naming the request and
its recipient, so the run is incomplete. That covers both a third recipient
and the buyer addressing itself. A request that is addressed elsewhere but
still answered by the run's one response is already caught by the response
correlation check.

I also agree with your framing of the badge, and the description no longer
calls old-badge eligibility a consequence of replay: they are separate
policies, and which evaluator versions may render a badge is still open.

src/nandatown/evaluator.py
JamesCarnley: Verified by static source review at a9650f1 and merged main 00a013b: an explicit unreadable application claim now prevents exactly-once PASS even beside one boolean true; duplicate recognition also requires no unreadable claim. The regression definitions retain the two-real-applications failure and the stock duplicate acknowledgement control. Resolving this original finding. No tests were rerun in this follow-up.

src/nandatown/evaluator.py
JamesCarnley: Verified by static source review at a9650f1 and merged main 00a013b: name_first is enabled for 0.4 only, preserving the exact 0.3 diagnostic; the non-object-note guard is also gated. The added regression pins the original 0.3 wording. Resolving this original replay finding. No tests were rerun in this follow-up.

src/nandatown/evaluator.py
JamesCarnley: Verified by static source review at a9650f1 and merged main 00a013b: unanswered requests addressed elsewhere now make the response stage incomplete while leaving seller-specific handling stages alone. The regression definitions include both an auditor recipient and the buyer itself. This addresses the scope clarification; resolving it. No tests were rerun in this follow-up.

PR #280, opened 2026-09-15
fix: use credentials written into an endpoint URL, never record them

JamesCarnley:
Follow-up to the review of merged `00a013b`: URL confidentiality.

### Problem

An endpoint behind basic authentication can be tested by writing its credentials
into the URL, and httpx sends them. Reproduced on `main`:

- `test-agent` printed the password and recorded it in the run, events, intents,
  report and rerun command.
- `receipt` signed it into a receipt while announcing that "nothing private
  leaves the bundle".
- `pulse` stored it, and after the target was re-pointed at a clean URL, still
  printed it as an "earlier endpoint", as did `--report` and `--records`.

### Decision

Codex's constraints were: preserve exact endpoint distinctions, don't silently
strip credentials, and don't rewrite signed evidence. Within those, the
maintainer chose to **keep credential URLs usable, and never print or record
the credentials Town recognises**, rather than refuse them. Stripping would test a different endpoint:
against a real basic-auth agent, the same URL without credentials gets 401.

### Change

- **Credentials are recognised where httpx finds them**, in the URL's
  authority, never by searching text for things that look like URLs. A password
  may hold quotes, brackets or spaces. A URL httpx cannot parse is not called,
  and everything before its last `@` is withheld.
- **The recorder replaces only the operator's own credentials.** Only the
  operator's locators are registered (the subject URL, and a resolved index
  entry), and only their exact credentials are replaced, in every spelling they
  take: as written, httpx-normalised, percent-decoded, and shell-quoted, always
  inside `://…@`. An agent's own text is recorded as it said it.
- **Evidence and Pulse history show `<credentials 1a2b3c4d>`**, an HMAC of the
  credentials httpx sends, keyed by a random secret in the Town home that is
  never recorded. Two credential sets for one host stay distinguishable, and a
  short password can't be guessed from the label. If the key can't be used,
  Town records `<credentials withheld>` instead and warns: it never crashes and
  never leaks.
- **Receipts show `<credentials withheld>`,** the same on every machine; the
  bundle fingerprint still pins the exact endpoint. The CLI no longer claims
  nothing private leaves the bundle.
- **The rerun asks for `<operator-supplied-url>`,** as a Track rerun already does
  for an unrecorded command. An index rerun stays exact.
- **Evidence recorded earlier is not rewritten.** Reports, `replay`,
  `visualize` and new receipts withhold its credentials, and receipts issued
  before this still verify.
- **Echoes are closed.** `a2a test` and the Track `--agent` errors no longer
  repeat credentials. A Pulse refusal of any `--target` holding an `@` repeats
  none of it, because a mistyped or mis-split target can put part of a
  password in its name.

### What Town does not recognise

The guarantee covers recognised credentials only. The guide
(`docs/testing-an-existing-agent.md`) now says exactly where it stops:

- **A token outside the user information,** such as in a query string or a
  header, is used, printed and recorded as written.
- **A password with an unencoded `/`, `?` or `#`** ends the host early. Two
  outcomes:
  - **The URL no longer parses,** as in `user:pass/word@host`, which is the
    usual case. Town does not call it, and withholds everything before its last
    `@`.
  - **It still reads as a host,** as in `user:1234/word@host`,
    `token/word@host` or `user:a@b/c@host`. The URL goes to that host, and the
    rest of the password is used, printed and recorded as written. Town can't
    tell this from an ordinary `@` in a path such as `/users/a@b`, so it
    neither refuses nor rewrites the URL. `test-agent --url`, `a2a test` and
    `pulse` print a note without the URL; a URL from an `--index` file gets
    none. Refusing such URLs would change what Town accepts, so that is left
    for a maintainer.

Also known:
- Labels belong to one Town home.
- A Town from before this change reports a new receipt's withheld subject as a
  mismatch when checked with `--bundle`.
- A card that advertises its own authenticated URL is fingerprinted as
  served, so `card_retrieved.digest` covers the password: a short one can be
  guessed offline from a bundle. Pinning depends on that digest, so this is
  documented rather than changed.
- A bundle recorded before this change whose preview was cut inside the
  credentials still shows that fragment in displays, which need the `@` to
  recognise them. As the guide says, rerun rather than share such evidence.
- A report of a Track bundle recorded before 4 September, when harness commands
  were still recorded raw, can withhold a host that was never credentials, such
  as `http://host:8000 --contact ops@example.com` read as one URL. That hides
  more than it should; it never shows a secret.

### Verification

- **Real localhost:** the reference A2A agent behind real basic auth, with
  passwords containing quotes and brackets. The right credentials pass, and
  wrong or missing ones fail with 401. No printed line, bundle file, report,
  receipt or Pulse history contains them.
- **Old evidence:** a genuine bundle and receipt written by `00a013b` with fake
  credentials. Reports, `replay`, `visualize` and new receipts withhold them,
  and the old receipt still verifies. With #284 merged too, replaying that Path
  bundle shows `<credentials withheld>`.
- **Unchanged without credentials:** report, replay and visualize output is
  byte-identical to the previous revision for Lab, Track and credential-free
  Path bundles.
- **All six follow-up branches merged together on `main`** (#280–#285):
  1,739 passed, and every Track profile passes and verifies. A credentialed Path
  run reaches `TOWN-TESTED`, and after `replay` and `visualize` the password is
  nowhere on disk.
- **Full suite:** 1,664 passed.
- **Independent review, seven rounds.** Each round found real problems, all
  fixed here:
  - a regex that ended URLs at quotes;
  - agent text being rewritten;
  - key races, and a key claim waited on for five seconds per command;
  - old Track harness URLs, including a password with a space;
  - shell-quoted passwords;
  - a credential-free report withholding its own rerun command;
  - Pulse refusals repeating a mistyped target, then not saying which target
    failed;
  - `visualize` still writing old evidence's credentials into `town.html`;
  - from the maintainer review: `a2a test` printing a card or artifact that
    repeats the authenticated URL, and 200-character previews cut after the
    password and before its `@`;
  - privacy wording that overstated what is recognised, corrected against
    actual behaviour for each ambiguous URL shape.

  Its reproductions pass here, including 106 credential spellings and 60,000
  fuzzed URLs.


src/nandatown/cli.py
```diff
@@ -694,7 +731,18 @@ def cmd_a2a(args: argparse.Namespace) -> int:
...
+            AT_AFTER_HOST_NOTE,
+            at_after_host,
+            safe_message,
+        )
+
+        if at_after_host(args.url):
+            print("note: this URL has an '@' after its host."
+                  f" {AT_AFTER_HOST_NOTE}")
         report = probe_endpoint(args.url)
+        report["problems"] = [safe_message(args.url, problem)
```
JamesCarnley, line 751: [P1] Sanitize successful A2A report fields, not only problems

Only `report["problems"]` goes through the new protection, but `probe_endpoint()` copies card `name`, `version`, and `url` into `report["card"]` (a2a_adapter.py:230), and artifact text into `report["artifact"]` (:244). Source-traced trigger: test `http://alice:s3cret-pw@h` against an endpoint whose AgentCard advertises that same authenticated URL. The successful report has no problems, and the JSON printed below still contains `s3cret-pw`. This is the exact operator credential this PR promises to withhold, not an unrelated secret in arbitrary agent text. Apply the registered-URL scrubber across the report before serialization, preserving the malformed-URL error handling, and add a successful-card/artifact echo regression. For truncated artifact previews, redact the full source before truncating, as in the companion finding. This trace was not executed.

src/nandatown/path_runner.py
```diff
@@ -258,24 +259,40 @@ def _semantic_fulfillment_stage(
...
+        self.scrubber.register(url)
 
     def emit(self, observer: str, kind: str, subject: str,
              detail: dict[str, Any] | None = None) -> None:
         self.events.append(TownEvent(
             event_id=f"ev-{len(self.events) + 1}", run_id=self.run_id,
             at=time.time(), observer=observer, kind=kind,
-            subject=subject, detail=detail or {}))
+            subject=scrub(subject, self.scrubber),
+            detail=scrub(detail or {}, self.scrubber)))
```
JamesCarnley, line 288: [P1] Redact full fulfillment text before the preview truncates it

The new recorder scrub runs too late for `fulfillment_unparseable`: lines 550-555 first take `text[:200]`, then call `emit`. Source-traced trigger: with registered endpoint `http://alice:s3cret-pw@h`, let the completed task's malformed text be 178 `x` characters followed by that URL. Its 200-character preview ends with the complete `s3cret-pw` but excludes the following `@`. `Scrubber` only replaces the complete `://userinfo@` token, so the preview is persisted in events.jsonl with the password intact. Scrub the full text before slicing/formatting it, and cover truncation inside userinfo with a regression. This is recognized credential input defeated by truncation, not the documented ambiguous-URL limitation. Source inspection only; not an executed reproduction.

JamesCarnley (commented):
Static correctness/privacy review at e7630e5 (base 00a013b), with an independent reviewer and primary-agent source cross-check. No tests, reproductions, contributor code or builds were executed; the CI results are existing author/CI evidence. The recognized-userinfo boundary, preserved authenticated transport, and non-mutating historical displays are directionally sound, but two reachable output paths still disclose recognized operator credentials. Please address the inline findings before merging. These are gaps in the new protection, not claims that the original raw-output behavior was introduced by this PR.

src/nandatown/cli.py
JamesCarnley: Fixed in 02f7d85, with follow-ups in ab93eb4 and aa30686. I reproduced it first: a real localhost agent whose card advertises its authenticated URL made `a2a test` print the password.

- `a2a test` now runs the whole report through the operator URL's withholding before printing it: card name, version and URL, the artifact, and task state. Problems still go through `safe_message`, so malformed-URL handling is unchanged.
- `probe_endpoint` withholds the artifact before cutting its 200-character preview. It accepts any shape and cuts only a string or a list, so a dict or other non-string text part no longer crashes.
- The walk is iterative, so deeply nested agent output can't turn into a Town error.

Regression tests:
- `test_a2a_test_withholds_credentials_a_card_repeats` runs a real agent process.
- `test_a2a_test_withholds_before_cutting_its_artifact_preview` goes through the CLI with text, list and object artifacts. The text case fails if the pre-cut withholding is removed.
- `test_an_a2a_artifact_preview_cannot_cut_credentials_short`
- `test_deeply_nested_agent_output_is_the_agents_failure_not_towns`

An independent re-review swept every cut offset in and around `://user:password@` over 662 runs and found no leaks. On e7630e5 the same sweep leaked in 531 runs.

src/nandatown/path_runner.py
JamesCarnley: Fixed in 02f7d85, with a follow-up in ab93eb4. I reproduced your trace first: 178 `x` characters followed by the registered URL put `s3cret-pw` into `events.jsonl`.

- The unparseable fulfillment is now withheld before its preview is cut, for text and non-string values alike. Non-strings are withheld before `repr`.
- `test_a_fulfillment_preview_cannot_cut_credentials_short` cuts exactly after the password and checks that no bundle file contains it.
- An independent re-review swept every cut offset through `test-agent --url` on @0.3 and @0.1: no leaks in 1,098 runs, against 98 leaks in 549 runs on e7630e5.
- The first version of this fix walked nested output recursively, so a 600-deep text part became a Town error. ab93eb4 makes the walk iterative, and a test covers it.

I found no other place where Path or `a2a test` output is cut before withholding. Evaluator notes read events that were already withheld.

JamesCarnley: Static re-review at current head 5bfb90c confirms the truncation order is fixed: the fulfillment value is scrubbed first, then the string or representation is cut to 200 characters, and recorder.emit applies the normal second defensive scrub. The exact cut-after-password regression is present, alongside non-string and deep-nesting controls. I did not rerun tests or contributor code in this review; execution evidence is the existing green CI. Resolving this thread.

src/nandatown/cli.py
JamesCarnley: Static re-review at current head 5bfb90c confirms this is addressed. The CLI registers the operator URL, passes that scrubber into probe_endpoint so artifact text is withheld before the 200-character preview, and then scrubs the complete report before JSON serialization; that covers card name/version/URL, task state, and nested artifact content. The added successful-card and string/list/object artifact cases exercise the intended boundary. I did not rerun tests or contributor code in this review; execution evidence is the existing green CI. Resolving this thread.

PR #285, opened 2026-09-15
fix(identity): let concurrent runs share a keystore without breaking attestations

JamesCarnley:
Follow-up found while reviewing the merged `00a013b`, prioritised by Codex.

### Problem

Runs that share a Town home share its keystore. Every Lab, Track and Path run
calls `attest_bundle`, which creates the operator's controller key on first use.
In a fresh home, concurrent runs raced to create it.

Reproduced on `main` through the CLI, with no test patching: 8 concurrent
`nandatown run voting` runs in each of 10 fresh homes. Of the 80 runs, 26 wrote
attestations that fail `nandatown verify`, and 4 crashed with
`JSONDecodeError`. Three causes:

- **Keys replaced.** `new_identity` checked for a key and then wrote one, so a
  second creator overwrote a key the first was already signing with.
- **Registry torn and lost.** The registry was truncated and rewritten in
  place. A concurrent reader could find it empty, and concurrent writers lost
  each other's entries or added duplicates.
- **Wrong entry chosen.** `identity()` returned the first registry entry with
  the name, which after a race need not match the key on disk. A home damaged
  this way kept producing unverifiable attestations.

### Change

`src/nandatown/identity_portable.py` only. There is no change to file formats,
key or registry locations, or any CLI.

- **Keys are never replaced.** A new key is created under an exclusive lock on
  the keystore (`flock`, or `msvcrt` on Windows), which the system releases if
  the process dies. It is written whole under a staging name and then
  hard-linked into place, which fails rather than replace a key already there.
  Without hard links, the path is first claimed with an exclusive create and
  the whole key then replaces that claim. Either way no key is replaced,
  whether or not any writer holds the lock, and a reader sees an empty claim
  (which it waits for) or a whole key, never part of one.
- **The registry is written whole.** It is updated under the same lock and
  replaced atomically, so readers never see it half written. It keeps its mode,
  and a symlinked registry is updated where the link points. In that case the
  lock of the directory it really lives in is taken too, so homes sharing one
  registry don't lose entries. Directories are identified and ordered by device
  and inode, so one directory spelled two ways, as on a case-insensitive disk,
  is never locked twice.
- **The key on disk decides,** provided the registry lists it under that name.
  So a home the race already damaged attests verifiably again, and no entry is
  removed. A key the registry lists under another name, or not at all, is an
  `IdentityError`, as on `main`.
- **Recovery:**
  - A key left without its registry entry, by a process stopped between the
    two writes, is registered on next use.
  - A staged private key left by a killed writer is deleted when a later
    process takes the lock, once it is ten minutes old. Only regular files
    named exactly as Town stages them are removed.
  - An empty key left by a writer killed after claiming it is reported with
    how to recover.
- **Fallbacks:**
  - **No hard links:** replace, under the lock, only when no key exists.
  - **Locks unavailable** (e.g. NFS without a lock daemon, or a shared
    registry directory whose lock this user cannot open): keys are still never
    replaced, but concurrent writers can lose registry entries, and Town
    warns.

### Verification

- **CLI reproduction, same 80 runs:** all verify, with one registry entry per
  home. On `main`, 5 of the 10 homes had bad runs.
- **Regression tests.** They start real processes with widened timing windows,
  so they don't depend on luck. They cover:
  - concurrent runs attesting verifiably;
  - established keys and other names' entries preserved;
  - concurrent new names all registered;
  - a home damaged by the old race;
  - a key that arrives between the check and the link, with and without hard
    links;
  - a leftover staged key;
  - keys listed under another name or not at all, and damaged keys;
  - a symlinked registry with a strict mode;
  - two homes sharing a symlinked registry;
  - one directory spelled two ways (case-insensitive filesystems only);
  - a shared registry whose lock cannot be opened.

  Each test fails on the commit before its fix, except three that guard
  behaviour `main` already had. With the lock removed, the concurrency tests
  fail on every run. With the link replaced by a plain replace, the arrival
  tests fail. Repeated runs all pass.
- **Full suite:** 1,592 passed.
- **Merged with #280–#284 on `main`:** 1,739 passed. Eight concurrent runs in
  one fresh home all verify, with one registry entry.
- **Independent review:**
  - **Stress:** 3,480 processes in fresh homes, 16–32 at a time, mixing runs,
    receipts, `identity new`, grants and readers. No failures. The same harness
    failed every round on `main` and without the lock.
  - **Crashes and fallbacks:** `kill -9` while holding the lock, and filesystems
    without hard links.
  - **Findings addressed in later commits:**
    - the tests missing a swapped link;
    - leftover staged keys;
    - registry mode and symlinks;
    - `identity()` accepting unregistered keys;
    - homes sharing a symlinked registry losing entries;
    - a self-deadlock on case-insensitive disks;
    - from the maintainer review: cleanup deleting an established key named
      like a staging file, and the no-lock, no-hard-link fallback replacing a
      key (also when only some writers can lock);
    - a failed publication deleting a key that had already replaced its
      claim.
- **Not run:** the Windows branch; CI is Ubuntu only.


src/nandatown/identity_portable.py
```diff
@@ -70,47 +76,266 @@ def verify_signature(public_hex: str, payload: Any,
...
+            else:
+                _remove_staged(directory)
+            yield
+    finally:
+        os.close(fd)
+
+
+def _remove_staged(directory: str) -> None:
+    for entry in os.listdir(directory):
+        if entry.startswith(_STAGED_PREFIX):
```
JamesCarnley: [P1] Do not identify abandoned private-key files by this broad prefix

`_key_path(name)` returns `f"{name}.controller.key"`, and `identity new` accepts an unrestricted positional name (cli.py:584,861). Source-traced sequential trigger in a disposable keystore: `nandatown identity new .staged-seller --dir <temp-keystore>`, then `nandatown identity new bob --dir <temp-keystore>`. The first establishes `.staged-seller.controller.key`; the second takes `_directory_lock`, and this prefix check silently unlinks that established key as if it were abandoned staging. Signing with the existing identity then fails, and calling `new_identity('.staged-seller')` creates a different key/identity. Cleanup also runs in a symlinked registry's containing directory, where generic `.staged-*` neighbors need not belong to Town. Restrict removal to positively identified Town staging artifacts, never established controller keys or arbitrary neighboring files, and add preservation controls for both cases. Not executed; this follows the public CLI → `_locked` → `_remove_staged` path at the reviewed head.

src/nandatown/identity_portable.py
```diff
@@ -70,47 +76,266 @@ def verify_signature(public_hex: str, payload: Any,
...
+        path = self._key_path(name)
+        try:
+            try:
+                os.link(staged, path)
+            except FileExistsError:
+                pass
+            except OSError:
+                # No hard links here. Under the lock, with no key found,
+                # replacing is safe from every writer that takes the lock.
+                if not os.path.exists(path):
```
JamesCarnley: [P2] Require an acquired lock before the replacement fallback

This branch assumes own-directory exclusivity, but `_directory_lock` yields without it when flock raises ENOLCK/ENOTSUP/EOPNOTSUPP (lines 113-126). On a filesystem/environment that also rejects hard links, two same-name creators can both pass this existence check: B publishes and returns key B, then A replaces it with key A. B's subsequent `sign()` rereads the key file, so its attestation can name B while being signed with A. This is the original race persisting in a newly explicit fallback, not a regression on the successfully locked local path. The current warning mentions registry-entry loss only; it does not make replacing an established controller key safe. Permit replacement only with confirmed own-directory exclusivity, or refuse creation when neither exclusivity nor no-clobber publication is available. Add a combined no-lock/no-hard-link interleaving control; the separate fallback tests do not cover this composition. Source-traced only, not executed.

JamesCarnley (commented):
Static correctness review at 59fe528 (base 00a013b), with an independent reviewer and primary-agent cross-check; no tests, reproductions, builds or contributor code executed. Serializing creation, publishing complete key files, and atomically replacing registry contents address the reported race on the normal locked path. However, the new abandoned-file cleanup can delete established private keys through the ordinary CLI, so this is not ready to merge. Please preserve existing key ownership/names when narrowing cleanup; the inline trigger needs no concurrency. A second, conditional finding concerns the newly explicit no-lock plus no-hard-link fallback: it retains the original key-replacement race and needs a guard, rather than a broader redesign.

src/nandatown/identity_portable.py
JamesCarnley: Fixed in e42866a, with follow-ups in e1d831a and 993fd43. I reproduced your CLI trigger first: after `identity new .staged-seller`, creating `bob` deleted `.staged-seller.controller.key`, and signing then failed.

- Town stages under `.nandatown-staged-<32 hex>.tmp`.
- Cleanup removes only regular files with exactly that name, and only when they are older than 10 minutes. No `<name>.controller.key` can match, a directory with that name is left alone, and a file another writer is still publishing is never removed.
- In a symlinked registry's directory, other files are left alone.

`test_cleanup_removes_only_what_town_staged` covers identities named `.staged-seller` and like the new pattern, plus non-Town neighbours beside a symlinked registry. `test_a_private_key_a_dead_writer_left_staged_is_removed` covers in-flight files and a directory with the staged name. Both fail on 59fe528.

An independent re-review confirmed that the reviewed-head trigger and variants keep every identity and neighbour file.

src/nandatown/identity_portable.py
JamesCarnley: Fixed in e1d831a, with follow-ups in 993fd43. My first attempt, e42866a, still replaced keys when only some writers could lock, so I changed the approach.

- **Publication.** Without hard links, every writer, locked or not, first claims the key path with an exclusive create. That fails if a key or another claim is there. The writer then puts its whole key over its own claim. No key is replaced whoever holds the lock, and a reader sees an empty claim or a whole key, never part of one.
- **Readers.** A reader that finds an empty claim waits up to 5 s, deciding from that same read.
- **Failure.** If publishing fails, only a claim that is still ours and empty is withdrawn, so a key that already replaced it is never deleted.
- **Warning.** The unlocked-mode warning's claim that a key is never replaced is now true.

Tests:
- `test_without_hard_links_a_key_arriving_before_publication_is_kept` writes the rival key at the exclusive create itself, with and without the lock.
- `test_a_writer_without_the_lock_waits_for_one_that_has_it` has a no-lock writer arrive while a locked writer publishes. On e42866a it gets a different key.
- Claimed-key waiting, the abandoned-claim message, and failed-publication withdrawal are also covered.

An independent re-review ran 920 creators under no-lock, no-hard-link, mixed and jittered modes. It found no signature mismatches, replaced keys, crashes or hangs.

src/nandatown/identity_portable.py
JamesCarnley: Static re-review at current head 993fd43 confirms this finding is addressed. Staging now uses the Town-specific .nandatown-staged-<32 hex>.tmp namespace; cleanup requires a full match, a regular file, and age beyond the abandoned threshold. Established <name>.controller.key files, including .staged-seller.controller.key and identities named like the staging token, cannot match, while the added controls preserve symlink-directory neighbours. I did not rerun tests or contributor code in this review; execution evidence is the existing green CI. Resolving this thread.

src/nandatown/identity_portable.py
JamesCarnley: Static re-review at current head 993fd43 confirms the controller-key replacement race is addressed. When hard links are unavailable, every writer now claims the final key path with O_EXCL before publication; a writer that does not own that claim waits for the complete key instead of replacing it. Failure cleanup checks the original claim inode and emptiness before removal. This holds independently of whether _directory_lock was acquired; the separately documented unlocked registry-entry race remains, but it no longer permits controller-key replacement. The combined no-lock/no-hard-link and mixed-lock test definitions cover the requested composition. I did not rerun tests or contributor code in this review; execution evidence is the existing green CI. Resolving this thread.

